PR #1, opened 2020-05-19
WIP: replace ingest cache infrastructure with chancacher

djfritz:
Replace bolt based cache with much simpler (and more performant) chancacher, which is simply a two-step pipeline of `chan interface{}` that can cache to disk using gob, recover on restart, drain to disk on close, and absorb bursts of writes. 

This also creates two new cache parameters in the ingesters and renames others.

Default behavior changes include having the cache enabled by default, and functions like WaitForHot() can, depending on mode, return before connections are made (allowing caching immediately). 

@floren @traetox


ingest/muxer.go
```diff
@@ -634,11 +597,13 @@ mainLoop:
 				//we haven't hit our timeout yet, just continue
 				continue
 			}
-			//timeout, check state and force a return
-			//if we have a hot, filebacked cache, then endpoints are go for ingest
-			if im.cacheRunning && im.cacheError == nil && im.cacheFileBacked {
+
+			if im.cacheEnabled {
```
djfritz, line 601: check cache mode, not if it's enabled.

ingest/muxer.go
```diff
@@ -513,17 +471,22 @@ func (im *IngestMuxer) NegotiateTag(name string) (tg entry.EntryTag, err error)
...
 	// update the tag list and map
 	im.tags = append(im.tags, name)
-	for i, v := range im.tags {
-		im.tagMap[v] = entry.EntryTag(i)
-	}
-	if im.cacheEnabled && im.cacheFileBacked {
-		// Now update the stored tags list
-		if err = im.cache.UpdateStoredTagList(im.tags); err != nil {
-			return
+
```
djfritz, line 474: remove the cached tag map if the cache is empty.

chancacher/chancacher.go
```diff
@@ -0,0 +1,365 @@
...
+	if c.cache {
+		var err error
+
+		err = os.MkdirAll(c.cachePath, 0755)
+		if err != nil {
+			return nil, err
+		}
+
+		// create r and w files
+		r, err := os.OpenFile(filepath.Join(c.cachePath, "cache_a"), os.O_APPEND|os.O_CREATE|os.O_RDWR, 0644)
```
djfritz, line 107: This should merge the R and W files and create a single R file if W has data in it so we avoid new encoders against a single file.

chancacher/chancacher.go
```diff
@@ -0,0 +1,365 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		w, err := os.OpenFile(filepath.Join(c.cachePath, "cache_b"), os.O_APPEND|os.O_CREATE|os.O_RDWR, 0644)
+		if err != nil {
+			return nil, err
+		}
+
+		c.cacheR = NewFileCounter(r)
```
djfritz, line 117: NewFileCounter should take a starting value, equal to the size of the file.

chancacher/chancacher.go
djfritz: I think this is resolved by https://github.com/gravwell/gravwell/pull/1/commits/c4ec6e6e5fca1b669470beb6122421a061dc1cd9

chancacher/chancacher.go
djfritz: Resolved by https://github.com/djfritz/gravwell/pull/2

ingest/muxer.go
djfritz: Resolved by https://github.com/djfritz/gravwell/commit/34e537694c278b79fda2b717cd8b4a14df774a03

ingest/muxer.go
djfritz: Resolved by https://github.com/djfritz/gravwell/pull/3

PR #5, opened 2020-05-26
Dev

traetox:
This reworks the kafka consumer to allow for dynamically deriving tags from kafka headers.

May of the other platforms that expect to send through kafka don't like to use lots of topics, so they send additional metadata via kafka message headers.

This will allow us to be more compatible with other kafka producers.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,32 +7,42 @@ require (
 	collectd.org v0.3.1-0.20181025072142-f80706d1e115
 	github.com/Pallinder/go-randomdata v1.2.0
 	github.com/Shopify/sarama v1.24.1
+	github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d // indirect
 	github.com/aws/aws-sdk-go v1.25.46
+	github.com/bet365/jingo v0.10.0
 	github.com/buger/jsonparser v0.0.0-20191004114745-ee4c978eae7e
 	github.com/bxcodec/faker/v3 v3.3.1
 	github.com/dchest/safefile v0.0.0-20151022103144-855e8d98f185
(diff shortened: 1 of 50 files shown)


go.mod
```diff
@@ -20,10 +20,12 @@ require (
 	github.com/floren/o365 v0.0.1
 	github.com/fsnotify/fsnotify v1.4.9
 	github.com/go-ole/go-ole v1.2.4 // indirect
+	github.com/gobwas/glob v0.2.3
 	github.com/google/go-write v0.0.0-20181107114627-56629a6b2542
 	github.com/google/gopacket v1.1.17
 	github.com/google/uuid v1.1.1
 	github.com/gravwell/gcfg v1.2.5
+	github.com/gravwell/ingest/v3 v3.3.12
```
floren, line 28: Can you run 'go mod tidy' and see if it gets rid of this? I don't see a reference to gravwell/ingest in your *code* so maybe it's just old stuff.

ingesters/kafka_consumer/main.go
```diff
@@ -108,6 +108,9 @@ func main() {
 			}
 		}
 	}
+	for name, c := range cfg.Consumers {
+		fmt.Printf("%s: %+v\n", name, c)
```
floren, line 112: Printf? Or should this use the logging stuff?

traetox: resolved with https://github.com/gravwell/gravwell/pull/5/commits/f1e3846c2089a02ee247e0162af36eb5dcabc6f6

go.mod
traetox: go mod tidy appears to have cleaned all that garbage up.

floren:
lgtm

PR #16, opened 2020-06-01
muxer bouncing and tag corruption

traetox:
fixing issue where muxer could build up a list of entries and resend them, potentially even corrupting tag assignments when an ingest connection repeatedly bounces while under high load

### DIFF
--- a/ingest/muxer.go
+++ b/ingest/muxer.go
@@ -48,17 +48,16 @@ var (
 )
 
 const (
+	mb               = 1024 * 1024
 	empty   muxState = 0
 	running muxState = 1
 	closed  muxState = 2
 
 	defaultRetryTime     time.Duration = 10 * time.Second
 	recycleTimeout       time.Duration = time.Second
-	maxEmergencyListSize int           = 256
+	maxEmergencyListSize int           = 64
 	unknownAddr          string        = `unknown`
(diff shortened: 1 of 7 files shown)


ingest/muxer.go
```diff
@@ -1439,7 +1452,7 @@ func (eq *emergencyQueue) clear(igst *IngestConnection, tt *tagTrans) (ok bool)
 					if !ok {
 						// could not translate, push it back on the queue and bail
 						// first we need to reverse the ones we have already translated, ugh
-						for j := 0; j < i; j++ {
+						for j := 0; j <= i; j++ {
```
floren, line 1455: This will attempt to reverse a tag which never got translated in the first place... I think j < i was correct.

ingesters/kafka_consumer/consumer.go
```diff
@@ -306,6 +306,7 @@ func (kc *kafkaConsumer) flush(session sarama.ConsumerGroupSession, msgs []*sara
 		if ent.Tag, ent.SRC, err = kc.resolveSourceAndTag(m); err != nil {
 			return
 		}
+		fmt.Println(ent.Tag, string(ent.Data))
```
floren, line 309: Get rid of this

PR #17, opened 2020-06-02
migrate to gravwell/ipfix library

floren:


go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/eapache/go-resiliency v1.2.0 // indirect
 	github.com/elastic/beats v7.6.2+incompatible
-	github.com/floren/ipfix v1.4.1
+	github.com/floren/ipfix v1.4.1 // indirect
```
traetox, line 19: can we run go mod tidy?  why is this still hanging out?

traetox: why indirect?

floren:
ptal

PR #21, opened 2020-06-02
ingest: add transition stub from old boltdb cache to neucache

djfritz:
@floren the only code in boltcache.go that needs review is `boltTranstition()`, the rest of that file is the old cache. 

The idea here is to trigger on someone pointing neucache to a regular file, which we assume to be an old cache. We attempt to translate it to neucache in place, and just carry on.

### DIFF
--- a/ingest/boltcache.go
+++ b/ingest/boltcache.go
@@ -94,6 +94,7 @@ func boltTransition(c MuxerConfig) error {
 	if err != nil {
 		return err
 	}
+	defer os.RemoveAll(tmpd)
 
 	cc, err := chancacher.NewChanCacher(0, tmpd, 0)
 	if err != nil {
--- a/ingest/boltcache.go
+++ b/ingest/boltcache.go
@@ -1,0 +1,867 @@
+/*************************************************************************
+ * Copyright 2017 Gravwell, Inc. All rights reserved.
+ * Contact: <legal@gravwell.io>
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD 2-clause license. See the LICENSE file for details.
+ **************************************************************************/
+
+package ingest
+
+import (
+	"encoding/json"
+	"errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+	"unsafe"
+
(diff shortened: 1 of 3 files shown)


ingest/boltcache.go
```diff
@@ -0,0 +1,867 @@
...
+	if err != nil {
+		return nil
+	}
+	if fi.IsDir() {
+		return nil
+	}
+
+	// We probably do. Create a temporary chancacher, read the bolt cache
+	// into it, and then shuffle the files around.
+	tmpd, err := ioutil.TempDir("", "ingestCacheTransition")
```
traetox, line 93: where is this tempdir firing off?  We want to make sure it lands on the same location as the existing cache in case anyone has a dedicated cache dir (I often do).

ingest/boltcache_test.go
```diff
@@ -0,0 +1,109 @@
...
+	// build a MuxerConfig and transition
+	err = boltTransition(MuxerConfig{
+		CachePath: tmp.Name(),
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// tmp is now a directory bcause of the transition
+	defer os.RemoveAll(tmp.Name())
```
traetox, line 79: why are there two defers?  one to Remove and one to RemoveAll ?

ingest/boltcache.go
djfritz: The cache eventually gets back to the specified location but we can fire a temp file in the filepath.Base() of the provided dir.

ingest/boltcache_test.go
djfritz: tmp starts as a regular file, and the name becomes a directory under the transition, so we ignore what will be an erroneous os.Remove and add a RemoveAll for the new cache.

djfritz:
> Are the two defers intended?

Yes - the name is overloaded because it transitions from a regular file to a directory of files so we have to upgrade the Remove to a RemoveAll. We discard the first error so it doesn't matter.

ingest/boltcache.go
djfritz: Updated.

djfritz:
@traetox PTAL

PR #45, opened 2020-07-01
adding an additional operator on the jsonarraysplit preprocessor

traetox:
JSONARRAY preprocessor now supports an `Additional-Fields` argument that lets you pack along some more stuff when cracking apart arrays.


floren:
lgtm

PR #74, opened 2020-09-04
add LogSourceOverride to muxer

floren:
allows log messages to be sent with their own particular IP instead of the same default as everyone else


traetox (changes requested):
Can we add that field to `ingest/config/IngestConfig` too so that all the configs can "just have it"


Then rip through ingesters and add the extraction/config setup?

floren:
PTAL

traetox (approved):
LGTM, we can add/update other ingesters as it makes sense.

PR #82, opened 2020-09-23
rewrite filefollow regex reader to be more intelligent, and add user-…

floren:
…defined time formats to the log handler


filewatch/handler.go
```diff
@@ -82,6 +84,15 @@ func NewLogHandler(cfg LogHandlerConfig, w logWriter) (*LogHandler, error) {
 				return nil, err
 			}
 		}
+		if cfg.UserTimeRegex != `` {
```
traetox, line 87: this pattern should probably be extended to all ingesters.  Its SUPER useful.

traetox (approved):
LGTM

PR #86, opened 2020-09-25
Add special ${_SRC_} sigil to regexextract preprocessor.

floren:
If your regexextract Template contains `${_SRC_}`, it will be replaced with a string representation of the SRC field of the entry. Thus you can now do something like this:
```
[Preprocessor "srcextract"]
	Type=regexextract
	Regex="msg: (?P<msg>.+)$"
	Template=`{"srcip":"${_SRC_}", "msg":"${msg}"}`
```
and get entries which look like this:
```
{"srcip":"127.0.0.1", "msg":"foo bar baz"}
```


traetox (approved):
LGTM

PR #88, opened 2020-10-01
adding in the new drop module preprocessor

traetox:


ingest/processors/drop.go
```diff
@@ -0,0 +1,55 @@
...
+	return
+}
+
+func NewDrop(cfg DropConfig) (*Drop, error) {
+	return &Drop{
+		DropConfig: cfg,
+	}, nil
+}
+
+// Drop does not have any state, and doesn't do shit
```
djfritz, line 36: maybe don't swear in the public repo

djfritz (changes requested):

ingest/processors/drop.go
traetox: agreed

PR #95, opened 2020-10-14
remove log message that causes deadlock

### LINKED ISSUE #94: deadlock in WriteEntryTimeout
IngestMuxer.WriteEntryTimeout can be called via NegotiateTag, which already holds the IngestMuxer.mtx for writing. WriteEntryTimeout attempts to get a read lock on mtx, causing a deadlock.

muxer.go:770
muxer.go:491

djfritz:
Fixes #94

### DIFF
--- a/ingest/muxer.go
+++ b/ingest/muxer.go
@@ -536,7 +536,6 @@ func (im *IngestMuxer) NegotiateTag(name string) (tg entry.EntryTag, err error) 
 			remoteTag, err := v.NegotiateTag(name)
 			if err != nil {
 				// something went wrong, kill it and let it re-initialize
-				im.Error("NegotiateTag on %v for %v: %v", v.conn.RemoteAddr(), name, err)
 				v.Close()
 				continue
 			}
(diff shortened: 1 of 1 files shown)


ingest/muxer.go
```diff
@@ -536,7 +536,6 @@ func (im *IngestMuxer) NegotiateTag(name string) (tg entry.EntryTag, err error)
 			remoteTag, err := v.NegotiateTag(name)
 			if err != nil {
 				// something went wrong, kill it and let it re-initialize
-				im.Error("NegotiateTag on %v for %v: %v", v.conn.RemoteAddr(), name, err)
```
traetox, line 539: this log might be super handy, should we:
```
mtx.Unlock()
im.Error("NegotiateTag on %v for %v: %v", v.conn.RemoteAddr(), name, err)
mtx.Lock()
```

?  If that has other affects i am not seeing, feel free to reject

traetox (commented):
LGTM

djfritz:
@traetox PTAL

traetox (approved):
LGTM

PR #111, opened 2020-10-27
fix recursive read locks, bump version

djfritz:


ingest/muxer.go
```diff
@@ -748,9 +746,7 @@ func (im *IngestMuxer) WriteEntry(e *entry.Entry) error {
 	if e == nil {
 		return nil
 	}
-	im.mtx.RLock()
 	runok := im.state == running
```
traetox, line 749: don't create this weird temporary variable if we don't need the lock

```
if im.state != running {
    return ErrNotRunning
}

File: ingest/muxer.go [hunk 2]
```diff
@@ -766,9 +762,7 @@ func (im *IngestMuxer) WriteEntryContext(ctx context.Context, e *entry.Entry) er
...
```
traetox, line 765: same

ingest/muxer.go
```diff
@@ -788,9 +782,7 @@ func (im *IngestMuxer) WriteEntryTimeout(e *entry.Entry, d time.Duration) (err e
 	if e == nil {
 		return
 	}
-	im.mtx.RLock()
 	runok := im.state == running
```
traetox, line 785: same

ingest/muxer.go
```diff
@@ -1286,7 +1278,6 @@ loop:
 			}
 			continue
 		}
-		im.Info("Connection to %v established, completing negotiation & requesting approval to ingest", tgt.Address)
```
traetox, line 1289: can we make this a LocalInfo

It helps with debugging.

traetox (changes requested):

ingest/muxer.go
floren: he's just moved it lower down

floren (approved):

traetox (approved):

PR #117, opened 2020-11-02
Allow jsonarraysplit preprocessor to split on a top-level array

floren:
Leaving the Extractions parameter empty will just make it split the data instead of trying to extract a json field first.


traetox (approved):
LGTM

PR #119, opened 2020-11-12
improve error checking around tagcache and use renameio for atomic writing

djfritz:


ingest/muxer.go
```diff
@@ -390,18 +393,15 @@ func readTagCache(p string) (map[string]entry.EntryTag, error) {
...
-	if err != nil {
-		return err
-	}
+	var b bytes.Buffer
 
-	enc := gob.NewEncoder(f)
-	err = enc.Encode(&t)
+	enc := gob.NewEncoder(&b)
+	err := enc.Encode(&t)
 	if err != nil {
```
traetox, line 400: if err := gob.NewEncoder(&b).Encode(&t); err != nil {
    return err
}

traetox (approved):
LGTM if the tests are good

djfritz:
Testing appears to work with corrupted, empty, and normal tagcache.

PR #121, opened 2020-11-18
remove old TLS cert verification flag from ingest/config

djfritz:


traetox (changes requested):
you can't remove the actual config, or people that have the old config value will fail to start.

djfritz:
NEVERMIND

PR #125, opened 2020-12-10
Ise preprocessor

traetox:
new ISE preprocessor for Cisco garbage


traetox:
@djfritz @floren PTAL   Should be good to go now

floren (approved):

djfritz (approved):
LGTM

PR #136, opened 2021-01-07
Implement callbacks for IngestState messages on the EntryReader. 

floren:
This will help the Federator keep its Children up to date


traetox (approved):
LGTM

PR #137, opened 2021-01-07
converting preprocessors to use blocks on input and output

traetox:
This should reduce some allocations for the nominal cases.

I also added some more tests to get better coverage on some of the preprocessors we didn't have tests for


djfritz (approved):
LGTM

PR #139, opened 2021-01-08
Make all major ingesters (except for Windows ones) report their full …

floren:
…config via IngesterState messages.

Please review and verify that we are leaving out sensitive config elements (secrets / auth tokens)


traetox (approved):
seems reasonable.

PR #141, opened 2021-01-14
actually enforce the maximum entry sizes

traetox:


ingest/entryReader.go
```diff
@@ -286,6 +287,9 @@ func (er *EntryReader) read() (*entry.Entry, error) {
 	if err = er.fillHeader(ent, &id, &sz); err != nil {
 		return nil, err
 	}
+	if sz > uint32(MAX_ENTRY_SIZE) {
```
floren, line 290: this is already checked at the end of fillHeader

djfritz (approved):
LGTM

ingest/entryReader.go
traetox: shit, fixing

PR #143, opened 2021-01-15
add SRC router preprocessor

floren:


ingest/processors/srcrouter.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+		if r.tag == `` {
+			r.drop = true
+		}
+		rts = append(rts, r)
+	}
+	return
+}
+
+func getSrcRoute(v string) (a string, b string, err error) {
```
traetox, line 166: can we have this execute an `ingest.CheckTag(b)` here too, that way we catch bag tags right out of the gate.

ingest/processors/srcrouter_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+const ()
+
+var (
+	src = SrcRouteConfig{
+		Route: []string{`1.1.1.1:foo`, `2.2.2.2:bar`, `3.3.3.3:`},
+	}
+)
+
+func TestSrcRouteConfig(t *testing.T) {
```
traetox, line 27: add in a test with some bad route defs and some files that are malformed?

traetox (changes requested):

djfritz (commented):
Agree with @traetox's comments, otherwise LGTM

floren:
PTAL

traetox (approved):
LGTM

PR #157, opened 2021-02-16
Client API so we can fire preview searches

traetox:
updating client to also be able to fire off preview mode searches, this also simplified the implementation of a few APIs

### DIFF
--- a/client/searchCtrl.go
+++ b/client/searchCtrl.go
@@ -219,6 +219,10 @@ func (c *Client) StartSearchEx(sr types.StartSearchRequest) (s Search, err error
 		return
 	}
 	searchSubProto := s.SearchSockets.Search
+	// we get a new set of sockets on each GetSearchSocket.
+	// So we can close the main search suproto when done
+	// The main protocols for the search will be left open
+	defer searchSubProto.Close()
 
 	if err = searchSubProto.WriteJSON(sr); err != nil {
 		return
--- a/client/searchCtrl.go
+++ b/client/searchCtrl.go
@@ -184,23 +184,43 @@ func (c *Client) ParseSearchWithResponse(query string, filters []types.FilterReq
 // and the specified start and end times. If "nohistory" is set, the search will
 // be hidden in the user's search history; if false, it will be visible.
 func (c *Client) StartBackgroundSearch(query string, start, end time.Time, nohistory bool) (s Search, err error) {
-	s.SearchString = query
-	s.SearchStart = start.Format(time.RFC3339Nano)
-	s.SearchEnd = end.Format(time.RFC3339Nano)
-	s.Background = true
-	s.NoHistory = nohistory
+	sr := types.StartSearchRequest{
+		SearchString: query,
+		SearchStart:  start.Format(time.RFC3339Nano),
+		SearchEnd:    end.Format(time.RFC3339Nano),
+		NoHistory:    nohistory,
+		Background:   true,
+	}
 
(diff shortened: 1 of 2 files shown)


traetox:
here is an example snippet to fire preview searches based off the examples in your repo:

```
package main

import (
	"errors"
	"flag"
	"fmt"
	"log"
	"strings"
	"time"

	"github.com/gravwell/gravwell/v3/client"
	"github.com/gravwell/gravwell/v3/client/objlog"
	"github.com/gravwell/gravwell/v3/client/types"
)

var (
	server   = flag.String("s", "", "Address and port of Gravwell webserver")
	username = flag.String("u", "admin", `Username (default "admin")`)
	useHttps = flag.Bool("https", false, "Set true to use https")
	password = flag.String("p", "changeme", `Password (default "changeme")`)
	query    = flag.String("q", "", "Query string")
	duration = flag.String("d", "-1h", "Search duration")
	background = flag.Bool("b", false, "background")
)

func main() {
	flag.Parse()

	// First things first: make sure they gave us a valid duration
	start, end, err := getTimeDuration(*duration)
	if err != nil {
		log.Fatal(err)
	}

	// For simplicity, we ignore cert validity (self-signed ok)
	c, err := client.NewClient(*server, false, *useHttps, &objlog.NilObjLogger{})
	if err != nil {
		log.Fatalf("NewClient failed: %v", err)
	}
	defer c.Close()

	// Log in
	if err = c.Login(*username, *password); err != nil {
		log.Fatalf("Login failed: %v", err)
	}

	// Call Sync to update client internal data
	if err = c.Sync(); err != nil {
		log.Fatalf("Sync failed: %v", err)
	}

	// Parse the search and make sure they're using one of the basic renderers
	if _, err := c.ParseSearchWithResponse(*query, []types.FilterRequest{}); err != nil {
		log.Fatalf("Parse request failed: %v", err)
	}

	// Now start the search
	sr := types.StartSearchRequest{
		SearchString: *query,
		SearchStart: start.Format(time.RFC3339Nano),
		SearchEnd: end.Format(time.RFC3339Nano),
		NoHistory: true,
		Preview:   true,
		Background: *background,
	}
	s, err := c.StartSearchEx(sr)
	if err != nil {
		log.Fatalf("Failed to start search: %v", err)
	}

	log.Println("Waiting for search to finish")
	// Wait for the search to be completed
	if err = c.WaitForSearch(s); err != nil {
		log.Fatalf("Failed to wait for search: %v", err)
	}

	log.Println("Getting available entry count")
	cnt, done, err := c.GetAvailableEntryCount(s)
	if err != nil {
		log.Fatalf("Failed to get available entry count: %v", err)
	} else if !done {
		log.Fatalf("Search is not done after waiting")
	}
	fmt.Printf("%d search entries available\n", cnt)

	c.DetachSearch(s)
	c.Close()
}

func getTimeDuration(trs string) (start time.Time, end time.Time, err error) {
	trs = strings.TrimRight(strings.TrimLeft(trs, " \t"), " \t\n")
	if len(trs) == 0 {
		err = errors.New("Invalid duration")
		return
	}
	if trs[0] != '-' {
		err = errors.New("Durations must be negative")
		return
	}
	var dur time.Duration
	dur, err = time.ParseDuration(trs)
	if err != nil {
		return
	}
	//we have a legit duration, lets make some start and end times
	end = time.Now()
	start = end.Add(dur)
	return
}

```

floren (approved):

PR #169, opened 2021-03-25
Add IPMI Ingester

djfritz:


traetox (changes requested):
I think we need to split the SEL logs.

floren (approved):

PR #170, opened 2021-03-26
WIP: Adding persistent buffer preprocessor

traetox:


traetox:
also includes some new feature overrides

Issue #158, opened 2021-02-17
HTTP ingester bind info with -verbose flag


traetox:
Request: `http ingester print the network port binding being used when fired off with the verbose flag`

Issue #171, opened 2021-03-29
HEC compatible listener block on HttpIngester


traetox:
Make it super easy to be 100% compatible with HEC, just drop in the required fields and rock on.

PR #173, opened 2021-03-30
Http fixup

traetox:
Cleaning up some debug functionality, removing some fmt usage, and adding a new listener type that should be fully compatible with the Splunk HEC system.

This way applications that have been designed against HEC can just point at our ingester and "just work"


traetox:
closes Issue #158 and issue #171

djfritz (approved):
It looks fine? LGTM

floren (approved):

PR #183, opened 2021-04-05
allow multiple targets in ipmi stanza, change ipmi poll rate

### LINKED ISSUE #175: IPMI ingester config should support multiple targets per block
most deployments will have a shared credential across many machines, blocks should support multiple targets.

Turn this: https://github.com/gravwell/gravwell/blob/dev/ingesters/IPMIIngester/config.go#L25 into a `[]string` so that i can have a list of targets.  Forcing a user to specify a whole new block with creds for every machine is tedious.

This way if you have 3 "enclaves" you can specify 3 blocks with a set of target IPs in each rather than a whole new block for every single machine.

### LINKED ISSUE #182: make ipmi poll rate 1 minute

djfritz:
Fixes #175
Fixes #182

### DIFF
--- a/ingesters/IPMIIngester/main.go
+++ b/ingesters/IPMIIngester/main.go
@@ -49,7 +49,7 @@ var (
 	ipmiConns map[string]*handlerConfig
 )
 
-const PERIOD = 10 * time.Second
+const PERIOD = time.Minute
 
 type handlerConfig struct {
 	target           string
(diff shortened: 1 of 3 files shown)


ingesters/IPMIIngester/main.go
```diff
@@ -49,7 +49,7 @@ var (
 	ipmiConns map[string]*handlerConfig
 )
 
-const PERIOD = 10 * time.Second
+const PERIOD = time.Minute
```
traetox, line 52: any reason to make this configurable ?

traetox (commented):
i think having a solid default of 10 minutes is good, but it should be configurable.

ingesters/IPMIIngester/main.go
djfritz: agreed - update incoming

djfritz:
@traetox PTAL - also added description field on certain messages

traetox (approved):
LGTM

PR #186, opened 2021-04-07
Create client/types/kits package

floren:


client/types/kits/reader.go
```diff
@@ -0,0 +1,334 @@
...
+			return
+		} else if hsh != item.Hash {
+			err = fmt.Errorf("%s in kit is corrupted", item.Name)
+			return
+		}
+	}
+	// Do the signing verification if provided
+	if sigVerify != nil {
+		if err = sigVerify(m, s); err != nil {
+			err = ErrInvalidSignature
```
traetox, line 243: seems like this should hand back the sigVerify error wrapped so that the user can still get access to their returned error: `err = fmt.Errorf("%w - %v", err, ErrInvalidSignature)`

traetox (changes requested):
hand back a wrapped error on signature verify callback return

floren:
PTAL

PR #187, opened 2021-04-08
initial kitctl, supports only pack and unpack

floren:


kitctl/main.go
```diff
@@ -0,0 +1,382 @@
...
+
+	switch args[0] {
+	case "unpack":
+		// Unpack the given kit into the current directory
+		unpackKit(args[1:])
+	case "pack":
+		// Pack the current directory into a kit
+		packKit(args[1:])
+	case "info":
+		// Information about the kit in the current directory
```
traetox, line 47: can we throw an error here about "no ready yet" rather than silently do nothing.  Just in case someone tries to use the dev branch.

traetox (changes requested):
also, is it possible to add tests on these?

traetox (approved):

PR #193, opened 2021-04-13
Make kitctl way more intelligent

floren:
Now breaks out far more fields when it unpacks. To properly deal with templates, we had to include the data type the GUI uses, so we can extract the query into its own file.


traetox (approved):
LGTM

PR #197, opened 2021-04-15
fix persistent buffer preprocessor which cannot be used on windows

floren:


traetox (changes requested):
rather than duplicate all this code, lets keep the universal in and then on default switch statements  call a sub that is implemented with build tags.

djfritz (approved):
LGTM

floren:
PTAL

traetox (approved):
LGTM

PR #199, opened 2021-04-19
bounce ipmi connections when we can't read

djfritz:


traetox (changes requested):
the handler config context isn't actually used by any of the connection routines.

When a quit signal lands it cancels the context then syncs and exits immediately, i think the routines should have a sync.WaitGroup and the context should be in a select group with a time.Ticker.  This would do a few things:

1. The sample interval would be a more consistent 10 minutes (even on delayed responses).
2. The routines would actually exit properly on system exit
3. You wouldn't run the risk of racing the input /ingest output on exit.

djfritz:
talked offline - this should be good as is

traetox (approved):
LGTM

PR #213, opened 2021-05-10
add DeleteKitEx function

floren:


client/kit.go
```diff
@@ -147,6 +149,39 @@ func (c *Client) DeleteKit(id string) (err error) {
...
+	resp, err = c.methodRequestURL(http.MethodDelete, kitIdUrl(id), ``, nil)
+	if err != nil {
+		// this means we weren't able to get a request to the server, return the error
+		return []types.SourcedKitItem{}, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		// There are basically two kinds of errors:
+		// 1. Kit items have been modified; body contains a list of modified items
+		// 2. Other errors (kit doesn't exist, malformed ID, etc.)
+		body, err := ioutil.ReadAll(resp.Body)
```
traetox, line 167: why do this over handing the body resp.Body into a json umarshaller?  The unmarshaller will bail i there is a problem, here you will attempt to read 1.21 exabytes if the server tries to give it to you.

traetox (commented):
ReadAll always makes me nervous.

client/kit.go
floren: So I can write the body into the error message if the unmarshal fails, per line 177. As the guy who wrote the server code, I'm not going to give you 1.21 exabytes.

traetox: touche!

THANK YOU FLOREN!

traetox (approved):
LGTM

PR #228, opened 2021-06-03
Remove deprecated SortRequired field and add EntryCountValid field.

### LINKED ISSUE #227: Remove SortRequired field from types.RenderModuleInfo

floren:
Removes SortRequired from types.RenderModuleInfo, as it is no longer used.
Adds the EntryCountValid field to types.BaseResponse so renderers can inform
clients whether or not the EntryCount number represents the actual number
of results which can be pulled back or not. (EntryCount gives the number
of entries which came *into* the renderer, but since some renderers do further
processing, it doesn't always indicate how many results are available).

Closes #227


traetox (approved):
LGTM

PR #234, opened 2021-06-22
Implement custom marshaller for types.Element 

floren:
We don't want to send a "null" value, so this implements a little bit of logic to get around that.


djfritz (approved):
LGTM

traetox (approved):
LGTM

PR #235, opened 2021-06-22
Revert "Implement custom marshaller for types.Element "

floren:
Reverts gravwell/gravwell#234

We didn't want to do that after all, misunderstanding on my part.


djfritz (approved):
LGTM

PR #236, opened 2021-06-24
Remove outdated todos

floren:
The first TODO comment is essentially handled by the call to `isLetterNumberPeriod` immediately preceding. As for validating macros, there's just not a lot you can do to validate macro IDs except checking if they're 0, given that we don't have a way to query the system for which macros actually exist.


djfritz (approved):
LGTM

PR #250, opened 2021-08-17
adding close method to resource update

traetox:
one more helper method


traetox:
force merging because i was told i could

PR #261, opened 2021-09-13
windows service shutdown controls

traetox:
changing the order of signaling and shutdown to see if it fixes the issue of not releasing the cache lock in time.

So apparently windows doesn't consider a service stopped when the process exits it considers it stopped when the state signal changes.  This seems entirely reasonable.  Definitely trust the signal... definitely don't actually check if the process quit.  A system service with an active process that declares itself stopped is totally ok... yep... won't be abuse... won't cause problems...

## THIS IS FINE


djfritz (approved):
LGTM

PR #270, opened 2021-09-27
Change behavior of notification functions to handle random notificati…

floren:
…on IDs

Notification IDs are now random instead of monotonically
increasing, so you can't say "give me all notifications
after #3". Instead, we set a timestamp in the request,
e.g. ?after=2006-01-02T15:04:05.999999999Z07. This change makes the
MyNewNotification* functions properly handle this.


traetox (approved):
LGTM

Issue #277, opened 2021-10-11
Inclusion of scriggo breaks 32-bit builds due to upstream bug


lkarlslund:
https://github.com/gravwell/gravwell/blob/b470b361be965b7d051bb6fc6a0c26575c5920ec/ingest/processors/plugin/plugin.go#L22

Bug has been reported in the scriggo repo here: 

https://github.com/open2b/scriggo/issues/901

PR #278, opened 2021-10-12
updating plugin system to stub itself out when building for a 32bit arch

traetox:
Addresses https://github.com/gravwell/gravwell/issues/277


ingest/processors/plugin/packages.go
```diff
@@ -1,6 +1,7 @@
+// +build linux,go1.16,!go1.17,!386,!arm,!mips,!mipsle,!s390x
+
 // Code generated by scriggo command. DO NOT EDIT.
-//go:build linux && go1.16 && !go1.17
-// +build linux,go1.16,!go1.17
+//go:build linux && go1.16 && !go1.17 && linux,go1.16 && !go1.17 && !386 && !arm && !mips && !mipsle && !s390x
```
floren, line 4: this should probably also be at the top of the file, or removed until we decide to support go 1.17 officially

PR #283, opened 2021-10-18
timegrinder additional format and a tool to help test formats

traetox:
adding the bind time format to timegrinder and adding a little tool to test timestamp hits.  Also fixing some bad return values that are used in timegrinder debugging


djfritz (approved):
LGTM

PR #286, opened 2021-10-20
Generator work

traetox:
updating the generator base package so that anything that uses it can also throw line broken values at TCP socket too

### DIFF
--- a/generators/base/base.go
+++ b/generators/base/base.go
@@ -174,10 +174,8 @@ func NewIngestMuxer(name, guid string, gc GeneratorConfig, to time.Duration) (co
 		if conn, err = newRawConn(gc, to); err == nil {
 			src, err = conn.SourceIP()
 		}
-		fmt.Printf("%T %v\n", conn, src)
 		return
 	}
-	fmt.Println("not in raw")
 	if name == `` {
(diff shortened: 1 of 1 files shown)


generators/base/base.go
```diff
@@ -138,11 +153,31 @@ func GetGeneratorConfig(defaultTag string) (gc GeneratorConfig, err error) {
...
+
+func NewIngestMuxer(name, guid string, gc GeneratorConfig, to time.Duration) (conn GeneratorConn, src net.IP, err error) {
 	if !gc.ok {
 		err = errors.New("config is invalid")
 		return
+	} else if gc.modeRaw {
+		if conn, err = newRawConn(gc, to); err == nil {
+			src, err = conn.SourceIP()
+		}
+		fmt.Printf("%T %v\n", conn, src)
```
floren, line 177: debug print

generators/base/base.go
```diff
@@ -138,11 +153,31 @@ func GetGeneratorConfig(defaultTag string) (gc GeneratorConfig, err error) {
...
+		return
 	}
+	fmt.Println("not in raw")
```
floren, line 180: debug print

PR #297, opened 2021-11-01
WIP: Rework how ingesters log.

floren:
Instead of having a bunch of logic for the ingest muxer to send logs to the
gravwell tag, instead just provide convenience functions which write to
the Logger which was passed in the config. In turn, that logger can set
the ingest muxer as a *relay*, which writes the log messages to the gravwell
tag. This has the side effect of getting *all* log messages into the gravwell
tag, provided you don't fill up the (generous) buffer before finally connecting
to an indexer.

Note: This is not yet ready, the rest of the ingesters need to be walked to deal with things properly. networkLog and SimpleRelay demonstrate what needs doing, though.

### DIFF
--- a/ingest/gravwell_log.go
+++ b/ingest/gravwell_log.go
@@ -9,26 +9,11 @@
 package ingest
 
 import (
-	"fmt"
-	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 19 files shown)


djfritz (approved):
LGTM

PR #331, opened 2022-01-06
updating abac to have default with overrides the same way we do with …

traetox:
…tags, also moving some of the capability checking code here

### DIFF
--- a/client/types/abac.go
+++ b/client/types/abac.go
@@ -88,27 +88,36 @@ const (
 	DefaultDeny  = true
 )
 
+// ABACRules is the main structure that holds default stats and overrides for for API and tag access
+// the Capabilities and Tags sub structures handle access independently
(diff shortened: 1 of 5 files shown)


client/types/abac.go
```diff
@@ -69,6 +69,7 @@ const (
 	SystemInfoRead    Capability = 44
 	TokenRead         Capability = 45
 	TokenWrite        Capability = 46
+	_maxCap           Capability = 47 //REMINDER - when adding capabilities, make sure to enpand this number
```
djfritz, line 72:
```suggestion
	_maxCap           Capability = 47 //REMINDER - when adding capabilities, make sure to update this number
```

djfritz (changes requested):

djfritz (approved):
LGTM

PR #334, opened 2022-01-13
laying out tenant level authentication over the ingest protocol

traetox:


generators/jsonGenerator/gen.go
```diff
@@ -14,7 +14,8 @@ import (
 	"time"
 
 	rd "github.com/Pallinder/go-randomdata"
-	"github.com/bet365/jingo"
+	"github.com/goccy/go-json"
+	//"github.com/bet365/jingo"
```
floren, line 18: Do we want to leave all the commented stuff in here?

floren (approved):

PR #335, opened 2022-01-13
adding tag name to info log when watching files

traetox:
the info log when the file follower begins tracking a new file will now also output the tag being attached to the followed file.

this is so that its easier to debug file follower states from the gravwell tag


filewatch/followers_test.go
```diff
@@ -249,3 +251,9 @@ func (h *trackingLH) HandleLog(b []byte, ts time.Time) error {
 	}
 	return nil
 }
+
+type testTagger struct{}
+
+func (tt testTagger) Tag() string {
+	return `defualt`
```
floren, line 258: spelling

floren (changes requested):

PR #351, opened 2022-02-05
working on file follower startup mode where it can be more intelligent about mass import

traetox:
Also merged in master to get some of the other timegrinder fixes in.

BOTH reviewers should review before merging please.


djfritz:
This seems fine, although explicitly setting GOMAXPROCS=1 in file follower somehow feels like a bad move (though I can't articulate why)...

djfritz (approved):
LGTM

filewatch/filters.go
```diff
@@ -500,6 +500,31 @@ func (f *FilterManager) launchFollowers(fpath string, deleteState bool) (ok bool
 	return
 }
 
+func (f *FilterManager) checkState(wf watchedFile) (hasWork bool, err error) {
+	//get base dir
+	fname := filepath.Base(wf.pth)
+	fdir := filepath.Dir(wf.pth)
+	var si *int64
+	//swing through all filters and launch a follower for each one that matches
```
floren, line 508: comment says "launch a follower" which is inaccurate. Would appreciate a comment at the top of the function, too.

filewatch/filewatch.go
```diff
@@ -248,28 +260,41 @@ func (wm *WatchManager) Start() error {
...
+	//generate list of files that could be processed
+	//this could be MANNY files if people are doing some mass ingest
+	toProcess, err := wm.getWatchedFileList()
+	if err != nil {
+		return err
+	}
+	return wm.fman.LoadFileList(toProcess)
+}
+
+func (wm *WatchManager) Catchup(qc chan os.Signal) (bool, error) {
```
floren, line 272: Need a comment for this: what does the boolean mean.

filewatch/filters.go
```diff
@@ -558,6 +583,81 @@ func (f *FilterManager) matchFile(mtchs []string, fname string) (matched bool) {
 	return
 }
 
+func (f *FilterManager) CatchupFile(wf watchedFile, qc chan os.Signal) (bool, error) {
```
floren, line 586: Comment this: what is the bool?

filewatch/filters.go
```diff
@@ -558,6 +583,81 @@ func (f *FilterManager) matchFile(mtchs []string, fname string) (matched bool) {
...
+	} else if isRename {
+		return true, nil //just a file renaming, continue
+	}
+
+	//get base dir
+	fname := filepath.Base(wf.pth)
+	fdir := filepath.Dir(wf.pth)
+	var si *int64
+
+	//swing through all filters and launch a follower for each one that matches
```
floren, line 608: "launch a follower" implies a goroutine, but this is all synchronous

filewatch/filters.go
```diff
@@ -558,6 +583,81 @@ func (f *FilterManager) matchFile(mtchs []string, fname string) (matched bool) {
...
+	for i, v := range f.filters {
+		//check base directory and pattern match
+		if v.loc != fdir || !f.matchFile(v.mtchs, fname) {
+			continue
+		}
+		//see if we have state information for this file
+		si = f.seekInfo(v.bname, wf.pth)
```
floren, line 615: could you use your new checkState function here?

filewatch/followers.go
```diff
@@ -130,6 +130,39 @@ func (f *follower) FileId() FileId {
...
+func (f *follower) Sync(qc chan os.Signal) (bool, error) {
+	f.mtx.Lock()
+	defer f.mtx.Unlock()
+	if f.lnr == nil {
+		return false, ErrNotReady
+	}
+	if f.abortCh != nil || f.running != 0 {
+		return false, ErrAlreadyStarted
+	}
+	for {
```
floren, line 142: This code duplicates a lot of processLines, can we just use that or is there special sauce?

timegrinder/timeGrinder.go
```diff
@@ -186,15 +186,14 @@ func New(c Config) (tg *TimeGrinder, err error) {
 }
 
 func (tg *TimeGrinder) SetFormatOverride(v string) (err error) {
-	if v == `` {
```
floren, line 189: Didn't we just commit this change a week or two ago in the opposite direction?

floren (changes requested):

filewatch/followers.go
traetox: the quit channel is kind of unique, i made the judgement call that the duplication was better than trying to handle having a quitchan on process lines

timegrinder/timeGrinder.go
traetox: this matches the current state of master, which is what got the fix.  I don't think we brought the fix into dev, there was a little optimization on dev that I also brought in here.

traetox:
@floren PTAL

floren (approved):

PR #352, opened 2022-02-07
Migrate work

traetox:


ingesters/migrate/state.go
```diff
@@ -67,6 +68,7 @@ func loadExistingState(pth string) (ents []StateEntry, err error) {
 				break
 			}
 			fin.Close()
+			fmt.Println("Failed to decode", err)
```
floren, line 71: left-over debug?

traetox: yep, addressed

floren (approved):

PR #353, opened 2022-02-09
adding preprocessor example that can force dates

traetox:
some old garbage timestamps don't have dates or years, this lets you fix that by hand.


djfritz (approved):
LGTM

PR #367, opened 2022-03-09
add some much needed stringification

traetox:


floren (approved):
good enough to start with

PR #370, opened 2022-03-10
don't exit on overflow, just nag

djfritz:


filewatch/filewatch.go
```diff
@@ -344,23 +344,15 @@ watchRoutine:
...
-				// ignore the error but let us know please
-				d = []byte(fmt.Sprintf("could not read from %v", MAX_QUEUED_EVENTS_PATH))
-			}
-			wm.logger.Error("Filesystem notification error. Increase the queued events kernel parameter or decrease the number of tracked files.", log.KVErr(err), log.KV("max_queued_events", string(d)), log.KV("help", "https://docs.gravwell.io/#!ingesters/file_follow.md"))
-
-			wm.mtx.Lock()
-			defer wm.mtx.Unlock()
-			if wm.fman != nil {
-				wm.fman.Close()
+			if err == fsnotify.ErrEventOverflow {
```
traetox, line 347: clear err here too, this isn't an error we are handling so it should not be returned.

traetox (changes requested):
small change

traetox (approved):
LGTM

PR #374, opened 2022-03-15
Conf.d work

traetox:


ingesters/IPMIIngester/ipmi.conf
```diff
@@ -1,4 +1,5 @@
 [Global]
+Ingester-UUID="91cf685f-9558-455c-a7c0-e2aab3c4375f"
```
floren, line 2: Get rid of this

ingest/config/validate/validate.go
```diff
@@ -23,7 +23,13 @@ var (
...
 )
 
-func ValidateConfig(fnc interface{}, pth string) {
+// ValidateConfig will take a configuration handling function and two paths.
+// The first path is the base config file, and the second path points at the conf.d directory.
+// The config handling function will take those two paths and process a config and return two values.
+// The first value is an opaque object (hence the reflect voodoo).  The second value is an error object
+// the point of this function is to make it easy for ingester writers to just hand in their GetConfig function
+// and the two paths and get a "go/no go" on the configurations
+func ValidateConfig(fnc interface{}, pth, confdPath string) {
```
floren, line 32: So should we really have ValidateConfig calling os.Exit? Is that appropriate in a library? Shouldn't we return an error instead and let the caller decide?

traetox: i was trying to make it easy on ingester writers, if we go with a return code every ingester will need to ad something like the following:
```
if exit, code := validate.ValidateConfig(GetConfig, foo, bar); if exit {
   os.Exit(code)
}
```

which is fine, but it means more imports and a bunch more change to existing ingesters.

floren: We can leave it as-is for now, but if we ever want to e.g. add configuration via the web UI, which means validation on the frontend, we'll have to get rid of the os.Exit calls.

PR #389, opened 2022-04-12
Add flow support to kitctl.

floren:
Also update the way we name automations as we unpack to apply the name given in the kit rather than the Name field, to align with the webserver behavior

### DIFF
--- a/kitctl/pack.go
+++ b/kitctl/pack.go
@@ -394,25 +394,30 @@ func readPlaybook(dir, name string) (x types.Playbook, err error) {
  * Scheduled Search
  **************************************************************************/
 
-func writeScheduledSearch(dir string, x kits.PackedScheduledSearch) error {
+func writeScheduledSearch(dir string, name string, x kits.PackedScheduledSearch) error {
 	// Make sure the parent exists
 	p := filepath.Join(dir, "scheduled")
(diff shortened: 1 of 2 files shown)


kitctl/pack.go
```diff
@@ -394,25 +394,30 @@ func readPlaybook(dir, name string) (x types.Playbook, err error) {
...
-	// Now drop three files: .meta, .search, and .script
-	metaPath := filepath.Join(p, fmt.Sprintf("%v.meta", x.Name))
-	searchPath := filepath.Join(p, fmt.Sprintf("%v.search", x.Name))
-	scriptPath := filepath.Join(p, fmt.Sprintf("%v.script", x.Name))
+	// Now drop files: .meta, .search, .flow, and .script
+	metaPath := filepath.Join(p, fmt.Sprintf("%v.meta", name))
+	searchPath := filepath.Join(p, fmt.Sprintf("%v.search", name))
+	scriptPath := filepath.Join(p, fmt.Sprintf("%v.script", name))
+	flowPath := filepath.Join(p, fmt.Sprintf("%v.flow", name))
 	if err := ioutil.WriteFile(searchPath, []byte(x.SearchString), 0644); err != nil {
```
traetox, line 409: seems like this should only drop the one that is populated?

traetox (approved):
LGTM

PR #397, opened 2022-05-04
Add RegexListener to SimpleRelay ingester.

floren:
A RegexListener specifies a regular expression which marks the split between entries.
The regular expression can contain "prefix" and "suffix" capture groups which will be
kept with the appropriate entry, thus given the following regex:

	Regex=`(?P<suffix>END)|(?P<prefix>\d{4}-\d{1,2}-\d{1,2})`

and the following data input:

	2022-03-30 message 1 END|2022-04-01 message 2 END

you will get two entries:

	2022-03-30 message 1 END
	2022-04-01 message 2 END

### DIFF
--- a/ingesters/SimpleRelay/config.go
+++ b/ingesters/SimpleRelay/config.go
@@ -151,7 +151,7 @@ func verifyConfig(c *cfgType) error {
 		}
 	}
 	for k, v := range c.RegexListener {
-		if err := v.base.Validate(); err != nil {
+		if err := v.Validate(); err != nil {
 			return fmt.Errorf("RegexListener %s configuration error: %v", k, err)
 		}
 		if len(v.Tag_Name) == 0 {
--- a/ingesters/SimpleRelay/config.go
+++ b/ingesters/SimpleRelay/config.go
@@ -59,19 +59,21 @@ type base struct {
 }
 
 type cfgReadType struct {
-	Global       config.IngestConfig
-	Listener     map[string]*listener
-	JSONListener map[string]*jsonListener
-	Preprocessor processors.ProcessorConfig
-	TimeFormat   config.CustomTimeFormat
+	Global        config.IngestConfig
+	Listener      map[string]*listener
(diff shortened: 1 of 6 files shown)


ingesters/SimpleRelay/config.go
```diff
@@ -147,6 +150,33 @@ func verifyConfig(c *cfgType) error {
 			return fmt.Errorf("Listener %s preprocessor invalid: %v", k, err)
 		}
 	}
+	for k, v := range c.RegexListener {
```
traetox, line 153: should this also attempt to compile the regex?  seems like we should catch a bad regex here.

traetox: looks like it just needs to call Validate on v?

ingesters/SimpleRelay/regexHandlers.go
```diff
@@ -0,0 +1,327 @@
...
+	var prefix, suffix []byte
+	var done bool
+	var n int
+	for !done {
+		// we build the entry data into this buffer.
+		// make a new one on each loop so we're safe to call Bytes() at the end
+		// read a bunch of bytes off the connection
+		if n, err = bio.Read(rd); err != nil && err != io.EOF {
+			lg.Error("error reading from regex connection", log.KVErr(err))
+			return
```
traetox, line 251: if you have half an entry in the buffer and then get a timeout we lose that half buffered entry, seems like we should throw it.  Same for if you have a complete entry and are waiting on the next delimited entry and get a timeout or network break.  It would lose that buffered entry.

Seems like any "exit" should attempt to process the existing buffer

traetox (changes requested):
this will drop the last entry or partial entries on any read error, so if the connection drops won't we drop whats in the buffer?

Also catch bad regexes in the parse phase.

traetox (commented):
2 more nits/questions

traetox (approved):
thank you, this looks great.

PR #398, opened 2022-05-06
Use the reingest API for IngestEntries

floren:
IngestEntries previously used /api/ingest/json, which has a 24MB
default limit.  This means you could only ingest 24MB of entries at
a time or else you'd get an error from the webserver. This change
instead uses the reimport API, which can also handle search entries,
but does it via a streaming multipart request and goes around the
size limits.

Also adds a function to fetch the GUI Settings object.


traetox (approved):
LGTM

PR #400, opened 2022-05-16
adding a lock around the ingester state so that we can protect those …

traetox:
…internal maps, we also have a copy system now that will ensure we don't have a race that could cause a fault when handing maps around.


djfritz (approved):
LGTM

PR #411, opened 2022-06-07
add trimLength, trimPathLength, to RFC5424 message generator

djfritz:


ingest/log/logging.go
```diff
@@ -658,3 +665,22 @@ func checkName(v string) (err error) {
...
+// trimPathLength will trim the input path to no more than i bytes of the
+// basename of input. For example, "KafkaFederator/kafkaWriter.go:352" will be
+// trimmed to "kafkaWriter.go:352"
+func trimPathLength(i int, input string) string {
+	if len(input) <= i {
+		return input
+	}
+	base := filepath.Base(input)
+
+	return trimLength(i, base)
```
traetox, line 678: the last chunk is the most important here.  For example if it is:
`foobarbaz.go:244` i would rather have `barbaz.go:244` over `foobarbaz.go` so that we have a chance of deciphering what file and where.

traetox (changes requested):
on the path one even after a filepath.Base I think we should grab the trailing characters over the leading.

PR #412, opened 2022-06-07
Updates to migrate command

floren:
Replace old migrate command with the new interactive one.

Allow users to specify which timestamp to start from when migrating Splunk data.


djfritz (approved):
LGTM

PR #418, opened 2022-06-13
small fmt on client error and change to Template API

traetox:
update error constant on client and change the templates API to get the entire wireTemplate back


djfritz (approved):
LGTM

PR #421, opened 2022-06-14
Update google pubsub to limit needed privs

traetox:
This updates the Google pubsub ingester so that it doesn't needs as aggressive privs.  It now walks a series of operations trying to get a handle on a subscription that require the least privs.

For example, if the subscription config is set, then it will try to use that, if it can't get what it needs it fails.

If it is NOT set then it tries to test it and create it (requiring more privs).  Using this escalating procedure its possible to configure a security token that doesn't really let the ingester do THAT much.

But if you DO give it a ton of privs it will set stuff up itself.


floren (approved):
good idea

Issue #415, opened 2022-06-08
client DeleteUserFromGroup is using the wrong path


traetox:
Should be a `DELETE` on `/api/users/<uid>/group/<gid>` but it is trying to do a `DELETE` on the add group URL.

traetox:
done

Issue #414, opened 2022-06-08
client is missing GetDefaultSearchGroup method


traetox:
there is a method to set the default search group but not one to get it.  Corresponds to a `GET` on `/api/users/<uid>/searchgroup`

PR #423, opened 2022-06-17
Adding in methods to work with the DefaultSearchGroup that were missing

traetox:
Also fixing the URL and method used to Remove a user from a group, it
was broken.

Closes issues #415 and #414


djfritz (approved):
LGTM

PR #427, opened 2022-06-27
Fixes #420: Fatal log call depth.

### LINKED ISSUE #420: Fatal call on logging system is not resolving call dept correctly.
Any call on `Fatal` resolves to `log/logging.go:411` instead of the appropriate caller location

floren:
Calling FatalCode was messing up the call depth. It is not a complex bit of
code, so instead we just duplicate the logic into Fatal.


traetox (approved):
LGTM

PR #433, opened 2022-07-05
Make ConfigureMail call do a PUT instead of a POST

john-floren-gravwell:
Also, make putStaticURL take a recvObj argument, because PUT can return a body


kris-watts-gravwell (approved):
LGTM

PR #434, opened 2022-07-05
Add UserMailResponse struct and GetUserFileDetails function.

john-floren-gravwell:
UserMailResponse is what the server sends back after a successful email submission.

GetUserFileDetails lets you get information about a single file, rather than fetching the contents.


kris-watts-gravwell (approved):
LGTM

PR #437, opened 2022-07-06
Revert "Change the URLs used for GetUserFile and GetUserFileDetails"

john-floren-gravwell:
This reverts commit a67d1b5d71319dd0ec065cbf3344c2eb6ff47e61.

The change in URLs would have broken playbooks, and that's just not worth the squeeze.


david-fritz-gravwell (approved):
LGTM

PR #444, opened 2022-07-14
API: update CloneDashboard

john-floren-gravwell:
The backend now returns a full dashboard, not just an ID


david-fritz-gravwell (approved):
LGTM

Issue #344, opened 2022-01-27
`failed to configure stream` on ingester connection


traetox:
## Background

I am using a tip of dev ingester `b93eeec7638f06b1d63f64d4e2cd8f6ac19aebee` pointed at 4.2.8-2 prod release.

When starting up the ingester it will fail to complete the handshake a few times (0-10) on startup.  It eventually succeeds.

Cache is active, happens on pipe and cleartext targets.  Ingester is colocated with the indexer.

Here is the output with the `-v` flag:

```
<14>1 2022-01-27T20:37:13.904357Z localhost httpingester - ingest/muxer.go:1394 [gw@1 indexer="tcp://127.0.0.1:4023" ingester="httppost" ingesteruuid="941af247-e3af-494b-8ccf-398f0636a81c"] initializing connection
<14>1 2022-01-27T20:37:13.919955Z localhost httpingester - ingest/muxer.go:1412 [gw@1 indexer="tcp://127.0.0.1:4023" ingester="httppost" ingesteruuid="941af247-e3af-494b-8ccf-398f0636a81c"] connection established, completing negotiation and requesting approval to ingest
<12>1 2022-01-27T20:37:13.938732Z localhost httpingester - ingest/muxer.go:1455 [gw@1 ingester="httppost" ingesteruuid="941af247-e3af-494b-8ccf-398f0636a81c" indexer="tcp://127.0.0.1:4023" error="EOF"] failed to configure stream
<14>1 2022-01-27T20:37:13.939296Z localhost httpingester - ingest/muxer.go:1394 [gw@1 indexer="tcp://127.0.0.1:4023" ingester="httppost" ingesteruuid="941af247-e3af-494b-8ccf-398f0636a81c"] initializing connection
<14>1 2022-01-27T20:37:13.967492Z localhost httpingester - ingest/muxer.go:1412 [gw@1 indexer="tcp://127.0.0.1:4023" ingester="httppost" ingesteruuid="941af247-e3af-494b-8ccf-398f0636a81c"] connection established, completing negotiation and requesting approval to ingest
<12>1 2022-01-27T20:37:13.971681Z localhost httpingester - ingest/muxer.go:1455 [gw@1 ingester="httppost" ingesteruuid="941af247-e3af-494b-8ccf-398f0636a81c" indexer="tcp://127.0.0.1:4023" error="EOF"] failed to configure stream
[...]

PR #447, opened 2022-07-19
improve Error logging

kris-watts-gravwell:
improving the error logging so we can better understand why configureStream sometimes gets an EOF

Should also allow for actually debugging https://github.com/gravwell/gravwell/issues/344


david-fritz-gravwell (approved):
LGTM

Issue #428, opened 2022-06-27
convert winevent ingester over the the structured logging format


traetox:
winevent needs a logger and we need to get rid of the infoout, errorout nonsense.

kris-watts-gravwell:
done and done

PR #449, opened 2022-07-20
standardizing winevent internal logs

kris-watts-gravwell:
working on converting winevent over to the standard logging structure

Had to add a new relay type on the `ingest/log` package.

closes https://github.com/gravwell/gravwell/issues/428

### DIFF
--- a/ingesters/winevents/main.go
+++ b/ingesters/winevents/main.go
@@ -30,6 +30,7 @@ import (
 
 const (
 	serviceName       = `GravwellEvents`
+	appName           = `winevent`
 	defaultConfigPath = `gravwell\eventlog\config.cfg`
 )
 
@@ -40,7 +41,6 @@ var (
 
 	confLoc string
 	verbose bool
-	elog    debug.Log
 	lg      *log.Logger
 )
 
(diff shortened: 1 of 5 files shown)


ingesters/winevents/main.go
```diff
@@ -74,27 +72,34 @@ func main() {
...
 	}
-	if !inter {
+	if inter {
+		lg = log.New(os.Stdout)
+	} else {
+		lg = log.NewDiscardLogger()
 		e, err := eventlog.Open(serviceName)
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "Failed to get event log handle: %v\n", err)
+			lg.Error("failed to get event log handle", log.KVErr(err))
```
john-floren-gravwell, line 81: We've just set lg to a NewDiscardLogger, so this message is guaranteed to be thrown away every time, right? I think printing to stderr is a better idea.

ingesters/winevents/main.go
```diff
@@ -169,46 +174,27 @@ func debugout(format string, args ...interface{}) {
...
 
-func warnout(format string, args ...interface{}) {
-	if warnW != nil {
-		warnW(format, args...)
+// levelLogger implements the log.LevelRelay interface
+func (l levelLogger) WriteLog(lvl log.Level, ts time.Time, msg []byte) error {
+	switch lvl {
+	case log.DEBUG:
+		if verbose {
+			fmt.Fprintln(os.Stdout, string(msg))
```
john-floren-gravwell, line 186: Stdout or Stderr?

kris-watts-gravwell: i think for pure debug it should be stdout.  I think the O.G. choice of stderr was incorrect

ingesters/winevents/main.go
kris-watts-gravwell: agreed, fix incoming

john-floren-gravwell (approved):

PR #453, opened 2022-07-27
Fix scheduled search & flow code to be more RESTful

john-floren-gravwell:
Creating a flow or scheduled search/script will now return the full object in the response.

Also merges dev into api-updates.


kris-watts-gravwell (approved):
LGTM

PR #460, opened 2022-08-02
Add WordOffsets field to ExploreResult struct.

john-floren-gravwell:
The WordOffsets array contains [2]int arrays, each specifying the
absolute offsets of a word within the DATA field of the associated
entry. This can be used to provide clickable filters in the UI.


kris-watts-gravwell (approved):
LGTM

PR #461, opened 2022-08-04
migrate fixes

kris-watts-gravwell:
fixing typo in config type that was referencing the wrong config object and removing linux build constraints.

This should work just fine in windows


david-fritz-gravwell (approved):
LGTM

PR #464, opened 2022-08-09
adding a meta structure for license usage reporting

kris-watts-gravwell:
Just add another structure so that we can handle mixed license usage


david-fritz-gravwell (approved):
LGTM

PR #466, opened 2022-08-10
Add the Ping() function on the Search type

john-floren-gravwell:
If you open a search and then don't interact with it for a long time, the websockets will be closed by the webserver.

If you want to do this, call Ping() periodically.


david-fritz-gravwell (approved):
LGTM

PR #470, opened 2022-08-16
feature: Add the first pass at an Azure Event Hubs ingester.

john-floren-gravwell:


ingesters/AzureEventHubs/main.go
```diff
@@ -0,0 +1,423 @@
...
+	if !ok {
+		lg.FatalCode(0, "could not read ingester UUID")
+	}
+	ingestConfig := ingest.UniformMuxerConfig{
+		IngestStreamConfig: cfg.Global.IngestStreamConfig,
+		Destinations:       conns,
+		Tags:               tags,
+		Auth:               cfg.Secret(),
+		Logger:             lg,
+		IngesterName:       "AzureEventHubs",
```
kris-watts-gravwell, line 134: this should match appName IMHO

ingesters/AzureEventHubs/main.go
```diff
@@ -0,0 +1,423 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Set up the disk persistence object
+	diskPersist, err := persist.NewFilePersister(cfg.Global.State_Store_Location)
+	if err != nil {
+		lg.FatalCode(0, "failed to set up state persistence", log.KVErr(err))
+	}
+	// Just to make sure: chmod the directory, because the code wants to set it 777 (ugh)
+	if err := os.Chmod(cfg.Global.State_Store_Location, 0700); err != nil {
```
kris-watts-gravwell, line 181: ouch...

ingesters/AzureEventHubs/main.go
```diff
@@ -0,0 +1,423 @@
...
+	// Tell every event handler to close
+	for _, h := range listeners {
+		cctx, _ := context.WithTimeout(ctx, 2*time.Second)
+		h.Close(cctx)
+	}
+
+	// Tell our goroutines to bail out
+	close(quitSig)
+
+	// Write out persistence info one last time by hand.
```
kris-watts-gravwell, line 394: looks like there is a race here.  The "ticker/persister" routine could be writing to the disk at the same time as you.

I would reccomend an "exit" channel on that routine so that when you hit here you can wait for that routine to fully exit before performing the final write back to the disk.

kris-watts-gravwell: maybe not, the mem and disk persister have mutexes internally.

kris-watts-gravwell (changes requested):
few comments, only change would be the appName matching the ingester name.

kris-watts-gravwell (approved):
LGTM

PR #472, opened 2022-08-17
working on getting dates to be respected

kris-watts-gravwell:
This fixes the problem where the splunk migration was not respecting the the ingest-up-to stuff in splunk configs.

However there is still a problem where its just jumping a year in reverse for no reason.

For example:  i set `Ingest-From-Unix-Time=1659312000` which is `2022-08-01 00:00:00 +0000 UTC`, the picker menu shows the correct starting time, but when i launch the job it decides to start from `2021-08-01 00:00:00 +0000 UTC`.  I have no idea why it jumps a year in reverse.

Please fix this and issue a PR to this PR


kris-watts-gravwell:
I think this is ready @john-floren-gravwell PTAL

john-floren-gravwell (approved):

PR #481, opened 2022-08-30
chore: Merge master -> dev after 5.0.6 

ashnwade:


ingesters/version/version.go
```diff
@@ -16,12 +16,12 @@ import (
 
 const (
 	MajorVersion = 5
-	MinorVersion = 1
```
kris-watts-gravwell, line 19: this needs to be manually fixed up.  We can do it after the merge

kris-watts-gravwell (approved):
LGTM - just need to fix up versions afterwords.

PR #489, opened 2022-09-20
this has some tests that show issues with how we do this conversion

kris-watts-gravwell:
Basically because of the "clean" phase of the map fixup we introduce some ambiguity and failures.

I think we need to remove the clean phase and just specify the headers in the exact form that corelight throws them, we don't emit the headers anywhere, so its OK to keep them as is.

Test included.

example logs here: https://docs.zeek.org/en/master/logs/index.html

### DIFF
--- a/ingest/processors/corelight.go
+++ b/ingest/processors/corelight.go
@@ -139,20 +139,7 @@ func (c *Corelight) processLine(s []byte) (tag string, ts time.Time, line []byte
 	return
 }
 
-func clean(ff map[string]interface{}) {
-	for k, v := range ff {
-		if len(k) == 0 {
-			continue
-		}
-		if bits := strings.Split(k, "."); len(bits) > 0 {
-			ff[bits[len(bits)-1]] = v
-		}
-	}
-	return
-}
-
 func (c *Corelight) process(mp map[string]interface{}, og []byte) (tag string, ts time.Time, line []byte) {
-	clean(mp)
 	var ok bool
 	var headers []string
 	if len(mp) == 0 {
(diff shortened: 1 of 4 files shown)


ingest/processors/corelight.go
```diff
@@ -162,7 +154,10 @@ func (c *Corelight) process(mp map[string]interface{}, og []byte) (tag string, t
 	} else if line, ok = emitLine(ts, headers, mp); !ok {
 		tag = defaultTag
 		line = og
+	} else {
+		line = bytes.TrimRight(line, "-\t\n")
```
john-floren-gravwell, line 158: Won't this trim off the last - column? Surely we want to keep that, if only for consistency's sake with files ingested natively as TSV

ingest/processors/corelight.go
```diff
@@ -191,7 +186,7 @@ func (c *Corelight) getTagTs(mp map[string]interface{}) (tag string, ts time.Tim
 func emitLine(ts time.Time, headers []string, mp map[string]interface{}) (line []byte, ok bool) {
 	bb := bytes.NewBuffer(nil)
 	var f64 float64
-	fmt.Fprintf(bb, "%.3f", float64(ts.UnixNano())/1000000000.0)
+	fmt.Fprintf(bb, "%.5f", float64(ts.UnixNano())/1000000000.0)
```
john-floren-gravwell, line 189: Looking at actual zeek logs, they appear to use 6 digits of precision:

```
1663714154.384267	CXGv4RcQXWSFXRzGd	192.168.0.77
```

ingest/processors/corelight.go
kris-watts-gravwell: fixed.

ingest/processors/corelight.go
kris-watts-gravwell: fixed

john-floren-gravwell (approved):

PR #490, opened 2022-09-20
Fix corelight conn log formatting.

john-floren-gravwell:
Also:

* Extract timestamps with timegrinder, just in case we have different time stamp formats from different systems
* Fix broken ftp tests


kris-watts-gravwell (approved):
LGTM

PR #492, opened 2022-09-21
adding ability to set a custom format and also adding tests for ssh

kris-watts-gravwell:


ingest/processors/corelight.go
```diff
@@ -87,22 +89,41 @@ func (c *Corelight) Config(v interface{}, tagger Tagger) (err error) {
...
 		tagName := c.Prefix + k
-		c.tagFields[tagName] = strings.Split(v, ",")
+		c.tagFields[tagName] = cleanHeaders(strings.Split(v, ","))
 		if tv, err := c.tg.NegotiateTag(tagName); err != nil {
 			return err
 		} else {
 			c.tags[tagName] = tv
 		}
 	}
+	for _, spec := range specs {
```
john-floren-gravwell, line 118: this is almost exactly the loop directly above, could we turn it into a function call? maybe tagHeaders should be a []corelightSpec, then we could use the same set of logic

kris-watts-gravwell: udpated.

john-floren-gravwell (approved):

Issue #501, opened 2022-09-26
SimpleRelay:Keep-Priority=false does not work


kris-watts-gravwell:
### What are the steps to reproduce this issue?

Data:
```
<185>date=2022-09-26 time=10:11:54 devname="fortigate" devid="FGT60E4Q16015706" eventtime=1664212314092262061 tz="-0700" logid="0100032002" type="event" subtype="system" level="alert" vd="root" logdesc="Admin login failed" sn="0" user="root" ui="ssh(192.168.1.100)" method="ssh" srcip=192.168.1.100 dstip=192.168.1.99 action="login" status="failed" reason="name_invalid" msg="Administrator root login failed from ssh(192.168.1.100) because of invalid user name"
```

Listener:

```
[Listener "syslogudp"]
        Bind-String="udp://0.0.0.0:9999" #standard UDP based RFC5424 syslog
        Reader-Type=rfc5424
        Keep-Priority=false
        Tag-Name=udpfortigate
```
### What happens?

the `<185>` is not removed in the log

### What were you expecting to happen?

output data should not have `<185>` pre-pended

### Any logs, error output, etc?

this option is dumb, moving to "Drop-Priority" instead.

### What versions of software are you using?

5.1.0-1

PR #502, opened 2022-09-26
Fixing up line splitting on UDP

kris-watts-gravwell:
Adding Drop-Priority option for RFC handlers

Adding some additional tests

Fixing the way tests run so we can do them unpriv

This PR addresses https://github.com/gravwell/gravwell/issues/501


ingesters/SimpleRelay/config.go
```diff
@@ -43,7 +43,7 @@ type listener struct {
 	base
 	Tag_Name      string
 	Reader_Type   string
-	Keep_Priority bool // Leave the <nnn> priority value at the start of the log message
```
john-floren-gravwell, line 46: won't this break existing configs if someone has set Keep-Priority=false?

kris-watts-gravwell: its undocumented and was never actually used.  but yeah, i'll leave it just in case

kris-watts-gravwell: addressed

john-floren-gravwell (approved):

PR #507, opened 2022-09-28
Update the Azure Event Hubs ingester dependency

john-floren-gravwell:
Uses v3 of the Azure libraries.


kris-watts-gravwell (approved):
LGTM

PR #509, opened 2022-09-28
Unify generators into a single program

john-floren-gravwell:
This gathers all the generator types into a single program. Once a lot of the accessory stuff is deduplicated, each generator tends to become a very small function.


kris-watts-gravwell (changes requested):
very nice.  I think we can delete the old ones and just have a "generator", no?

generators/unifiedGenerator/binary.ax
```diff
@@ -0,0 +1,6 @@
+# Extraction defintion for the csv generator with everything
```
david-fritz-gravwell, line 1:
```suggestion
# Extraction defintion for the binary generator with everything
```

david-fritz-gravwell (changes requested):

kris-watts-gravwell (approved):
LGTM

david-fritz-gravwell (approved):
Lgtm

Issue #495, opened 2022-09-21
WinEvent: long filters require structured XML


kris-watts-gravwell:
### What are the steps to reproduce this issue?

1. make a config with about 20 EventID filters
2. start ingester

### What happens?

fails to create the subscription

### What were you expecting to happen?

not that

### Any logs, error output, etc?

`2022-09-21T14:58:32.003235-06:00 user-PC winevent - winevents\processing.go:416 [gw@1 name="security" channel="Security" error="The specified query is invalid."] failed to create new eventStream`

### Any other comments?

[read this](https://learn.microsoft.com/en-us/windows/win32/api/winevt/nf-winevt-evtsubscribe)

this line hits: `A query that specifies the types of events that you want the subscription service to return. You can specify an XPath 1.0 query or structured XML query. If your XPath contains more than 20 expressions, use a structured XML query. To receive all events, set this parameter to NULL or "*".`

No idea what a structured XML query is...

### What versions of software are you using?

**Browser:** ...

**GUI Build Version:** ...

Issue #494, opened 2022-09-21
WinEvent: filters


kris-watts-gravwell:
## Who was the customer who requested this feature?

active customer

## Why should we add this feature? (Business justification? What problem is the feature trying to solve?)

The filters are hard to use, individual `EventID={filter}` should support CSVs, thats what we send anyway.

So `EventID=1,2,3,4,5,5` as well as ranges `EventID=1000-2000`.

## How does this feature address the above problem?

Makes it so you don't have 1000 lines in a config when filtering a bunch of EventIDs

PR #510, opened 2022-09-30
winevent filtering improvements

kris-watts-gravwell:
This PR does a few things:

1. It allows specifying EventID filters as a CSV list as well as a range
2. It intelligently breaks up filters into a more complicated XML filtering system so that we can have more than 20 terms
3. it does away with the dog shit template based XML generation and uses a real XML generator

This PR addresses https://github.com/gravwell/gravwell/issues/494
This PR addresses https://github.com/gravwell/gravwell/issues/495


david-fritz-gravwell (approved):
LGTM

PR #513, opened 2022-10-04
client: Do a sync in the Login function.

### LINKED ISSUE #474: CLIENT: client reports incorrect UID for logged in user
### What are the steps to reproduce this issue?

Run this code:
```
nl, _ := objlog.NewNilLogger()
	cli, err := client.NewClient(server, false, true, nl)
	if err != nil {
		log.Fatal("failed to get a client", err)
	} else if err = cli.Login(user, password); err != nil {
		log.Fatal("login failed", err)
	}
        uid := cli.MyUID()
	fmt.Println("your uid", uid)

```

### What happens?

it says my UID is zero

### What were you expecting to happen?

show my correct UID

### What versions of software are you using?

5.0.5

john-floren-gravwell:
Previously, we told users to call Login, then call Sync, which fetches information about the user. This tends to be a bit fraught in practice, because users reasonably assume that after calling NewClient and Login, you should be able to use the client -- but some calls *will* fail, because the client doesn't know the user's UID. This change just calls syncNoLock as the last step of Login.

Closes #474

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -282,7 +282,11 @@ func (c *Client) Login(user, pass string) error {
 	if err := json.NewDecoder(resp.Body).Decode(&loginResp); err != nil {
 		return err
 	}
-	return c.processLoginResponse(loginResp)
+	if err := c.processLoginResponse(loginResp); err != nil {
+		return err
+	}
+
+	return c.syncNoLock()
 }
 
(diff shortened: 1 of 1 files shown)


kris-watts-gravwell (approved):
LGTM

PR #516, opened 2022-10-06
client: don't check if user is admin or not

### LINKED ISSUE #473: Client: client does not detect that you are not logged on on calls
### What are the steps to reproduce this issue?

run this code:

```
nl, _ := objlog.NewNilLogger()
	cli, err := client.NewClient(server, false, true, nl)
	if err != nil {
		log.Fatal("failed to get a client", err)
	}

	defer cli.Close()
	afd, err := cli.UserFiles()
	if err != nil {
		log.Println("Failed to get user files", err)
		return
	}

```

### What happens?

It says that i am not an admin, NOT that I am not logged in.

### What were you expecting to happen?

Complain that i am not logged in

### What versions of software are you using?

5.0.5

john-floren-gravwell:
We decided in #473 that that client shouldn't be attempting to validate whether or not the user is an admin--the webserver will handle it! This strips out checks everywhere I could find them.

Closes #473.

### DIFF
--- a/client/admin.go
+++ b/client/admin.go
@@ -667,22 +667,16 @@ func (c *Client) Restore(rdr io.Reader) (err error) {
 
 // DistributedWebservers queries to determine if the webserver is in distributed mode
 // and therefore using the datastore.  This means that certain resource changes may take some
-// time to fully distribute.
+// time to fully distribute. This is an admin-only function.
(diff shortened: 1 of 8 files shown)


kris-watts-gravwell (approved):
LGTM

Issue #512, opened 2022-10-04
winevents installer MSI: "change" button in installer doesn't do anything


kris-watts-gravwell:
### What are the steps to reproduce this issue?

1. Install the win events ingester on windows using the MSI
2. Fire up the MSI again and get to the "change", "replair", "remove" dialog
3. Click "change"

### What happens?

Nothing.  The button appears to do nothing at all

### What were you expecting to happen?

Either not have that button at all or have it fire up the config dialog that is popped during initial uninstall.  Probably the button should not be there at all

### Any logs, error output, etc?

See screenshot
![Screenshot from 2022-10-04 10-04-50](https://user-images.githubusercontent.com/108431167/193869296-66501911-3675-4ac0-9c8a-d0f8d74164d5.png)

### Any other comments?

We probably need to remove that change button and have a GUI configuration system, Windows people struggle with config files.

### What versions of software are you using?

5.1.0

kris-watts-gravwell:
validated.  buttons are disabled in win10 and the display isn't even shown in win 11

PR #525, opened 2022-10-25
disabling the change and repair buttons in MSI installers

kris-watts-gravwell:
addresses issue https://github.com/gravwell/gravwell/issues/512

we disable the change and repair buttons now.


david-fritz-gravwell (approved):
That's a lot of XML!

PR #530, opened 2022-10-25
Change default ipfix listener port.

john-floren-gravwell:
Port 6343 is more commonly used for s-flow, and it confuses some appliances to try and send ipfix there.


kris-watts-gravwell (approved):

david-fritz-gravwell (approved):
LGTM

PR #536, opened 2022-10-28
Make sure we use concurrency-safe random sources for auth challenges.

john-floren-gravwell:


ingest/utils.go
```diff
@@ -0,0 +1,55 @@
...
+	return rand.New(NewLockedSource(rand.NewSource(SecureSeed())))
+}
+
+func NewInsecureRNG() *rand.Rand {
+	return rand.New(NewLockedSource(rand.NewSource(rand.Int63())))
+}
+
+func SecureSeed() int64 {
+	bts := make([]byte, 8)
+	crand.Read(bts)
```
kris-watts-gravwell, line 53: we should check this error and do something about it.  Failing a cryptographic reseed means a seed of zero which means a predictable PRNG.  Basically on a secure seed if we get an error keep hitting the secure random source until we get something back?  A bad cryptographic source is so bad i honestly think we panic here....

Lets do this:
1. Try to get a secure seed up to 3 times, if it fails all 3 times, go ahead and panic

kris-watts-gravwell (changes requested):
i think we need to handle the event that a secure seed can't be established and panic, OR continue to use the existing securely seeded PRNG.

kris-watts-gravwell (approved):
LGTM

PR #540, opened 2022-11-08
removing the cpu and memory profiling of ingesters

kris-watts-gravwell:
removing cpu and memory profile flags for production ingesters


david-fritz-gravwell (approved):
LGTM

PR #541, opened 2022-11-09
Reactive work

kris-watts-gravwell:
Laying out data structures for reactive flows

The goal here is to have all the stuff needed in the data structures to target a flow with a scheduled search and validate that the we CAN wire it to a floor

### DIFF
--- a/client/types/scheduledsearch.go
+++ b/client/types/scheduledsearch.go
@@ -54,10 +54,10 @@ type ScheduledSearch struct {
 	Schedule    string // when to run: a cron spec
 	Timezone    string // a location to use for the timezone, e.g. "America/New_York"
 	Updated     time.Time
-	Disabled    bool
+	Disabled    bool // scheduling is disabled, do not run this search/script/flow
 	OneShot     bool // Set this flag to 'true' to make the search fire ONCE
 	DebugMode   bool // set this to true to enable debug mode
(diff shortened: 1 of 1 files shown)


client/types/scheduledsearch.go
```diff
@@ -88,6 +94,22 @@ type ScheduledSearch struct {
...
+	Target   uuid.UUID         //The flow or script that will be fired if the query has results (may be blank)
+	Metadata map[string]string `json:",omitempty"` //set of additional free form keys that can be attached to an alert
+}
+
+// ReactiveInput represents a configured key that must be present in an reactive flow
+// The name indicates the name to be injected into the payload of a flow or the global environment for a script
+type ReactiveInput struct {
+	Required   bool   // if set to true then the reactive flow will not start if the value is missing
+	Name       string // name of the value to be injected
+	DebugValue string // temporary debug value to be injected into the payload when doing debug/testing runs
```
john-floren-gravwell, line 110: By doing this, are we essentially saying that every ReactiveInput will always be a string?

kris-watts-gravwell: yes, i want to keep this simple.  The goal is to simplify the process of turning query into an alert

kris-watts-gravwell:
closing this PR, we are going a different route

Issue #551, opened 2022-11-16
S3 ingester bucket url intelligence


kris-watts-gravwell:
## Why should we add this feature? (Business justification? What problem is the feature trying to solve?)

There are a few ways to specify the Bucket URL for an S3 bucket

We should detect and handle the `https://... form` as well as the `arn:aws:s3:::aws-cloudtrail-logs-.....` form as well as `aws-cloudtrail-....` form

## How does this feature address the above problem?

people can stuff whatever in.

## Any other comments?

make this happen soon

PR #552, opened 2022-11-16
fixing up bucket ARN/URL/Name logic 

kris-watts-gravwell:
so it handles things better and stores the state more intelligently

This PR addresses https://github.com/gravwell/gravwell/issues/551


david-fritz-gravwell (approved):
LGTM

Issue #555, opened 2022-11-17
Ingesters: report hostname


kris-watts-gravwell:
## Who was the customer who requested this feature?

Customers

## Why should we add this feature? (Business justification? What problem is the feature trying to solve?)

When there are a few ingesters of the same type using just the IP can be tough, especially if the ingester is behind a NAT.

Resolve the hostname and include it in the report.

## How does this feature address the above problem?

Makes it easier for humans to identify where an ingester is.

## Any other comments?

Coordinate with GUI folks to get it incorporated.

john-floren-gravwell:
Should we send:

* the contents of /etc/hostname?
* whatever the ingester machine gets for a reverse DNS lookup on itself?
* whatever the indexer gets for a reverse DNS lookup on the ingester?

The last two could differ, of course. I lean toward /etc/hostname, personally

PR #569, opened 2022-12-05
ingesters: add hostname to the ingest state object

john-floren-gravwell:
Per #555, we now ship the hostname when we report ingest state.


david-fritz-gravwell (approved):
LGTM

PR #578, opened 2022-12-16
adding some helpers on chartables

kris-watts-gravwell:
adding a new sorting method in a chartable so that we can control series order a little more intelligently on some cases


david-fritz-gravwell (approved):
sure

PR #603, opened 2023-01-13
addressing path clean concerns from security review

kris-watts-gravwell:
Brought up in security review, not exploitable but it flags some automated tools so might as well clean it up and fix.


david-fritz-gravwell (approved):
LGTM

PR #604, opened 2023-01-13
Cleaning up base ingester config to reduce more boilerplate code.

kris-watts-gravwell:
S3 ingester now uses the new base system entirely and the result is dramatically cleaner.

I want to get this methodology vetted before I convert the other ingesters.


david-fritz-gravwell (approved):
LGTM

Issue #602, opened 2023-01-13
SimpleRelay: JSONListener expects line delimited json


kris-watts-gravwell:
## Who was the customer who requested this feature?

Discord user

## What is the feature to be added?

the JSON listener expects to receive compact line delimited json, if you feed it prettyfied JSON or more than one JSON per line it can't handle it.

This listener needs a proper byte.Reader that knows how to consume json.RawMessage and recover from errors.

This could be nasty...

## Why should we add this feature? (Business justification? What problem is the feature trying to solve?)

If anyone expects to send pretty JSON or if a JSON object is broken everything breaks.

## How does this feature address the above problem?

would let us consume serial json objects no matter the source and recover from errors.

## Any other comments?

this might be nasty.

PR #605, opened 2023-01-13
JSON Listeners can eat real JSON objects

kris-watts-gravwell:
JSONListener no longer HAS to have extractors
we now read json objects proper, this means pretty printed as well

This PR addresses https://github.com/gravwell/gravwell/issues/602

### DIFF
--- a/ingesters/SimpleRelay/json.go
+++ b/ingesters/SimpleRelay/json.go
@@ -203,8 +203,10 @@ func jsonConnHandler(c net.Conn, cfg jsonHandlerConfig, igst *ingest.IngestMuxer
 	var lip net.IP // just used for loggin
 	var ts entry.Timestamp
 	var tg *timegrinder.TimeGrinder
-	var tag entry.EntryTag
 	var ok bool
+
+	//initialize our tag to the default tag
+	tag := cfg.defTag
 
 	if ipstr, _, err := net.SplitHostPort(c.RemoteAddr().String()); err != nil {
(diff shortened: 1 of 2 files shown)


ingesters/SimpleRelay/json.go
```diff
@@ -217,65 +231,106 @@ func jsonConnHandler(c net.Conn, cfg jsonHandlerConfig, igst *ingest.IngestMuxer
...
+				if err == io.EOF || errors.Is(err, io.EOF) {
+					ll.Error("oversized json object", log.KV("max-size", cfg.maxObjectSize))
+				} else {
+					ll.Error("oversized json object", log.KV("max-size", cfg.maxObjectSize), log.KVErr(err))
+				}
+			} else if err != io.EOF {
+				//just a plain old error
+				ll.Error("invalid json object", log.KV("max-size", cfg.maxObjectSize), log.KVErr(err))
+			}
+			break // we pretty much have to just hang up
```
john-floren-gravwell, line 280: this has the same effect as calling return, right? I'm leery of break, because for loops and other logic have a tendency to multiply, and before long break is *not* the same as return.

john-floren-gravwell (approved):

ingesters/SimpleRelay/json.go
kris-watts-gravwell: it does, but good point, will address.

kris-watts-gravwell:
PTAL

john-floren-gravwell (approved):

PR #607, opened 2023-01-16
adding new member to ingester stats

kris-watts-gravwell:
to also throw up the list if ingesters that an indexer has seen but hasn't seen lately

This is so we can inform the API about ingesters we saw before but aren't currently connected


david-fritz-gravwell (approved):
LGTM

Issue #564, opened 2022-11-28
SNMP trap ingester: people want it


kris-watts-gravwell:
## What is the feature to be added?

An ingester that can receive SNMP traps and ingest them

Some qualifications on this ingester to keep it simple:
1. Listen only, no polling
2. Ingest traps as they are, no polling to resolve MIBs

## Why should we add this feature? (Business justification? What problem is the feature trying to solve?)

Many switches and routers can't log critical events via standard logging channels, SNMP Traps allow for at least capturing those events.

## How does this feature address the above problem?

Gets logs where there may not be any

## Any other comments?

Cisco is going to do dumb stuff, so make sure to test on them.

john-floren-gravwell:
@kris-watts-gravwell do we know either of the following:

1. What version of SNMP they want to use?
2. What devices they intend to use?

kris-watts-gravwell:
1. no
2. no

PR #615, opened 2023-01-23
Add basic SNMP ingester

john-floren-gravwell:
Per #564, implements an ingester which can accept SNMP Trap messages and ingests them as JSON. Supports SNMP versions 2c and 3.

For SNMPv3, supports MD5 and SHA auth protocols and DES privacy protocol. Messages with mis-matching auth/privacy settings will be dropped.


kris-watts-gravwell (approved):
LGTM

PR #617, opened 2023-01-24
fix pointer silliness

john-floren-gravwell:
I was overwriting changes to gosnmp.Default, this fixes that. Also makes sure we don't drop on MsgFlags unless it's v3.


kris-watts-gravwell (approved):
LGTM

Issue #623, opened 2023-01-31
SimpleRelay: RegexExtractor just does nothing if set to use UDP as the  bind-string


kris-watts-gravwell:
## What are the steps to reproduce this issue?

1. Setup a RegexExtractor that uses UDP as the bind-string protocol (udp://0.0.0.0:6666)
2. Start the ingester

## What happens?

Ingester fires up and announces its intention to bind to and handle regex on that UDP port, then... just doesn't.

## What were you expecting to happen?

Support UDP

## Any other comments?

looks like we have an if statement that just straight doesn't handle UDP.

## What versions of software are you using?

5.1.4-1

kris-watts-gravwell:
JSON extractor does the exact same thing.   Need to fix that too.

PR #624, opened 2023-01-31
fixing regex handler so that it can do TCP, UDP, or TLS

kris-watts-gravwell:
Updates the regexListener in simple relay so that we can do UDP, TCP, TLS.  UDP was missing.

I had to change some form and function a bit for regex loop so that i didn't have to duplicate a bunch of code for UDP.

This should remove some needless async too though, so maybe resource reduction.

This PR addresses https://github.com/gravwell/gravwell/issues/623

### DIFF
--- a/ingesters/SimpleRelay/json.go
+++ b/ingesters/SimpleRelay/json.go
@@ -239,6 +239,7 @@ func jsonAcceptorUDP(conn *net.UDPConn, id int, igst *ingest.IngestMuxer, cfg js
 			return
 		}
 	}
+	ll := log.NewLoggerWithKV(lg, log.KV("json-listener", cfg.name))
 	for {
 		n, raddr, err := conn.ReadFromUDP(buff)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


john-floren-gravwell (approved):

kris-watts-gravwell:
had to fix the JSON side too, it was doing the same thing.

ingesters/SimpleRelay/json.go
```diff
@@ -193,20 +206,75 @@ func jsonAcceptor(lst net.Listener, id int, igst *ingest.IngestMuxer, cfg jsonHa
...
+			continue
+		}
+		var rip net.IP
+		if cfg.src == nil {
+			rip = raddr.IP
+		} else {
+			rip = cfg.src
+		}
+		// get a local logger up that will always add some more info
+		ll := log.NewLoggerWithKV(lg, log.KV("json-listener", cfg.name), log.KV("remoteaddress", rip.String()))
```
john-floren-gravwell, line 263: a new allocation per packet seems excessive?

john-floren-gravwell (commented):
I also feel like most of jsonAcceptorUDP and regexAcceptorUDP could be factored into another function which takes a timegrinder config and a callback function, then we just hand it an appropriate lambda which does either the regex or json processing code... but expediency may take precedence over neatness of implementation

john-floren-gravwell (approved):

PR #631, opened 2023-02-16
if Community is unset, accept traps from any community

john-floren-gravwell:
This gives people a little more flexibility. You can set the  `Community` to require a community string on traps, or leave it blank to accept everything.


kris-watts-gravwell (approved):
LGTM

PR #633, opened 2023-02-21
Improve the HEC listener.

john-floren-gravwell:
Change the URL parameter for HEC-Compatible-Listener to be the base (`/services/collector`) instead of the "event" path (`/services/collector/event`). This may be a breaking change for some setups, but I think we can make the following assumptions:

1. Most deployments with the HEC listener are using the default
2. Even if someone specified `URL=/services/collector/event`, it will still work ok because we'll listen on that endpoint with the same handler as on `/services/collector/event/event`. ACKs won't be quite right, but ACKs weren't quite right anyway.

The raw endpoint now splits on lines. You can specify a different line break string via e.g. `Raw-Line-Breaker=";"`

If an ACK is requested (by specifying a channel in the request) we will always send an ACK ID.

Also improves logging to the gravwell tag.


kris-watts-gravwell (approved):
LGTM

PR #636, opened 2023-02-22
Detect the old URL specification style for HEC so as to not break old…

john-floren-gravwell:
… configs


kris-watts-gravwell (approved):
Lgtm

PR #640, opened 2023-02-24
Add LastSeen to missing ingesters

kris-watts-gravwell:


client/types/render.go
```diff
@@ -447,21 +452,30 @@ func (eis emptyIngesterStates) MarshalJSON() ([]byte, error) {
 	return json.Marshal([]ingest.IngesterState(eis))
 }
 
+type emptyMissingIngester []MissingIngester
```
john-floren-gravwell, line 455: not wild about this naming although i acknowledge it's internal

kris-watts-gravwell: I agree, its just for the marshalling garbage

I wish so much i could just annotate the field and say "not null"

john-floren-gravwell (approved):

PR #644, opened 2023-02-28
adding code to update the LastSeen when adding children

kris-watts-gravwell:
just adding some more scaffolding so we have more up-to-date status of children all the way through the ingest mesh.


david-fritz-gravwell (approved):
LGTM

PR #654, opened 2023-03-10
fix the debug statements

kris-watts-gravwell:
and move to a non-regex based barrier finder.  Got a clean 19.5X speed up.

### DIFF
--- a/ingesters/SimpleRelay/main.go
+++ b/ingesters/SimpleRelay/main.go
@@ -17,6 +17,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime/debug"
+	"runtime/pprof"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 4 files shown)


ingesters/SimpleRelay/main.go
```diff
@@ -88,6 +89,14 @@ func mainInit() {
 }
 
 func main() {
+	f, err := os.Create("/tmp/cpu.prof")
```
david-fritz-gravwell, line 92: We probably don't want profiling enabled on every run

david-fritz-gravwell (changes requested):

ingesters/SimpleRelay/main.go
kris-watts-gravwell: great catch.

PR #669, opened 2023-03-23
add a new method to assign directly into an enumerated data item

kris-watts-gravwell:
this is  for efficiently converting from types.StringTagEntry


david-fritz-gravwell (approved):
LGTM

Issue #316, opened 2021-11-19
new plugin: syslog router


traetox:
everyone sucks at syslog, add a new plugin that can do routing based on hostname and appname/tag.

use this: https://github.com/jeromer/syslogparser

floren:
Things we might route on:

* application name
* hostname
* msgid
* priority/facility/severity
* message (maaaaaybe)
* structured data ID
* contents of a structured data item

the last two aren't really possible with that library

david-fritz-gravwell:
It would be pretty trivial and quick to write a flexible syslog parser in pbpg if there isn't a library that can do the structured extraction, etc., for us.

PR #672, opened 2023-04-03
Adding syslog router preprocessor

kris-watts-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/316

basically you can define a template that extracts bits from the header of a syslog message and then create a new tag based on that syslog message.  CAN BE DANGEROUS.  But we have never refused to make dangerous tools

Check tests for examples.

### DIFF
--- a/ingest/processors/syslogrouter.go
+++ b/ingest/processors/syslogrouter.go
@@ -1,0 +1,207 @@
+/*************************************************************************
+ * Copyright 2018 Gravwell, Inc. All rights reserved.
+ * Contact: <legal@gravwell.io>
+ *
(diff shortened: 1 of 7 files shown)


ingest/processors/syslogrouter.go
```diff
@@ -0,0 +1,207 @@
+/*************************************************************************
+ * Copyright 2018 Gravwell, Inc. All rights reserved.
```
john-floren-gravwell, line 2: year

john-floren-gravwell (changes requested):

kris-watts-gravwell:
fixed, also fixed go.mod tidy.  @john-floren-gravwell PTAL

PR #673, opened 2023-04-04
Fix the function name stuttering

kris-watts-gravwell:
and add comment about stuttered name as deprecated

leaving old function, but marking it deprecated

The original function names stutter

Also fixing link to docs page in file follower error.


david-fritz-gravwell (approved):
LGTM

PR #676, opened 2023-04-11
fix the kinesis ingester config example

kris-watts-gravwell:
example in config was wrong.

Kinesis delivery stream is unsupported, but the config snippets should still be right.


david-fritz-gravwell (approved):
LGTM

PR #689, opened 2023-04-19
updating versions and fixing existing install detection on file follower

kris-watts-gravwell:
we weren't drilling deep enough into the config directory to detect the existing config.


ashnwade (approved):
LGTM

PR #690, opened 2023-04-19
updating verisons and fixing existing install detection on file follower

kris-watts-gravwell:
cherry-picking hotfix out of dev


ashnwade (approved):
LGTM

PR #692, opened 2023-04-21
Add custom marshallers for some ABAC types.

john-floren-gravwell:
Javascript really does not like to get a null when it expects an empty array, but that's what Go will send if your struct contains e.g. a []string and you don't explicitly initialize the slice. These custom marshallers make sure we send an empty array, making the frontend's life easier.

Go desperately needs some way to tell the JSON encoder that we want this behavior, see:

https://github.com/golang/go/issues/37711
https://github.com/golang/go/issues/27589
https://github.com/golang/go/pull/27813

### DIFF
--- a/client/types/abac.go
+++ b/client/types/abac.go
@@ -9,6 +9,7 @@
 package types
 
 import (
+	"encoding/json"
 	"errors"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


kris-watts-gravwell (approved):
LGTM

Issue #668, opened 2023-03-23
File Follow does not ingest content if the final (or only) record does not have a newline.


kris-watts-gravwell:
## What are the steps to reproduce this issue?

1. Stand up file follow watching some set of files
2. Output data into one of the watch files **WITHOUT** a training newline

## What happens?

File follow catches that the file has data but does not consume the entry because it is waiting for a newline.  We have to do this to support dumb logging systems that write out lines in bits and pieces where you might get multiple inotify events over a single line.

## What were you expecting to happen?

Eventually "timeout" and consume the entry

## Any other comments?

We may need to look into something that will "tick" on those file handling routines and if there is data to be had but no events have occurred in XYZ time we go ahead and consume it.

The risk is if there is something INCREDIBLY stupid where half a line gets written, the writer blocks for an extended period of time, then writes the other half.  We would get two entries.

## What versions of software are you using?

5.2.1

kris-watts-gravwell:
this is fixed now

PR #699, opened 2023-04-28
fixing up the file follower 

kris-watts-gravwell:
so that if its been long enough and we don't have a trailing delimiter, we just consume the last entry.  This to deal with files that don't write a trailing newline

This PR addresses https://github.com/gravwell/gravwell/issues/668


david-fritz-gravwell (approved):
LGTM

PR #701, opened 2023-05-01
adding OSV TOML to announce invalid vuln reports

kris-watts-gravwell:
We don't use the S3 library in the way the vuln describes and don't want/need write permissions for S3 buckets.   We also don't support KMS in the ingester, so basically its not possible to trigger the Amazon library vulnerability using our ingester.


david-fritz-gravwell (approved):
LGTM

PR #702, opened 2023-05-01
make the catchup functions also check for and deal with data missing a termination value

kris-watts-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/668


david-fritz-gravwell (approved):
LGTM

PR #706, opened 2023-05-08
change igester forget/tracking path route

kris-watts-gravwell:
as per requested comments on API


michael-wisely-gravwell (approved):
LGTM :+1:

PR #710, opened 2023-05-08
fixing json handler and adding new optimization flag

kris-watts-gravwell:
the JSON and Regex listener on SimpleRelay could not actually handle custom timeformats due to some overly aggressive validation at startup.

We also added a `Disable-Multithreading` boolean that can be stuffed into ingesters that are typically very IO bound and need to set GOMAXPROCS(1) to eek out the most possible performance on NUMA machines.


david-fritz-gravwell (approved):
LGTM

PR #718, opened 2023-05-17
Add JSON responses to HEC listeners

john-floren-gravwell:
The docs indicate that you should send back a JSON response message with status codes etc. We weren't doing that because everything we'd tested with thus far was happy with just HTTP status codes, but some things are looking for a proper response.

@kris-watts-gravwell If you can build this and try it out yourself before we merge, I'll feel a lot better about the change. It shouldn't affect anything, since it's just adding a response body where previously there was none, but who knows.


kris-watts-gravwell:
built and pushed up to an HTTP ingester receiving corelight.  It is happy.

kris-watts-gravwell (approved):
LGTM

PR #725, opened 2023-05-22
Remove GetTags capability

john-floren-gravwell:
We control access to tags separately, and it doesn't make sense to hide the *list* of tags the user has access to -- they've been explicitly granted permission to use those tags!


kris-watts-gravwell (approved):
LGTM

PR #727, opened 2023-05-24
make the ingest library build on macos, with or without m1

david-fritz-gravwell:
We don't actually use close on exec in Dup3, so Dup2 is fine, which macOS requires. The only other fix is a build constraint.


kris-watts-gravwell (approved):
LGTM

PR #738, opened 2023-06-01
Add SearchURl and SearchStatus functions to the client library

john-floren-gravwell:
SearchURL returns a URL that should be valid to access the search results.

SearchStatus fetches the SearchCtrlStatus for a single query.


kris-watts-gravwell (approved):
LGTM

PR #741, opened 2023-06-06
Add ExternalAddr to GUISettings object

john-floren-gravwell:
This allows the webserver to give its "canonical" address, the one it thinks users should be coming in through.

Also updates the SearchURL function to use that address if available.


kris-watts-gravwell (approved):
LGTM

Issue #707, opened 2023-05-08
Add Timestamp and regex delimeter handling to import ingester file handler


kris-watts-gravwell:
OEM customer asked that the file stuff for the import ingester support all the same flags as the file follower.  It is currently missing `Ignore-Glob`.

Here are the pertinent configs for file follower:

```
type follower struct {
        Base_Directory            string // the base directory we will be watching
        File_Filter               string // the glob for pattern matching
        Tag_Name                  string
        Ignore_Timestamps         bool //Just apply the current timestamp to lines as we get them
        Assume_Local_Timezone     bool
        Recursive                 bool // Should we descend into child directories?
        Attach_Filename           bool // attach the full path of the file to each entry
        Ignore_Line_Prefix        []string
        Ignore_Glob               []string
        Timestamp_Format_Override string //override the timestamp format
        Timestamp_Delimited       bool
        Timezone_Override         string
        Regex_Delimiter           string
        // these two must be used together
        Timestamp_Regex         string
        Timestamp_Format_String string
        // so that we can initialize the timegrinder
        timeFormats config.CustomTimeFormat
        Preprocessor              []string
}
```

Here are the configs for Migrate:
```
type files struct {                                      
        Base_Directory            string // the base directory we will be watching
        File_Filter               string // the glob for pattern matching
        Tag_Name                  string
        Ignore_Timestamps         bool //Just apply the current timestamp to lines as we get them
        Assume_Local_Timezone     bool
        Timestamp_Format_Override string //override the timestamp format
        Timezone_Override         string                                      
        Recursive                 bool // Should we descend into child directories?
        Ignore_Line_Prefix        []string
        Preprocessor              []string
}
``` 

Things that are missing:

- [x] Attach-Filename
- [x] Ignore-Glob
- [ ] Timestamp-Delimted
- [ ] Regex-Delimiter

PR #752, opened 2023-06-16
Add Attach-Filename and Ignore-Glob to the migrate ingester

john-floren-gravwell:
Addresses #707.

Moved the LineIgnorer definition from filewatch to utils so we could apply it here.

### DIFF
--- a/ingesters/utils/lineignorer.go
+++ b/ingesters/utils/lineignorer.go
@@ -1,0 +1,54 @@
+/*************************************************************************
+ * Copyright 2023 Gravwell, Inc. All rights reserved.
+ * Contact: <legal@gravwell.io>
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD 2-clause license. See the LICENSE file for details.
+ **************************************************************************/
+
+package utils
+
+import (
+	"bytes"
+
+	"github.com/gobwas/glob"
+)
+
+type LineIgnorer struct {
+	prefixes [][]byte
+	globs    []glob.Glob
+}
+
+func NewIgnorer(prefixes, globs []string) (*LineIgnorer, error) {
+	li := &LineIgnorer{}
+	for _, v := range prefixes {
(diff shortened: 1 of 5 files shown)


ingesters/utils/lineignorer.go
```diff
@@ -0,0 +1,54 @@
...
+
+type LineIgnorer struct {
+	prefixes [][]byte
+	globs    []glob.Glob
+}
+
+func NewIgnorer(prefixes, globs []string) (*LineIgnorer, error) {
+	li := &LineIgnorer{}
+	for _, v := range prefixes {
+		li.prefixes = append(li.prefixes, []byte(v))
```
kris-watts-gravwell, line 25: check if its empty, otherwise an empty prefix drops everything

ingesters/utils/lineignorer.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+		li.globs = append(li.globs, c)
+	}
+	return li, nil
+}
+
+// Ignore returns true if the given byte slice matches any of the prefixes or
+// globs in the ignorer.
+func (l *LineIgnorer) Ignore(b []byte) bool {
+	for _, prefix := range l.prefixes {
```
kris-watts-gravwell, line 40: if len(b) == 0 {return false}

ingesters/utils/lineignorer.go
```diff
@@ -0,0 +1,54 @@
...
+		if bytes.HasPrefix(b, prefix) {
+			return true
+		}
+	}
+
+	bString := string(b)
```
kris-watts-gravwell, line 46: the cast to a string creates a copy, because we are checking and never touching again...

```
bString := unsafe.String(&b[0], len(b))
```

kris-watts-gravwell (changes requested):
minor changes, then we are good

kris-watts-gravwell (approved):
LGTM, thank you

PR #755, opened 2023-06-22
adding a proper HTTP response draining function

kris-watts-gravwell:
and checking that it is used

### DIFF
--- a/client/utils.go
+++ b/client/utils.go
@@ -1,0 +1,49 @@
+/*************************************************************************
+ * Copyright 2023 Gravwell, Inc. All rights reserved.
+ * Contact: <legal@gravwell.io>
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD 2-clause license. See the LICENSE file for details.
+ **************************************************************************/
+
+package client
+
+import (
+	"errors"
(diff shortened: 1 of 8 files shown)


kris-watts-gravwell:
@david-fritz-gravwell PTAL

client/utils.go
```diff
@@ -0,0 +1,49 @@
...
+	ErrMaxBodyDrained = errors.New("too many response bytes in body, closing")
+)
+
+// drainResponse will drain up to 4MB of data then close the response Body.
+// We do this so that http requests can re-use connects as per docs.
+func drainResponse(resp *http.Response) {
+	if resp.Body == nil {
+		return
+	}
+	var nw nilWriter
```
david-fritz-gravwell, line 31: why not use ioutil.Discard as the writer here?

david-fritz-gravwell (changes requested):

david-fritz-gravwell (approved):
LGTM

client/utils.go
kris-watts-gravwell: a limitedWriter + io.Discard would be equivalent.  This felt just as simple.

PR #761, opened 2023-07-03
fixing windows implementation of the setIngesterUUID function

kris-watts-gravwell:
the API on unix didn't translate exactly to windows, so we needed to rework the paths into their parts before calling os.Mktemp.


ashnwade (approved):
LGTM

PR #762, opened 2023-07-05
get rid of the GID field for kits, use GIDs instead

john-floren-gravwell:
Also adds InstallationGroups field to the KitConfig struct


kris-watts-gravwell (approved):
LGTM

PR #765, opened 2023-07-06
Improve errors returned by the client

john-floren-gravwell:
If there was a non-200 response, the Client functions will return a ClientError, which contains the status code and body of the error message returned. This makes it easier to test what failed, rather than doing string matching.


kris-watts-gravwell (approved):
Lgtm

PR #767, opened 2023-07-11
dev -> preview/search-store

kris-watts-gravwell:
add URL endpoint for getting entries using the new REST API and bring in dev.


david-fritz-gravwell (approved):
LGTM

PR #772, opened 2023-07-13
add support for quoted json fields when wrapped in raw quotes.

### LINKED ISSUE #558: Support more complicated json keys in preprocessors

If a json key has dots, quotes, or other monkey business our current key specifications for jsonextract, jsontimeextract, etc.. won't support it.

Consider this blob:
```
{
  "foo": {
    "bar.baz": "stuff"
  }
}
```

We will struggle to get at the key `foo."bar.baz"`

## Any other comments?

this implementation needs to match the internal one for the json search module.

david-fritz-gravwell:
Fixes #558

### DIFF
--- a/.github/workflows/golang-testing.yml
+++ b/.github/workflows/golang-testing.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v3
       with:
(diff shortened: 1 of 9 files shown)


kris-watts-gravwell (approved):
LGTM

Issue #770, opened 2023-07-13
Ingester logs: logs don't rotate or trim, so they just grow and grow


kris-watts-gravwell:
the file logging system used for ingesters does not rotate logs, they just grow and grow.

It should do the following:

1. enable the ability to truncate and rotate logs
2. keep a tail
3. compress the old ones

PR #775, opened 2023-07-17
Add log/rotate package to do file rotation on ingesters

kris-watts-gravwell:
This PR works on https://github.com/gravwell/gravwell/issues/770

Only SimpleRelay is wired now, if this looks good i will wire the others too.


david-fritz-gravwell (approved):
LGTM

PR #782, opened 2023-07-31
EV attach in ingesters

kris-watts-gravwell:
This implements the "attach" thing so that any ingester can slap a global config and the ingester will attach stuff.

It allows for attaching arbitrary strings as well as some keywords:

Here is an example config for Simple Relay:

```
Log-Level=INFO
Log-File=/opt/gravwell/log/simple_relay.log

[Attach]
        host="$HOSTNAME"
        foobar="foo to the bar"
        ingested="$NOW"
        ingester="$UUID"


#basic default logger, all entries will go to the default tag
# this is useful for sending generic line-delimited
# data to Gravwell, for example if you have an old log file sitting around:
#       cat logfile | nc gravwell-host 7777
#no Tag-Name means use the default tag
[Listener "default"]
        Bind-String="0.0.0.0:7777" #we are binding to all interfaces, with TCP implied
        Tag-Name=default4
```

This will attach the ingester UUID, timestamp at point of ingestion, and the derived hostname.  It also attaches arbitrary string `foo to the bar`

### DIFF
--- a/ingest/attach/attach.go
+++ b/ingest/attach/attach.go
@@ -1,0 +1,144 @@
+/*************************************************************************
+ * Copyright 2023 Gravwell, Inc. All rights reserved.
+ * Contact: <legal@gravwell.io>
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD 2-clause license. See the LICENSE file for details.
+ **************************************************************************/
+
+package attach
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/google/uuid"
+	"github.com/gravwell/gcfg"
+	"github.com/gravwell/gravwell/v3/ingest/entry"
+)
+
+const (
+	nowId  = `$NOW`
+	uuidId = `$UUID`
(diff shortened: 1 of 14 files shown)


david-fritz-gravwell:
Does this support the multiple EVs of the same name thing?

kris-watts-gravwell:
> Does this support the multiple EVs of the same name thing?

no, not on the same ingester.  But if you start stacking them through things like the federator then yes.

kris-watts-gravwell:
This needs some additional logic so that you can't push in the same EV multiple times

kris-watts-gravwell:
ok, this logic now updates not append, this means that if an ingester and/or preprocessor collides on a name we update.

This logic is much easier to grok rather than trying to have duplicates or allowing something like a vector of values with disparate types.

@john-floren-gravwell @david-fritz-gravwell  PTAL

john-floren-gravwell (approved):

PR #787, opened 2023-08-01
More boilerplate rollover

kris-watts-gravwell:
Convert the following ingesters:

- kinesis
- O365
- MSGraph
- GooglePubSub

This also standardizes the debugout code that is used all over hell so it looks the same everywhere

### DIFF
--- a/ingesters/GooglePubSubIngester/main.go
+++ b/ingesters/GooglePubSubIngester/main.go
@@ -10,21 +10,15 @@ package main
 
 import (
 	"context"
-	"flag"
 	"fmt"
 	"net"
 	"os"
-	"path"
-	"runtime/debug"
-	"syscall"
 	"time"
 
-	"github.com/gravwell/gravwell/v3/ingest"
-	"github.com/gravwell/gravwell/v3/ingest/config/validate"
 	"github.com/gravwell/gravwell/v3/ingest/entry"
 	"github.com/gravwell/gravwell/v3/ingest/log"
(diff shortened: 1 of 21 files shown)


ingesters/GooglePubSubIngester/main.go
```diff
@@ -34,123 +28,42 @@ const (
...
 
 var (
-	configLoc      = flag.String("config-file", defaultConfigLoc, "Location of configuration file")
-	confdLoc       = flag.String("config-overlays", defaultConfigDLoc, "Location for configuration overlay files")
-	verbose        = flag.Bool("v", false, "Display verbose status updates to stdout")
-	ver            = flag.Bool("version", false, "Print the version information and exit")
-	stderrOverride = flag.String("stderr", "", "Redirect stderr to a shared memory file")
-	lg             *log.Logger
+	lg *log.Logger
+	v  bool
```
david-fritz-gravwell, line 36: Having a global named "v" is probably going to bite us, as "v" is also the idiomatic loop variable when using `range`. We'll certainly shadow this variable in places.

I suggest we change all of these to "verbose" (renaming the flag variable), or just use the flag variable that's already there as we were doing before.

david-fritz-gravwell (changes requested):

ingesters/GooglePubSubIngester/main.go
kris-watts-gravwell: agreed, fixed, PTAL

PR #792, opened 2023-08-03
Add DebugEvent fields for flows.

john-floren-gravwell:
FlowParseRequest and ScheduledSearch types now have a DebugEvent type. If set, it will be injected into the incoming payloads of flows in the following cases:

- When parsing a flow via FlowParseRequest
- When running a flow in debug mode (DebugMode = true)

Also introduces the ParseReactiveFlow function on the client, to make it easier to validate flows as they are being built.


kris-watts-gravwell (approved):
LGTM

PR #798, opened 2023-08-14
fixing windows implementation of the setIngesterUUID function

kris-watts-gravwell:
this is already fixed in dev, just pulling it into master too


ashnwade (approved):
LGTM

PR #809, opened 2023-08-28
change GUID to UUID in packedscheduledsearch JSONMetadata

john-floren-gravwell:
Brings it in line with convention elsewhere


kris-watts-gravwell (approved):
LGTM, is this going to screw with kit upgrades after an update?

PR #810, opened 2023-08-29
Make GetFlow and GetScheduledSearch more flexible.

john-floren-gravwell:
You can fetch a flow or scheduled search by GUID or by ID. This change enables that in the client library.


kris-watts-gravwell (approved):
LGTM

PR #814, opened 2023-08-31
update s3/sns/sqs routine to handle multiple message types

david-fritz-gravwell:


ingesters/s3Ingester/manager.go
```diff
@@ -68,36 +68,76 @@ func sqsS3Routine(s *SQSS3Listener, wg *sync.WaitGroup, ctx context.Context, lg
...
+	}
+
+	var buckets []string
+
+	// all the buckets are the same in this message type
+	for range subMessage.S3ObjectKey {
+		buckets = append(buckets, subMessage.S3Bucket)
+	}
+
+	return buckets, subMessage.S3ObjectKey, nil
```
kris-watts-gravwell, line 119: Please do a final validation check that the number of buckets and keys match.  Above you iterate over the keys and just use that index to grab the bucket.

DO NOT trust amazon API to not do some really dumb shit.

ingesters/s3Ingester/manager.go
```diff
@@ -68,36 +68,76 @@ func sqsS3Routine(s *SQSS3Listener, wg *sync.WaitGroup, ctx context.Context, lg
...
+	}
+
+	var buckets []string
+	var keys []string
+	for _, v := range d.Records {
+		buckets = append(buckets, v.S3.Bucket.Name)
+		keys = append(keys, v.S3.Object.Key)
+	}
+
+	return buckets, keys, nil
```
kris-watts-gravwell, line 138: Also check that you have a valid bucket name and/or key.  Probably just check that you can parse the URL/ARN and the key isn't empty?

ingesters/s3Ingester/manager.go
```diff
@@ -68,36 +68,76 @@ func sqsS3Routine(s *SQSS3Listener, wg *sync.WaitGroup, ctx context.Context, lg
...
+
+func s3Decode(input []byte) ([]string, []string, error) {
+	b := bytes.NewBuffer(input)
+	jdec := json.NewDecoder(b)
+	var d *s3Records
+	err := jdec.Decode(&d)
+	if err != nil {
+		return nil, nil, err
+	}
+
```
kris-watts-gravwell, line 130: check the event type here, otherwise you are going to attempt to fire on restorations, deletes, etc...

Look for the `ObjectCreated:*` stuff

ingesters/s3Ingester/manager.go
```diff
@@ -68,36 +68,76 @@ func sqsS3Routine(s *SQSS3Listener, wg *sync.WaitGroup, ctx context.Context, lg
...
+		return nil, nil, err
+	}
+	sb := strings.NewReader(d.Message)
+	jdec = json.NewDecoder(sb)
+	var subMessage s3SubMessage
+	err = jdec.Decode(&subMessage)
+	if err != nil {
+		return nil, nil, err
+	}
+
```
kris-watts-gravwell, line 111: Is there an event type?  We only want to fire on created/new stuff

kris-watts-gravwell (changes requested):
little more defensive programming on API handling.

Just add some additional checks on the objects to make sure you are getting what you expect.

Also we need to filter on event types so we aren't trying to grab buckets on deletions, restorations, etc...

ingesters/s3Ingester/manager.go
david-fritz-gravwell: The range operation above literally does this -- just make a slice, the same size as S3ObjectKey, and replicate the bucket string that many times.

ingesters/s3Ingester/manager.go
david-fritz-gravwell: There is not.

ingesters/s3Ingester/manager.go
david-fritz-gravwell: done

david-fritz-gravwell:
@traetox PTAL

kris-watts-gravwell (approved):
LGTM

PR #815, opened 2023-08-31
Add MaxEvents field to AlertDefinition type.

john-floren-gravwell:
This lets the user specify how many events should be allowed per triggering of the alert. If zero, the search agent will use a default.

Also cleans up a bunch of omitempty tags that we're no longer using.


david-fritz-gravwell (approved):
LGTM

PR #837, opened 2023-09-25
fixup remaining ingesters to use cancel contexts where needed.

### LINKED ISSUE #836: ingest: change all ingesters to use ProcessContext() instead of Process()
## What are the steps to reproduce this issue?

1. Ingest and get into a very bad state -- disk full + cache full
2. Try to exit

## What happens?

If you aren't using ProcessContext(), we can't gracefully shutdown.

david-fritz-gravwell:
Fixes #836


kris-watts-gravwell (approved):
LGTM

PR #845, opened 2023-10-05
change alert validation functions to take UUIDs.

john-floren-gravwell:
they should not be using the int32 IDs

This PR addresses [ISSUE_LINK]


david-fritz-gravwell (approved):
LGTM

PR #851, opened 2023-10-12
dev -> master for 5.4.1

john-floren-gravwell:


david-fritz-gravwell (approved):
LGTM

john-floren-gravwell:
you'd think something as simple as dev->master merges would be handled without github getting all huffy about a branch out of date but you'd be wrong.

PR #861, opened 2023-10-16
remove ms-graph package from plugins

kris-watts-gravwell:
this package is awful anyway, not to mention it should not even be in these plugins.


david-fritz-gravwell (approved):
LGTM

PR #866, opened 2023-10-19
add well-specific UUID and make indexer UUID a required field when encoding with JSON.

david-fritz-gravwell:


evan-gravwell (approved):
LGTM!

PR #869, opened 2023-10-22
remove old drone file, update some go mod libs, embed tz database into binaries

kris-watts-gravwell:


david-fritz-gravwell (approved):
LGTM

PR #873, opened 2023-10-25
merge master and make clients set state on logout

kris-watts-gravwell:
client lib was not setting state correctly on logout, so you could not log back in.


david-fritz-gravwell (dismissed):
LGTM

kris-watts-gravwell:
@david-fritz-gravwell PTAL   i hate that go get does not update go sum... i hate it so so so much.

david-fritz-gravwell (approved):
LGTM

PR #878, opened 2023-10-27
Try to handle deleted BaseDirs in filewatch

john-floren-gravwell:
If one of our BaseDirs gets deleted, add it to a list which periodically gets checked. If a deleted dir gets recreated, start following it again.


kris-watts-gravwell:
i THINK this seems reasonable

kris-watts-gravwell (approved):
LGTM

PR #881, opened 2023-10-31
sigh...

david-fritz-gravwell:


ingest/muxer.go
```diff
@@ -270,6 +270,7 @@ func newIngestMuxer(c MuxerConfig) (*IngestMuxer, error) {
 
 	// connect up the chancacher
 	gob.Register(&entry.Entry{})
```
kris-watts-gravwell, line 272: should these be at the package level in an `init()` so that we don't run into something similar later when we try to use the package in an new and interesting way?

kris-watts-gravwell (changes requested):
i vote that the gob types are registered in an init function at the package level.

kris-watts-gravwell (approved):
LGTM

PR #884, opened 2023-11-01
first crack at a worker pool for s3 and sqs/sns

david-fritz-gravwell:


ingesters/s3Ingester/config.go
```diff
@@ -95,6 +97,7 @@ func GetConfig(path, overlayPath string) (*cfgType, error) {
 		IngestConfig:         cr.Global.IngestConfig,
 		Attach:               cr.Attach,
 		State_Store_Location: cr.Global.State_Store_Location,
+		Worker_Pool_Size:     cr.Global.Worker_Pool_Size,
```
kris-watts-gravwell, line 100: init cr.Global.Worker_Pool_Size to one prior to reading the file into the cr structure.

kris-watts-gravwell (changes requested):
init pool to 1 prior to reading in the config file.

ingesters/s3Ingester/manager.go
```diff
@@ -19,28 +19,37 @@ import (
...
 	return
 }
 
-func sqsS3Routine(s *SQSS3Listener, wg *sync.WaitGroup, ctx context.Context, lg *log.Logger) {
+func sqsS3Routine(s *SQSS3Listener, wg *sync.WaitGroup, ctx context.Context, lg *log.Logger, numWorkers int) {
 	defer wg.Done()
 
+	// create workers
+	var workerWg sync.WaitGroup
+	queue := make(chan *sqs.Message, QUEUE_DEPTH)
```
kris-watts-gravwell, line 47: I don't see that this ever actually gets closed?  how are workers shutdown on this code path?

kris-watts-gravwell (changes requested):
On the normal code path i don't see anything that looks like it closes the input queue

ingesters/s3Ingester/manager.go
david-fritz-gravwell: good catch -- PTAL

kris-watts-gravwell (approved):
Lgtm.  But let's test this on gin before we release to the world

PR #888, opened 2023-11-03
tweaks to make simple relay work on freebsd

john-floren-gravwell:


kris-watts-gravwell (changes requested):
Windows builds broke.  Not sure why

Issue #902, opened 2023-11-16
Stderr should not receive all internal log message, only fatal


kris-watts-gravwell:
## What are the steps to reproduce this issue?

1. Start any gravwell ingester/component
2. look at journal

## What happens?

all of our debug logs and info logs are in there

## What were you expecting to happen?

we should ONLY wire fatal logs out to stderr

## Any other comments?

error is debatable, but definitely not info, and warn

## What versions of software are you using?

5.4.2

david-fritz-gravwell:
LGTM

Issue #901, opened 2023-11-16
the stderr error logger should not contain entire RFC5424 message


kris-watts-gravwell:
## What are the steps to reproduce this issue

1. Install an ingester
2. Route systems syslog out to our ingester
3. observe output

## What happens?

This is what the entry is:
```
<30>1 2023-11-16T23:27:22.725587+00:00 hostname gravwell_simple_relay 35859 - -  <14>1 2023-11-16T23:27:22.721015Z hostname simplerelay - SimpleRelay/main.go:219 [gw@1 ingesteruuid="a9ae46fc-0312-47aa-b5ed-1be8b1330000" active="9"] Closing active connections
```

## What were you expecting to happen?

Just emit the ` [gw@1 ingesteruuid="a9ae46fc-0312-47aa-b5ed-1be8b1330000" active="9"] Closing active connections` portion when sending to stderr

## Any other comments?

the syslog  relay/journal/whatever already puts an appname and header on it, we shouldn't do that too

This also happens in the windows internal logging, we emit a full RFC5424 message into the windows event logger, thats bad.

## What versions of software are you using?

5.4.2

david-fritz-gravwell:
I'm still seeing full RFC5424 messages on fatal logs to stderr:

```
√ valinor:gravwell/ingesters/HttpIngester % ./HttpIngester -config-file test.conf


^C<11>1 2023-11-29T14:41:45.315424-07:00 valinor httpingester - HttpIngester/main.go:177 [gw@1 error="http: Server closed"] failed to serve HTTP server
```

kris-watts-gravwell:
You will because you did not redirect stderr so we assume you are in debug mode


without redirecting stderr, everything goes to stderr


WITH redirecting stderr, only fatal goes to stderr

david-fritz-gravwell:
When I run as above, _only_ fatal messages come to stderr, and the rest go to the log location. But yes, when I redirect stderr, it trims the output to just the message and structured output.

LGTM

PR #904, opened 2023-11-17
fix ingester stderr logger

kris-watts-gravwell:
This fixes the ingesters so that logs that go to stderr do NOT contain the RFC5424 error message, just the message and values.

Here is what happens now.

If you DO NOT redirect stderr:
1. logs come out to stderr as RFC5424
2. only error, critical, and fatal are emitted to stderr

If you DO redirect stderr:
1. logs going to the redirected stderr file are emitted as RFC5424
2. logs going to actual stderr (the journal) are emitted as raw with no timestamp (so the journal can handle it)
3. only error, critical, and fatal are emitted to stderr

This PR addresses #902 
This PR addresses #901

### DIFF
--- a/ingesters/winevents/main.go
+++ b/ingesters/winevents/main.go
@@ -77,13 +77,12 @@ func main() {
 	if inter {
 		lg = log.New(os.Stdout)
 	} else {
-		lg = log.NewDiscardLogger()
 		e, err := eventlog.Open(serviceName)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "failed to get event log handle: %v\n", err)
 			return
 		}
-		lg.AddLevelRelay(levelLogger{elog: e})
+		lg = log.NewLevelRelay(levelLogger{elog: e})
 	}
 	lg.SetAppname(appName)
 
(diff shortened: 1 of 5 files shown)


ingesters/winevents/main.go
```diff
@@ -189,22 +188,26 @@ type levelLogger struct {
 }
 
 // levelLogger implements the log.LevelRelay interface
-func (l levelLogger) WriteLog(lvl log.Level, ts time.Time, msg []byte) error {
+func (l levelLogger) WriteLog(lvl log.Level, ts time.Time, msg, rawMsg string) error {
 	switch lvl {
 	case log.DEBUG:
 		if debugOn {
-			fmt.Fprintln(os.Stdout, string(msg))
+			fmt.Fprintln(os.Stdout, rawMsg)
```
david-fritz-gravwell, line 195: Do we ever want log messages on stdout? Even debug should go to stderr right?

ingesters/winevents/main.go
```diff
@@ -189,22 +188,26 @@ type levelLogger struct {
...
 	case log.WARN:
-		return l.elog.Warning(1, string(msg))
+		return l.elog.Warning(1, rawMsg)
 	case log.ERROR:
 		fallthrough
 	case log.CRITICAL:
 		fallthrough
 	case log.FATAL:
-		return l.elog.Error(1, string(msg))
+		return l.elog.Error(1, rawMsg)
```
david-fritz-gravwell, line 206: Does the Error() method exit? Do we have a Fatal() method? It doesn't look like it from the fallthroughs above but I thought I'd ask.

david-fritz-gravwell (changes requested):

ingesters/winevents/main.go
kris-watts-gravwell: the `l.elog.Error` ?

ingesters/winevents/main.go
kris-watts-gravwell: no, debug is only done when you are actively debugging with a specific flag.  ITs not part of the traditional log chain

ingesters/winevents/main.go
kris-watts-gravwell: yes, we have a fatal method, and it does exit.  This switch just controls what makes it "out", the caller then terminates

david-fritz-gravwell (approved):
LGTM

PR #915, opened 2023-11-25
Add Pre-extractions to CustomTime Formats

kris-watts-gravwell:
Timegrinder currently has a hard time with unix seconds, or unix nano, or whatever when its embedded deep in some other structure.  This is because.. its just a number.

It also struggles when there are multiple timestamps.

This PR introduces Custom `Extraction-Regex` that can be applied to custom timestamps.

Consider the following data:
```
{"key1": "foobar", "TS": 1700874050}
{"key1": "foobar", "TS": 1700874051}
{"key1": "foobar", "TS": 1700874052}
```

Before we could not capture that, but with this `CustomTime` specification:

```
[TimeFormat "foo"]
	Format="UnixSeconds"
	Extraction-Regex=`"TS":\s*(?P<ts>\d+)`
```

We can now.

### See the before and after in a query
Screenshot from 2023-11-24 18-17-15

### DIFF
--- a/client/types/alert.go
+++ b/client/types/alert.go
@@ -56,7 +56,7 @@ type AlertDefinition struct {
 
 	LastUpdated time.Time `json:"LastUpdated"`
 
-	SaveSearchDuraton time.Duaration `json:"SaveSearchDuration"`
+	SaveSearchDuration time.Duaration `json:"SaveSearchDuration"`
 
 	// Maximum number of events allowed per firing of the alert. This is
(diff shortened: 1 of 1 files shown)


kris-watts-gravwell:
apparently `url.String()` doesn't include query parameters, so the debug post logic and logging has been changed to `RequestURI()` so we get the entire URL.

david-fritz-gravwell (approved):
LGTM

PR #918, opened 2023-11-30
fixing up file follower 

kris-watts-gravwell:
and recovering when lines get written really really slowly or in mass batches with long wait times.


david-fritz-gravwell (approved):
LGTM

Issue #917, opened 2023-11-29
HttpIngester: DebugPosts needs to document the tag being used 


kris-watts-gravwell:
## What is the enhancement to be made?

On the /raw endpoints, if we pick a tag based on some sourcetype translation, it needs to be attached to the debug entry

## Why should we make this change? (Business justification? What problem is the feature trying to solve?)

Because a single endpoint can route many tags, we need to log those

## Any other comments?

For non-raw endpoints we need to grab the first resolved tag and log that.

kris-watts-gravwell:
validated in 5.4.4-beta1

PR #927, opened 2023-12-08
if a tag override is set in the URL, we add it to the debug post log

kris-watts-gravwell:
This PR addresses [issue 917](https://github.com/gravwell/gravwell/issues/917)

### DIFF
--- a/ingesters/HttpIngester/hec.go
+++ b/ingesters/HttpIngester/hec.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/crewjam/rfc5424"
 	"github.com/gravwell/gravwell/v3/ingest"
 	"github.com/gravwell/gravwell/v3/ingest/entry"
 	"github.com/gravwell/gravwell/v3/ingest/log"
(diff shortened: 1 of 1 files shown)


ingesters/HttpIngester/hec.go
```diff
@@ -105,16 +119,26 @@ func (hh *hecHandler) getDefaultTag(h *handler, r *http.Request, ll *log.KVLogge
...
+					param: parameterTag,
+					value: val,
+				}
 			}
 		} else if val = v.Get(parameterSourcetype); val != `` && len(hh.tagRouter) > 0 {
-			tg, ok = hh.tagRouter[val]
+			if tg, ok = hh.tagRouter[val]; ok {
+				override = tagOverride{
+					param: parameterSourcetype,
+					value: val,
```
john-floren-gravwell, line 131: so here's we're returning the value of the sourcetype parameter, not the actual tag we looked up. I'm fine with that if it's what you want, just pointing out that the string we put here is *not* necessarily the tag where the data will go

kris-watts-gravwell: The goal is to log what was provided on the URL for debugging, not necessarily what it will be translated into.

I agree the naming isn't awesome, but calling it "tagSourcetypeWhateverOverride" feels bad?

john-floren-gravwell: works for me

john-floren-gravwell (approved):

PR #931, opened 2023-12-11
Improve handling of filefollow pre-process

john-floren-gravwell:
If a writer is writing really fast, we can sometimes get "stuck" chasing the tail of a given file during the initial sync. This change notes the size of the file when we begin the process and doesn't read past that size, even if the file has been written since then.


kris-watts-gravwell (approved):
LGTM

PR #935, opened 2023-12-15
Add Updated fields to WireUserFile and WireSearchLibrary

john-floren-gravwell:
Other types do it, and it makes sense to pack it. Also, don't set t.Updated to the current time when generating the Thing() from a Wire* object.


kris-watts-gravwell (approved):
LGTM

PR #940, opened 2023-12-19
Add EntriesPerSecond and BytesPerSecond to IngestStats.

john-floren-gravwell:
This makes it simpler for UIs to tell the user how much data is coming in per indexer (or over the whole cluster).


kris-watts-gravwell (approved):
LGTM

PR #941, opened 2023-12-21
adding ability to set no-cache headers on all Gravwell requests

kris-watts-gravwell:
Some APIs cache heavily, this allows for busting those caches when needed.


ashnwade (approved):
LGTM

Issue #942, opened 2023-12-21
Require TLS 1.2 or better on HTTP ingester and simple relay.


kris-watts-gravwell:
SSL, TLS 1.0 and 1.1 should not be used anywhere.

Force `MinVersion: tls.VersionTLS12` on all TLS listeners

kris-watts-gravwell:
confirmed fixed and closing

PR #943, opened 2023-12-22
Clean up some TLS min versions and default configs

kris-watts-gravwell:
Force TLS 1.2 or better for all servers, simple relay already did this, but now HTTP ingester does too

Also clean up some TLS related configs and bad settings.

This PR addresses https://github.com/gravwell/gravwell/issues/942


david-fritz-gravwell (approved):
LGTM

Issue #945, opened 2024-01-02
TimeGrinder: Failing test on Syslog output


ashnwade:
## What are the steps to reproduce this issue?

Run tests on monorepo

## What happens?

![image](https://github.com/gravwell/gravwell/assets/107071905/ceeffd59-5f29-48e2-86ba-8a1f3d55f183)

## What were you expecting to happen?

Tests pass with expected output

## Other comments?

This may be a real bug, not just a failed test

PR #947, opened 2024-01-03
fix bad test on timegrinder that will always fail from Jan 1 to Jan 8 every year

kris-watts-gravwell:
We have a couple of formats that do not contain years, so there is some wonky logic to try and figure out what the appropriate year is.

For example, if we only see the date Jan 10, then we need to figure out WHICH Jan 10.  So we basically add the current year and check if its too far in the future (to account for clock drift and timezone garbage).  If it is, then we take the previous year.  It is not great, but it generally handles the crappy time format.


ashnwade:
Addresses https://github.com/gravwell/gravwell/issues/945

ashnwade (approved):
LGTM

PR #948, opened 2024-01-03
fix unsafe usage of scanner Bytes() method

kris-watts-gravwell:
There were a few ingesters that were not Cloning the response of calls to a `bufio.Scanner.Bytes()` call.  It is extremely difficult to manifest the behavior on any ingester but the HTTP ingester due to the raciness of getting data into the ingest muxer. 

This fixes usage across all ingesters and also cleans up some kind of sloppy append() clone falls.

Addresses https://github.com/gravwell/gravwell/issues/949

Oops, this also includes the following:
1. tweak to megajson generator to simulate timestamps that suck (deeply embedded unix timestamp)
2. tweak to URL-Parameter-Attach to support a wildcard which means attach everything you see (but tag)
3. fix HEC compatible listener so that it supports custom timestamp formats (for /raw endpoint)

### DIFF
--- a/generators/gravwellGenerator/jsongen.go
+++ b/generators/gravwellGenerator/jsongen.go
@@ -28,12 +28,12 @@ type datum struct {
 }
 
 type megaDatum struct {
-	TS        string         `json:"time"`
 	Account   Account        `json:"account"`
 	Class     int            `json:"class"`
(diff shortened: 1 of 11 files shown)


david-fritz-gravwell (approved):
LGTM

PR #950, opened 2024-01-04
add new mega JSON data type to gravwell generator

kris-watts-gravwell:
new generator type that can flex all the craziness when records are huge.


ashnwade (approved):
LGTM

PR #953, opened 2024-01-08
adding ingest stats tail members

kris-watts-gravwell:
This adds in 4 new members to the IngestStats structure so that we can report a running "tail" of ingest on the stats pages.


evan-gravwell (approved):
LGTM!

PR #960, opened 2024-01-11
adding generator chaos mode and adding some profiling systems

kris-watts-gravwell:
we can use this profile thingy to include in all ingesters, it makes everything cleaner


david-fritz-gravwell (approved):
LGTM

PR #962, opened 2024-01-11
HTTP Ingester cleanup

kris-watts-gravwell:
Just some additional safety stuff to maximize the likelihood that we will get to perform socket reuse and also make sure that the Http server shuts down gracefully without terminating connections.


david-fritz-gravwell (approved):
LGTM

PR #966, opened 2024-01-12
More chaos on generator and add sanity checking ingest reader/writer connection configs

kris-watts-gravwell:


david-fritz-gravwell (approved):
LGTM

PR #973, opened 2024-01-19
s3/sqs ingester: delete messages in the same batch as we receive.

### LINKED ISSUE #969: s3 ingester -- ack/delete SQS messages in batch, not one by one

david-fritz-gravwell:
Fixes #969


kris-watts-gravwell (approved):
LGTM

PR #976, opened 2024-01-23
adding config option to load secrets from a file rather than a hardco…

kris-watts-gravwell:
…ded string

<!--

### DIFF
--- a/ingest/config/env.go
+++ b/ingest/config/env.go
@@ -10,13 +10,18 @@ package config
 
 import (
 	"bufio"
+	"bytes"
 	"errors"
+	"fmt"
+	"io"
 	"os"
 	"reflect"
 	"strings"
 )
 
-const ()
+const (
+	maxFileValueSize int64 = 1024 * 16 // secrets and the like cannot be bigger than 16k when loaded from a file
+)
 
 var (
 	errNoEnvArg     = errors.New("no env arg")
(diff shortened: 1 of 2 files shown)


ingest/config/env.go
```diff
@@ -491,3 +496,42 @@ func loadEnvVarList(lst *[]string, envName string) error {
...
+
+	//if we hit here, it means we have a size that is ok, load a buffer and read it
+	buff := make([]byte, sz)
+	_, err = io.ReadFull(fin, buff) //attempt the read
+	fin.Close()                     //close the file, because we don't need it anymore
+	if err != nil {
+		return fmt.Errorf("failed to read complete string from %q %w", pth, err)
+	}
+	//trim nulls, newlines, and carriage returns from the file because it will be a nightmare to debug
+	*val = string(bytes.Trim(buff, "\n\t\r\x00"))
```
david-fritz-gravwell, line 534: Does TrimSpace not do this in a more idiomatic way? Or is this to cover the null character case?

david-fritz-gravwell (changes requested):

ingest/config/env.go
kris-watts-gravwell: and i don't want to trim spaces, because we have a few people using them in their secrets... :/

kris-watts-gravwell:
@david-fritz-gravwell PTAL

david-fritz-gravwell (approved):
LGTM

PR #977, opened 2024-01-23
adding token based routing on HEC compatible listeners

kris-watts-gravwell:
This allows definitions like so:

```
[HEC-Compatible-Listener "testing"]
	URL="/services/collector" #If URL is omitted, the default is set to /services/collector
	TokenValue="test" #set the access control token
	Tag-Name=hec
	Debug-Posts=true
	Routed-Token-Value=`test2:hec2`
	Routed-Token-Value=`"this is a complex token":hecX`
	Tag-Match=`"foobar":hec99`
	Tag-Match=`foobar2:"hec88"`

[HEC-Compatible-Listener "testing2"]
	URL="/services/collector2" #If URL is omitted, the default is set to /services/collector
	TokenValue="test" #set the access control token
	Tag-Name=hec2
```

Basically enabling routes for tags to be controlled by `sourcetype` overrides OR tokens.  This is required because many external services that expect to talk to HEC do not know how to throw at alternate URLs.  They ONLY know about the hardcoded `/services/collector`

Notice that we can now do tag routing using `tag` parameter overrides, `sourcetype` parameter overrides, or just by pairing a tag to a token value.

### DIFF
--- a/ingesters/HttpIngester/hec.go
+++ b/ingesters/HttpIngester/hec.go
@@ -19,6 +19,7 @@ import (
 	"net"
 	"net/http"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


david-fritz-gravwell (approved):
LGTM

PR #985, opened 2024-01-30
increase service timeout on stopping Http ingester

kris-watts-gravwell:
give the ingester more time to stop


david-fritz-gravwell (approved):
LGTM

PR #986, opened 2024-01-30
tweaking some configs to allow generator to be used to just test

kris-watts-gravwell:
This allows the generator to be used to test tag negotiation and connections without sending entries


david-fritz-gravwell (approved):
LGTM

PR #987, opened 2024-01-30
tweak the validate function a bit so it can be used by internal components too

kris-watts-gravwell:
Validate config function was too strict and trying to assert that everything was an ingester, this broke validate on other components.

### DIFF
--- a/ingest/config/validate/validate.go
+++ b/ingest/config/validate/validate.go
@@ -38,7 +38,7 @@ func ValidateConfig(fnc interface{}, pth, confdPath string) {
 
 // ValidateIngesterConfig behaves same as ValidateConfig but also asserts that the provided config
 // can return an IngestBaseConfig object.
-func ValidateIngesterConfigBase(fnc interface{}, pth, confdPath string) {
+func ValidateIngesterConfig(fnc interface{}, pth, confdPath string) {
 	validateConfig(fnc, pth, confdPath, true) // this is used by ingesters
 }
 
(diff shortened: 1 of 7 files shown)


ingest/config/validate/validate.go
```diff
@@ -31,8 +31,19 @@ var (
...
 // The first value is an opaque object (hence the reflect voodoo).  The second value is an error object
 // the point of this function is to make it easy for ingester writers to just hand in their GetConfig function
-// and the two paths and get a "go/no go" on the configurations
+// and the two paths and get a "go/no go" on the configurations.
+// ValidateConfig also asserts that the provided config object can return an IngestBaseConfig object.
 func ValidateConfig(fnc interface{}, pth, confdPath string) {
+	validateConfig(fnc, pth, confdPath, true) // this is used by ingesters
+}
+
+// ValidateConfigBase behaves the same as ValidateConfig but does not assert that the provided config
```
john-floren-gravwell, line 40: It seems kind of confusing that "ValidateConfig**Base**" is the one that *doesn't* assert that the config can return an Ingest**Base**Config object...

kris-watts-gravwell: I chose this because it has less assertions.  The other one asserts it's an ingester config

john-floren-gravwell (approved):
lgtm

david-fritz-gravwell (approved):
LGTM

PR #992, opened 2024-02-02
regression in HEC that is fixed

kris-watts-gravwell:


ashnwade (approved):
John reviewed and approved the change for v5.4.4-3. This approves the change for v5.4.5

PR #994, opened 2024-02-02
First pass at fields to enable write access.

john-floren-gravwell:
We want to do query library first, which means we get to start with Things, the most complicated part. This adds information about writers to the Thing type.

It also adds a field to the WireSearchLibrary which is used to communicate to the client what the current user is allowed to modify on any given item; only owners and admins can change sharing fields, but anyone in the new WriteAccess lists can change the other stuff.


client/types/things.go
```diff
@@ -410,14 +422,21 @@ func (uf *UserFile) JSONMetadata() (json.RawMessage, error) {
...
 	ThingHeader
 	SearchLibrary
-	Updated time.Time
+	WritableFields EnabledFields
+	Updated        time.Time
+}
+
+type EnabledFields struct {
+	Details bool
+	Sharing bool
```
evan-gravwell, line 431: It would be nice to get info on if the current user can delete the thing, too. [Mentioned yesterday on the issue](https://github.com/gravwell/backend-issues/issues/5500#issuecomment-1928088171).

evan-gravwell (changes requested):
Code: LGTM

I made a comment on EnabledFields, noting a comment I made yesterday [on the issue](https://github.com/gravwell/backend-issues/issues/5500#issuecomment-1928088171). Curious if you have any thoughts on implementing similar @john-floren-gravwell .

kris-watts-gravwell (approved):
LGTM

PR #1012, opened 2024-02-18
fix imports for v4 and search store, just yolo it

kris-watts-gravwell:


corey-thuen-gravwell (approved):
death to v3. long live v4

PR #1013, opened 2024-02-20
Try to avoid sending a nil WriteAccess.GIDs field for Things.

john-floren-gravwell:
It makes Javascript clients unhappy; if we fix it here, we should get improvements for free in a lot of our APIs.


kris-watts-gravwell (approved):
LGTM

PR #1024, opened 2024-03-07
make collect ingester not bark every time you restart it

kris-watts-gravwell:
don't return an error if we are leaving a routine due to a context.Cancel


david-fritz-gravwell (approved):
LGTM

PR #1026, opened 2024-03-08
Fix logic on fsnotify Removed events.

### LINKED ISSUE #1025: File Follower: "beginning of line" delimiter can cause a dropped buffer and data loss
## What are the steps to reproduce this issue?

1. Use file follower with a regex delimiter that uses the beginning of a line as the delimiter (something like `<\d+>` for syslog
2. Start executing and then remove a tracked file within 5 seconds of writing to it

## What happens?

The last record in the file will not be ingested.

## What were you expecting to happen?

Write the remaining buffer to the indexer.

david-fritz-gravwell:
Fixes #1025


john-floren-gravwell (approved):
makes sense

PR #1040, opened 2024-03-23
remove websockets from parse search request

kris-watts-gravwell:
Do not use websockets for search parsing.


john-floren-gravwell (approved):
ding dong the socket's dead

Issue #981, opened 2024-01-26
Negotiate Tags on IngestMuxer can race and make scary error that isn't fatal or really that big of an error


kris-watts-gravwell:
## What are the steps to reproduce this issue?

1. Setup cluster with many upstream connections (16+)
2. Setup ingester to connect to them all (Http ingester works)
3. Start the ingester with logs of configs and tags to negotiate

## What happens?

When we call waitforhot we only wait for one indexer to come up, because that is all we need to start, but that means when we negotiate tags there is a chance for other indexer connections to not be ready yet.

This then throws a super scary error log.

The system recovers just fine, nothing is lost, no real fatal errors happen, but the log is too scary for what is happening

## What were you expecting to happen?

WaitForHot should always wait some amount of time to give all connections a CHANCE to come up, don't just race to the first one.

## Any logs, error output, etc?

...

## Any other comments?

I think the error on that negotiate tag should also dictate the log level, not ready should not cause an error, a warning AT MOST.

## What versions of software are you using?

5.4.4

PR #1041, opened 2024-03-25
Improve error logging on NegotiateTag

john-floren-gravwell:
If an ingest connection isn't running, it's very much a non-issue, so throw an info log and move on.

Other errors get a warning, but it's still not a big deal.

Addresses #981


kris-watts-gravwell (approved):
LGTM

PR #1044, opened 2024-03-28
Add functions to get alerts which use a specific consumer/dispatcher

john-floren-gravwell:
Also brings in a little function we can use to resolve, in the context of a particular user, exactly which scheduled search/script/flow is being referenced by a GUID.


david-fritz-gravwell (approved):
LGTM

PR #1045, opened 2024-03-29
Refine the alert filtering APIs a little bit

john-floren-gravwell:
As future-proofing, require the requester to specify which sort of dispatcher/consumer they're asking about.


david-fritz-gravwell (approved):
LGTM

PR #1056, opened 2024-04-09
Save backup and create new state file in filefollower when the state file is unreadable

### LINKED ISSUE #1032: file follower: don't stop on a corrupt state file
Currently we refuse to start if the state file can't be read in file follower. Instead, we should copy the corrupt file to a .backup or something, and startup with a new, empty state file.

david-fritz-gravwell:
Fixes #1032

### DIFF
--- a/filewatch/filters.go
+++ b/filewatch/filters.go
@@ -20,6 +20,8 @@ import (
 	"github.com/gravwell/gravwell/v3/ingest/log"
 )
 
+const backupSuffix = ".backup"
+
 type filter struct {
 	FollowerEngineConfig
 	bname string //name given to the config file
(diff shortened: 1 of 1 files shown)


john-floren-gravwell (approved):

filewatch/filters.go
```diff
@@ -824,8 +826,31 @@ func initStateFile(p string) (fout *os.File, states map[FileName]*int64, err err
...
+			fout.Close()
+
+			// find a suitable backup filename
+			fname := p + backupSuffix
+			if _, err := os.Stat(fname); err != nil {
+				// we have to start counting
+				var count int
+				for {
+					fname = fmt.Sprintf("%v%v%d", p, backupSuffix, count)
+					if _, err := os.Stat(fname); os.IsNotExist(err) {
```
kris-watts-gravwell, line 840: set a maximum here.  If by some crazy situation you can stat the first name but not the one with the number (symlinks... or whatever... then this will never stop.

Rather than attaching a number, I would recommend a backup file that appends a timestamp, if that fails... quit.

kris-watts-gravwell (changes requested):
change the backup file loop to just appending the current timestamp, if that fails quit.

filewatch/filters.go
david-fritz-gravwell: numbers are easier to grok, and the files have timestamps I'd argue. Limits definitely.

david-fritz-gravwell:
@kris-watts-gravwell PTAL

filewatch/filters.go
kris-watts-gravwell: sounds good to me.

kris-watts-gravwell (approved):
LGTM

PR #1071, opened 2024-04-19
throw the insufficient storage head on the health check if the cache is full

kris-watts-gravwell:
basically check the willBlock call on the health check too


david-fritz-gravwell (approved):
LGTM

PR #1078, opened 2024-04-24
Add GetSearchAgentCheckin function

john-floren-gravwell:
Get when the searchagent last checked in.


david-fritz-gravwell (approved):
LGTM

PR #1087, opened 2024-05-08
enable disk backed cache in default file follower config, but only in fail mode

david-fritz-gravwell:


kris-watts-gravwell (approved):
LGTM

PR #1091, opened 2024-05-15
Improve handling of MFA in UserDetails

john-floren-gravwell:
We'll ship information about what's enabled, but leave out the sensitive stuff.


kris-watts-gravwell (approved):
LGTM

PR #1093, opened 2024-05-17
remove PIDFile specifications from service files

kris-watts-gravwell:
These aren't used and systemd now barks about it.


david-fritz-gravwell (approved):
LGTM but just the 2?

Issue #1068, opened 2024-04-17
HTTP Ingester: embedded HTTP logger is throwing non-RFC5424 messages


kris-watts-gravwell:
## What are the steps to reproduce this issue?

1. Fire up HTTP ingester connected to system
2. Enable HTTPS/TLS keys
3. Hit the server with some garbage (causing a TLS error)

## What happens?

Message is emitted in its raw form from the HTTP server/TLS handler that is not well formed

## What were you expecting to happen?

wrap the HTTP server errors in our logger so that messages are always well formed

## Any logs, error output, etc?

```
2024/04/17 19:28:32 server.go:3212: http: TLS handshake error from 7.8.9.0:54342: read tcp 1.2.3.4:443->7.8.9.0:54342: i/o timeout
```

## Any other comments?

We also may not really even WANT the embedded HTTP server messages if we aren't in `INFO` level, its almost always garbage.

## What versions of software are you using?

5.4.7

david-fritz-gravwell:
LGTM

PR #1104, opened 2024-06-05
Issue1068

kris-watts-gravwell:
Added some functions on the logging package so that we can hand back a Handler for the `log/slog` standard library package.  This lets us hand in a standard library logger into the HTTP server and get much better logs out (with level handling).

This PR addresses https://github.com/gravwell/gravwell/issues/1068


kris-watts-gravwell:
I forced a panic in a handler, output is clean now:
Screenshot from 2024-06-05 10-35-17

john-floren-gravwell (approved):

PR #1107, opened 2024-06-07
Add SSOUser field to UserDetails struct

john-floren-gravwell:
This way we can track which users are created/managed via SSO.


david-fritz-gravwell (approved):
LGTM

PR #1108, opened 2024-06-07
[FEAT](N/A): incorpoorate weave output utility 

rflandau:
This PR does not address an issue.

This pull request incorporates the [weave](https://github.com/rflandau/weave) package, written to support [gwcli](https://github.com/rflandau/gwcli), as a Gravwell utility. It is able to consume arbitrary arrays of a struct and translate them into the desired output format (CSV, JSON, lipgloss table), while limiting which columns/fields are printed.


john-floren-gravwell (changes requested):
Can you add copyright notices to the tops of your new source files? Just do it the same way as e.g. the client library. Thanks!

john-floren-gravwell (approved):

PR #1109, opened 2024-06-10
Expose custom endpoint in AWS config for kinesis.

### LINKED ISSUE #1059: Kinesis Ingester: Support for Custom Endpoint
## Who was the customer who requested this feature?

Discord User

## What is the feature to be added?

Support for a Custom Endpoint in the Kineses Ingester.
   (For Private VPC's with no external network access,   AWS provides the ability to create a custom endpoint for services that will allow you to access the AWS service without requiring any connections to outside networks)

Adds support for additional AWS configurations

Adding the functionality available within the AWS SDK will allow for greater support for a variety of AWS  deployment designs.

## Any other comments?

I may have located the SDK function required, but would need someone more familiar with the code to verify.

func [EndpointResolverFromURL](https://github.com/aws/aws-sdk-go-v2/blob/service/kinesis/v1.27.4/service/kinesis/endpoints.go#L56) [¶](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/kinesis#EndpointResolverFromURL) added in v1.1.0

func EndpointResolverFromURL(url [string](https://pkg.go.dev/builtin#string), optFns ...func(*[aws](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws).[Endpoint](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws#Endpoint))) [EndpointResolver](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/kinesis#EndpointResolver)

EndpointResolverFromURL returns an EndpointResolver configured using the provided endpoint url. By default, the resolved endpoint resolver uses the client region as signing region, and the endpoint source is set to EndpointSourceCustom.You can provide functional options to configure endpoint values for the resolved endpoint.

...

david-fritz-gravwell:
Fixes #1059


kris-watts-gravwell (approved):
LGTM

PR #1111, opened 2024-06-12
Tweaks migrate tool to work with Splunk free licenses

john-floren-gravwell:
If the user doesn't set the `Token` parameter on the Splunk config block, we'll assume it's a free license and omit the auth header entirely.

Also adds Insecure-Skip-TLS-Verify on the Splunk config block so we can handle Splunk servers with self-signed certs.


kris-watts-gravwell (approved):
Seems simple enough!

PR #1119, opened 2024-07-09
Add a field to alerts indicating if the events will ingest

john-floren-gravwell:
If IngestBlocked is true, the owner of the alert is not allowed to ingest data (events will not be recorded)


david-fritz-gravwell (approved):
LGTM

PR #1132, opened 2024-07-22
proposed changes to the way the ingest Muxer throws configuration state updates

kris-watts-gravwell:
This PR does a few things around how we throw config blocks on ingesters:

1. Reduce the frequency from every 5 seconds to 30s
2. Allows for up to 5 minutes between throws if nothing has changed
3. Only throw config block if something is worth reporting
4. Aggressively document some gotchas on the ingester config block reporting stuff


john-floren-gravwell (changes requested):
Make a new branch & submit PR to avoid extra commits slipping in

PR #1134, opened 2024-07-22
bring in ingesterstate tweaks to control how and when config blocks are thrown

kris-watts-gravwell:
This PR addresses no specific issue and is for readability and performance only.


david-fritz-gravwell (approved):
LGTM

Issue #1138, opened 2024-07-23
Open Source CLI Client


rflandau:
## Overview

[Link to the current repo](https://github.com/rflandau/gwcli).

This project is to create an open-source, CLI client usable via an interactive TUI or scriptable shell commands.

### Background

The current CLI is unsupported and could do with a makeover. While it has relative feature parity with the web interface, it is unpleasant to use and built on an unstable foundation. On top of this, it is currently closed-source.

Rather than attempt to overhaul and upgrade the current CLI, this project seeks to wholesale replace it.

## Milestones

- [x] 1. Invoking the CLI spawns an interactive, shell-like prompt
- [x] 2. CLI builds a nested command tree with actions as leaves and both local and global flags
- [x] 3. Actions can be invoked interactively or via the user's shell
- [x] 4. Querying and result viewing
- [x] 5. Secure login
- [ ] 6. Actions are built on a boilerplate for easy extension
- [ ] 7. Feature parity with old CLI
- [ ] 8. OPTIONAL: Further Improvements

### Path to Production

gwcli has a stable, extensible framework and a small library of genericized tools for rapidly expanding functionality. The majority of the work these past 12 weeks has been on creating a foundation upon which it is easy to maintain and upgrade Gravwell's CLI client. This foundation is flexible, fast, and simple to modify or extend. Its interactive and non-interactive interfaces are tighlty coupled for unified usage. The primary functionality, querying, is fully implemented and provides a featureful interface with many of the amenities of the web GUI.

For gwcli to be production-ready, it needs to be brought up to feature parity with the current cli client (Milestone 7). This should be primarily supported by the scaffold package, which provides genericized actions for rapidly bolting on new functionality. Milestone 6 can be completed once the edit scaffold is in place. 

Milestone 8 is extra, nice-to-have features, such as query syntax highlighting and *dynamic* tab auto-completion (rather than the semi-static currently available).

### Milestone 1

**Completed**

gwcli has a shell-like interface for interactive use, complete with interactive action invocation, contextual help, tree navigation and presence, command history, and tab auto-completion.

### Milestone 2

**Completed**

gwcli is a [Cobra](cobra.dev) command tree that is self-building, easily extensible, and highly malleable.

A set of global flags is defined on the root node. These are available to every command. Each action can also have local-only flags defined. These flags are usable in interactive and non-interactive mode (see Milestone 3).

### Milestone 3

[...]

PR #1142, opened 2024-07-31
FEAT(#1138): incorporate new cli alpha (gwcli)

rflandau:
This PR does not address an issue. It incorporates the current, alpha version of open source CLI referenced in issue #1138.


john-floren-gravwell:
I suggest we move this out of utils/ and into its own top-level gwcli/ dir

rflandau:
I shifted it up to root. However, it is still its own go project in root, with its own go.mod. This allows it to be built independently, but means the dependencies are not merged. Will that be an issue?

john-floren-gravwell:
I think we can worry about merging go.mod later, if we decide we want that. Kris has some thoughts about organizing the libraries which I'll let him comment on

john-floren-gravwell (approved):

PR #1180, opened 2024-08-19
Add some fields to improve sharing of searches

john-floren-gravwell:
This will allow us to share search results with a list of groups or globally, same as just about everything else in the system.


kris-watts-gravwell (approved):
LGTM

PR #1192, opened 2024-08-21
ignored method on notifications and simplify logic

kris-watts-gravwell:
adding a method and cleaning up the logic so we don't fire on Zero at all


david-fritz-gravwell (approved):
LGTM

PR #1194, opened 2024-08-21
fixing up usage of notifications APIs

kris-watts-gravwell:
we didn't have an admin method to get all notifications and we were wrapping the notifications URL very stupidly.


david-fritz-gravwell (approved):
LGTM

PR #1199, opened 2024-08-22
check values and omit anything that cannot be encoded in JSON

kris-watts-gravwell:
don't puke on JSON encode errors because JSON can't encode many float64's


david-fritz-gravwell (approved):
LGTM

Issue #1200, opened 2024-08-22
preprocessor/plugin: start function can panic


david-fritz-gravwell:

john-floren-gravwell:
Can I get your test plugin/config to verify this?

david-fritz-gravwell:
```
package main

import (
    "fmt"
    "gravwell"

    "github.com/gravwell/gravwell/v3/ingest/entry"
)

const (
    PluginName = "example"
)

var foo []error

func Start() error {
	return foo[0]	
}


func Close() error {
    return nil
}

func Config(cm gravwell.ConfigMap, tgr gravwell.Tagger) error {
    return nil
}

func Flush() []*entry.Entry {
    return nil
}

func Process(ents []*entry.Entry) ([]*entry.Entry, error) {
    for _, v := range ents {
        fmt.Println(v.TS, string(v.Data))
    }
    return nil, nil
}

func main() {
    if err := gravwell.Execute(PluginName, Config, Start, Close, Process, Flush); err != nil {
        panic(err) //panic on failure, generally not needed
    }
}
```

@john-floren-gravwell -- basically just cause a panic in Start()

john-floren-gravwell:
lgtm

PR #1201, opened 2024-08-22
preprocessor/plugin: catch panics in execute and start functions

david-fritz-gravwell:


david-fritz-gravwell:
@kris-watts-gravwell we needed to get that error in start out in a different way and I removed the println. PTAL

david-fritz-gravwell:
Fixes #1200

PR #1238, opened 2024-09-06
Add Error fields to SearchInfo and SearchCtrlStatus

john-floren-gravwell:
This allows us to pass along information about why a query failed.


david-fritz-gravwell (approved):
LGTM

PR #1244, opened 2024-09-11
File follower and windows file lock fun

kris-watts-gravwell:
This fixes two main bugs:
1. The file follower could exit on error if a file is rotated out during the Catchup call.  This seems like one hell of a race to hit, but I am able to trigger it on the regular with Windows.  Never could in Linux.
2. Windows file follower calls its equivelent of a "stat" with SHARE attributes, because apparently "stat" in windows grabs a file handle in an exclusive way and holds it for a long time... because reasons.

### DIFF
--- a/filewatch/fwork_windows.go
+++ b/filewatch/fwork_windows.go
@@ -33,7 +33,6 @@ func getFileId(f *os.File) (id FileId, err error) {
 // getFileIdFromName is a windows version of stat that returns something roughly equal to a device and inode.
 // there is voodoo, we all hate it... yet here we are...
 func getFileIdFromName(name string) (id FileId, err error) {
-	var attrib *syscall.SecurityAttributes
 	p, lerr := syscall.UTF16PtrFromString(name)
 	if lerr != nil {
 		err = lerr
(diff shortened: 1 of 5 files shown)


filewatch/fwork_windows.go
```diff
@@ -30,22 +30,32 @@ func getFileId(f *os.File) (id FileId, err error) {
...
-		err = fmt.Errorf("Failed to open %s to check FileID: %v", name, lerr)
+		err = fmt.Errorf("Failed to open %s to check FileID: %w", name, lerr)
 		return
 	}
 	defer syscall.CloseHandle(h)
 	var bhfi syscall.ByHandleFileInformation
 	if err = syscall.GetFileInformationByHandle(h, &bhfi); err != nil {
-		err = fmt.Errorf("Failed to get FileID for %s: %v", name, err)
+		if os.IsNotExist(err) {
+			err = os.ErrNotExist
```
john-floren-gravwell, line 56: This is extraneous given the way you re-declare err on line 58

filewatch/fwork_windows.go
```diff
@@ -30,22 +30,32 @@ func getFileId(f *os.File) (id FileId, err error) {
...
+		}
+		err = &os.PathError{Op: "open", Path: name, Err: os.ErrNotExist}
```
john-floren-gravwell, line 58: Or maybe you wanted to say `Err: err` here?

filewatch/fwork_windows.go
```diff
@@ -97,6 +110,9 @@ func createDeletableFile(fpath string) (*os.File, error) {
 	h, err := syscall.CreateFile(p, syscall.GENERIC_READ|syscall.GENERIC_WRITE, shared,
 		attrib, syscall.CREATE_NEW|syscall.OPEN_ALWAYS, syscall.FILE_ATTRIBUTE_NORMAL, 0)
 	if err != nil {
+		if os.IsNotExist(err) {
+			err = &os.PathError{Op: "open", Path: fpath, Err: os.ErrNotExist}
+		}
 		return nil, err
```
john-floren-gravwell, line 116: did you want to do the os.PathError dance here too?

filewatch/fwork_windows.go
kris-watts-gravwell: good catch

filewatch/fwork_windows.go
kris-watts-gravwell: the problem is that for some damn reason after i wrap that error in path error it wasn't tripping.

The standard library does something similar in windows, where it checks the error, and if its X or Y return the constant stdlib error.  I am doing something similar.

kris-watts-gravwell: cleaning it up a bit though

john-floren-gravwell (approved):

PR #1262, opened 2024-09-20
ci: Run tests on any PR

michael-wisely-gravwell:
This PR addresses no issue.

We're currently limiting test jobs to PRs that target certain branches. We **should** be running tests on **all** PRs that are opened. That is, we should never merge code that hasn't had tests run and pass.

This PR proposes 

- removing the branch inclusion list, so that all PRs will have tests run on them
- whitespace cleanup


ashnwade (approved):
LGTM

PR #1277, opened 2024-09-29
perf(timegrinder): Optimize `DebugMatch` by only extracting matched portion

gabe565:
Currently, `DebugMatch` will perform both a match and an extract against the full input. Since `Match` returns a start/end index, a smaller slice can be passed to `Extract`. This results in less work for the regexp engine.

The speed difference is highly dependent on the length of the input data and the index of the timestamp, but with my test data, this gets a 3% speed improvement.


kris-watts-gravwell:
retargeting to next-patch and updating branch.

LGTM, i'll get this approved to run through CI/CD.

Welcome and thank you for the the PR!

kris-watts-gravwell (approved):
LGTM, good catch!

david-fritz-gravwell (approved):
LGTM

kris-watts-gravwell:
we are planning a release very soon and will do a merge down that will get 1.21.X out of the fold so that these CI/CD tasks complete.  Enabling automerge so everything just happens when that happens.

david-fritz-gravwell:
@ashnwade this will require a force merge as we aren't updating the go version in next-patch just yet.

david-fritz-gravwell:
Or what Kris said.

gabe565:
Thanks for reviewing this! Sounds good. If you guys end up bumping the version then I'll need to rebase this branch, so just let me know if you need that (I also allowed edits by maintainers so feel free to do it yourselves if you prefer)

Issue #1275, opened 2024-09-27
Timegrinder: `time/tzdata` should not be imported by libraries


gabe565:
## What needs to be done?
Timegrinder currently [imports `time/tzdata`](https://github.com/gravwell/gravwell/blob/cb303dfaffb8ddaecc69e2c48d153f9bd4aa2818/timegrinder/timeGrinder.go#L22-L23) which adds ~450 KB to the final build. Since there's no way to opt-out, the Godocs for this package [suggest only installing it in main packages](https://pkg.go.dev/time/tzdata):
> This package should normally be imported by a program's main package, not by a library. Libraries normally shouldn't decide whether to include the timezone database in a program.

Would it be possible to remove this import from Timegrinder and only add it to the main packages that depend on Timegrinder? This would allow for smaller binaries in controlled environments where the `tzdata` package can be installed externally (like in a container).

## Why is this task needed? 
I have a use-case for Timegrinder where I'd prefer to keep the binary small. I probably shouldn't fret over 450 KB, but since best practices recommend not embedding in libraries, I thought I'd open this.

## Any other comments?
I also just want to say that Timegrinder is a really impressive library! Thank you so much for open sourcing it.

john-floren-gravwell:
lgtm

PR #1286, opened 2024-10-09
remove tzdata import from timegrinder and add it to all our main packages

kris-watts-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/1275


kris-watts-gravwell:
going to have to get main merged up before govlun stops barking

Issue #1291, opened 2024-10-16
file follow: regex delimited files can drop the last entry in a file


david-fritz-gravwell:
## What are the steps to reproduce this issue?

1. Create a regex delimited stanza, something like `foo...`
2. Ingest some data that hits on this.
3. Do something like delete/rotate the file.

## What happens?

The last hit in the file doesn't appear to be ingested.

## What were you expecting to happen?

Ingest.

## What versions of software are you using?

5.4.11 (at least).

john-floren-gravwell:
lgtm, fritz and kris have tested mac and windows respectively

PR #1316, opened 2024-10-17
Do a final call to processLines on the way out of a file follow

john-floren-gravwell:
Addresses #1291


kris-watts-gravwell (approved):
LGTM

PR #1336, opened 2024-10-25
Tweaks to the generator

kris-watts-gravwell:
This PR addresses no issue.

The following was tweaked on the gravwellGenerator

- Fixed the flag that enabled/disabled bypassing TLS validation
- Added a flag to generate random SRC values on entries
- reworked constant generation so that you can override orthogonality using ENV variables
- reworked datatype management so that its easier to drop in a file that includes more datatypes without forcing you to include it in the main repo


david-fritz-gravwell (approved):
LGTM

PR #1367, opened 2024-11-04
Make ingesters not even try to write a state if its too large

kris-watts-gravwell:
deal with ingesters (namely federators) that have WAY too much state

this will first try to trim configs and metadata
then it will trim children
then it will give up and throw a log


john-floren-gravwell (approved):

david-fritz-gravwell (commented):
LGTM

Issue #1036, opened 2024-03-19
jsonextract preprocessor: should drop entries that aren't JSON


corey-thuen-gravwell:
Did some testing against the Drop-Misses directive on jsonextract preprocessor. It doesn't appear to be working as advertised. When set to true or false, the preprocessor will forward on records that are not JSON in their entirety, and it will also forward on properly jsonextracted records that do not contain all of the fields in the Extractions= directive.

david-fritz-gravwell:
LGTM

PR #1382, opened 2024-11-06
Address Issue 1036

kris-watts-gravwell:
If we are doing drop misses or strict extraction, pre-validate that we even have a JSON object.

This PR addresses https://github.com/gravwell/gravwell/issues/1036


kris-watts-gravwell:
fixed the test

david-fritz-gravwell (approved):
LGTM

Issue #1278, opened 2024-10-04
client: IngestEntries does not include EVs


kris-watts-gravwell:
## What are the steps to reproduce this issue?

1.  Using the OSS client, run a query and grab data that has intrinsic EVs
2. Use the client to then re-ingest using `IngestEntries()`

## What happens?

The re-ingested entries do not have the EVs attached

## What were you expecting to happen?

EVs are attached

## Any logs, error output, etc?

nope

## Any other comments?

Not sure if we want the default behavior to always be attach, or if this needs a new API.  Where Intrinsic EVs are getting more and more usage i think the default should be just include them by default

## What versions of software are you using?

5.6.0-beta6

david-fritz-gravwell:
LGTM with this program:

```
package main

import (
	"log"
	"net"

	"github.com/gravwell/gravwell/v3/ingest"
	"github.com/gravwell/gravwell/v3/ingest/entry"
)

var (
	dst          = "tcp://127.0.0.1:4023"
	sharedSecret = "IngestSecrets"
	simple_tags  = []string{"testtag"}
)

func main() {
	igst, err := ingest.InitializeConnection(dst, sharedSecret, simple_tags, "", "", false)
	if err != nil {
		log.Fatalf("Couldn't open connection to ingester: %v", err)
	}
	defer igst.Close()

	tagid, ok := igst.GetTag(simple_tags[0])
	if !ok {
		log.Fatal("couldn't look up tag")
	}

	ent := entry.Entry{
		TS:   entry.Now(),
		SRC:  net.ParseIP("127.0.0.1"),
		Tag:  tagid,
		Data: []byte("This is my test data!"),
	}
	ev, err := entry.NewEnumeratedValue("foo", "test")
	if err != nil {
		log.Fatal(err)
	}
	ent.AddEnumeratedValue(ev)

	igst.WriteEntry(&ent)
}
```

david-fritz-gravwell:
I'm wrong, the above program uses an ingest muxer. That already worked.

david-fritz-gravwell:
This appears to still be broken on beta3 @kris-watts-gravwell

kris-watts-gravwell:
This MIGHT be related to the Value vs RawValue in that structure.

test program that appears to show the behavior:

```
package main

import (
	"fmt"
	"log"
	"net"
	"time"

	"github.com/gravwell/gravwell/v3/client"
	"github.com/gravwell/gravwell/v3/client/objlog"
[...]

PR #1392, opened 2024-11-07
fix client to no wipe out enumerated values on ingest, covers issue #1278

kris-watts-gravwell:
We were intentionally removing EVs on IngestEntries, no idea why, these are used all over, they should be kept.

This PR addresses https://github.com/gravwell/gravwell/issues/1278


david-fritz-gravwell (approved):
LGTM

PR #1446, opened 2024-11-27
Tag rollover control

kris-watts-gravwell:
This PR adds some controls in negotiating tags at the ingest layer so that ingesters can fail faster when when trying to negotiate too many tags.

Before the remote side is tracking and will reject too many tags, but that can cause buffering and connection bumps.  This will make it so if a plugin or ingester goes haywire the ingest muxer can shut it down before it has to do the whole loop and have a remote federator/indexer reject the tag allocation.

Just a fail faster patch.


david-fritz-gravwell (approved):
LGTM

PR #1450, opened 2024-12-02
make the generator less crazy on very high throughput generator calls

kris-watts-gravwell:
before it was generating basically infinte orthogonality, this caps things to some reasonable constant.


david-fritz-gravwell (approved):
LGTM

PR #1454, opened 2024-12-02
making tag negotiation async and triggered by actual usage

kris-watts-gravwell:
This PR does a few things:

1. Dynamic Tag negotiation will only actually occur on an ingestConnect if an entry actually shows up
2. dynamic tag negotiation won't block and wait for each ingestConnection to do its thing

### DIFF
--- a/ingest/muxer.go
+++ b/ingest/muxer.go
@@ -112,6 +112,7 @@ type IngestMuxer struct {
 	tagTranslators       []*tagTrans
 	dests                []Target
 	errDest              []TargetError
+	tc                   tagMaskTracker
 	tags                 []string
 	tagMap               map[string]entry.EntryTag
 	pubKey               string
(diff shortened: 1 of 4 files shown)


ingest/muxer.go
```diff
@@ -1348,14 +1316,24 @@ inputLoop:
 
 			e := ee.(*entry.Entry)
 
-			ttag, ok = nc.tt.Translate(e.Tag)
-			if !ok {
+			if ttag, err = nc.translateTag(e.Tag); err != nil {
 				// If the ingest muxer has no idea what this tag is, drop it and notify
 				if name, ok := im.LookupTag(e.Tag); !ok {
```
david-fritz-gravwell, line 1321: The "does the muxer even know about this?" check should occur before we attempt to negotiate

david-fritz-gravwell (changes requested):

ingest/muxer.go
kris-watts-gravwell: the muxer lookup is a map lookup where the translate is an order 1 vector lookup, realistically, you would have to be using the muxer wrong to get here, so I don't think taking the performance hit is worth it here when you have to be going out of your way to even get here.

david-fritz-gravwell (approved):
LGTM

Issue #1455, opened 2024-12-03
Ingest Muxer: on close attempt to drain the entry and block channels


kris-watts-gravwell:
# Enhancement

## What is the enhancement to be made?

Some users are not calling Sync when shutting down the ingester, we should attempt to drain the block and entry channels on our way out, otherwise its possible to race and lose entries.

All our ingesters call sync before closing, so this isn't a huge deal, but users of the library may not.

## Why should we make this change? (Business justification? What problem is the feature trying to solve?)

When an ingester shuts down without calling sync or using a cache.

## Any other comments?

I think this is a high priority because it is easy to just call close after a successful write and no cache and assume everything will work.

david-fritz-gravwell:
NEVERMIND my search window was off by _one second_

david-fritz-gravwell:
LGTM

PR #1461, opened 2024-12-03
fixing up some of the shutdown logic

kris-watts-gravwell:
so that blocked writes fail immediately and we do a little better at cleanup with the cache on

This PR makes the following changes:
1. blocked writes fail immediately on muxer shutdowns
2. shutdowns when the cache is active will take some better care to try to get data into the cache
3. Close no longer sleeps, instead we rely on cache and cleanup logic

This PR addresses https://github.com/gravwell/gravwell/issues/1455


david-fritz-gravwell (approved):
LGTM

PR #1468, opened 2024-12-04
Fixup some more shutdown logic

kris-watts-gravwell:
This PR does the following:

1. separates dieChan into a write barrier and a context
2. cleans up how we recycle entries and optimizes when everything is down
3. adds a timeout sync method in the entryWriter so that we can actually time out flushes and syncs
4. Fixes the muxer so that an empty CacheDepth isn't treated as a synchronous channel when no cache is active


ingest/entryWriter.go
```diff
@@ -191,6 +191,40 @@ func (ew *EntryWriter) ForceAck() error {
...
+		if time.Since(now) > to {
+			ew.conn.ClearReadTimeout()
+			return ErrTimeout
+		}
+		if err := ew.serviceAcks(true); err != nil {
+			ew.conn.ClearReadTimeout()
+			return err
+		}
+	}
+	if ew.ecb.Count() > 0 {
```
david-fritz-gravwell, line 221: Unless there's a race, I don't see how this conditional is ever reached AND true.

david-fritz-gravwell (changes requested):

ingest/entryWriter.go
kris-watts-gravwell: Agreed but that code is ancient and I didn't want to change more stuff

david-fritz-gravwell (approved):
LGTM

PR #1472, opened 2024-12-06
sync before we close

kris-watts-gravwell:
its pretty easy to drop data if you don't Sync the ingest muxer before closing, all our ingesters always do a sync then close but other users may not.

If you DO do a proper sync then close then this call adds 10ms to the close call, but thats fine, we stripped out a 500ms delay earlier.


david-fritz-gravwell (approved):
LGTM

Issue #1442, opened 2024-11-25
ingest: add minimum version parameter to muxer config


john-floren-gravwell:
# Enhancement

## What is the enhancement to be made?

In muxer config, take a minimum version; if the server reports an older version, fail immediately.

## Why should we make this change? (Business justification? What problem is the feature trying to solve?)

Makes it easier to shut down if trying to ditto to an overly-old indexer.

kris-watts-gravwell:
## How I tested:

1. setup a `src` container based on v5.7.0-beta8
2. setup a `dst` container based on 5.6.10
3. setup a `dst2` container based on 5.6.7-beta8
4. ingested a bunch of stuff into the `src` container
5. configured ditto and pointed it at `dst` (5.6.10)
6. waited
7. reconfigured ditto to point at `dst2` (5.7.0)

### What Happened
1. `dst` (5.6.10) never got any data (yay)
2. `dst2) (5.7.0) got a bunch of data (yay)


### Test 2
1. Configure src to point at BOTH `dst` and `dst2`
2. Watch ingest



# VERDICT
LGTM 5.7.0-beta8

kris-watts-gravwell:
```
<12>1 2025-01-27T22:52:29.128252Z 93bba1f5a22e gravwell_indexer - ingest/muxer.go:1949 [gw@1 indexer="tcp://172.17.0.5:4023" ingester="gravwell-ditto" version="5.7.0" ingesteruuid="bd0ce529-6926-4960-983a-82526fbbe933" server-version="8" min-version="9"] indexer server version is less than specified minimum API level, refusing to connect
```

double LGTM

PR #1480, opened 2024-12-13
Add MinVersion field to muxer configs.

john-floren-gravwell:
Muxer will refuse to connect to any indexer reporting an API version lower than that specified in the config, thus if you intend to use ditto you might say:

		igCfg := ingest.UniformMuxerConfig{
			[...]
			MinVersion:         ingest.MINIMUM_DITTO_VERSION,
			[...]
		}
		igst, _ = ingest.NewUniformMuxer(igCfg)

An insufficient API version is treated as non-fatal; the muxer will attempt to reconnect to that indexer periodically, on the assumption that it will eventually be updated.

As long as one indexer meets the minimum version requirement, you will be able to use the muxer as usual. If not, the muxer will never go hot.

### DIFF
--- a/ingest/muxer.go
+++ b/ingest/muxer.go
@@ -157,6 +157,7 @@ type IngestMuxer struct {
 	start                time.Time    // when the muxer was started
 	attacher             *attach.Attacher
 	attachActive         bool
+	minVersion           uint16
 }
 
 type UniformMuxerConfig struct {
@@ -181,6 +182,7 @@ type UniformMuxerConfig struct {
 	RateLimitBps      int64
 	LogSourceOverride net.IP
 	Attach            attach.AttachConfig
+	MinVersion        uint16 // minimum API version of indexers
 }
 
 type MuxerConfig struct {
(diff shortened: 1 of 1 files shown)


john-floren-gravwell:
Addresses #1442

david-fritz-gravwell (approved):
LGTM

ingest/muxer.go
```diff
@@ -1831,6 +1836,24 @@ loop:
 			}
 			continue
 		}
+		// Make sure the version is new enough
+		if ig.ew.serverVersion < im.minVersion {
```
kris-watts-gravwell, line 1840: I think this needs an additional check.  

If minVersion > 0 && serverVersion < minVersion

kris-watts-gravwell (changes requested):
It looks like this will always fail if min version is not set.   I would argue we only do that check if minVersion is set

kris-watts-gravwell (approved):
LGTM

ingest/muxer.go
kris-watts-gravwell: yes I am a moron and yes I realized it several hours later.

PR #1488, opened 2024-12-18
muxer and ingest reporting resilience

kris-watts-gravwell:
This PR does teh following:

1. Removes the global muxer level ingester state reporting routine
2. Adds some functionality in the individual muxer routines to have individual ingestConnections report state
3. Reduces the frequency of ingester state updates when the only thing changing is ingest rates
4. Adds some timeouts to the ingestConnection close routines so that we can detect and close bad connections
5. Adds a new notification boolean on the ingestConnection routine that can signal to the system that grabs new connections that we think the upstream host is unhealthy and we should backoff and sleep before grabbing a new connection

### DIFF
--- a/ingest/muxer.go
+++ b/ingest/muxer.go
@@ -945,9 +945,8 @@ loop:
 			if err == nil {
 				good++
 			} else {
-				if lastErr == nil {
-					lastErr = err
-				}
+				//this will merge errors and even return nil if we need to
+				lastErr = mergeError(lastErr, err)
 				down++
 			}
 		case <-tmr.C:
@@ -964,7 +963,7 @@ loop:
 	} else if down == total {
 		return ErrAllConnsDown
 	} else if timeout {
-		// in this case and the lastError case,
+		// if its a timeout, throw the constant timeout error so that the caller can detect it and do the right thing
 		return ErrTimeout
 	}
 	return lastErr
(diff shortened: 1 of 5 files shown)


ingest/muxer.go
```diff
@@ -882,31 +889,85 @@ func (im *IngestMuxer) SyncContext(ctx context.Context, to time.Duration) error
...
+	if (good + down) == total {
+		//every single routine responded, its safe to close the channel
+		close(ch)
+	}
+	if good == total {
+		return nil //all good
+	} else if down == total {
 		return ErrAllConnsDown
+	} else if timeout {
+		// in this case and the lastError case,
```
john-floren-gravwell, line 967: it'd be nice to finish this comment

ingest/muxer.go
```diff
@@ -882,31 +889,85 @@ func (im *IngestMuxer) SyncContext(ctx context.Context, to time.Duration) error
...
+loop:
+	for count > 0 {
+		select {
+		case err := <-ch:
+			count--
+			if err == nil {
+				good++
+			} else {
+				if lastErr == nil {
+					lastErr = err
```
john-floren-gravwell, line 949: we only keep one error here. Might we merge them instead?

john-floren-gravwell (changes requested):

ashnwade:
Would like to get approval from both John and Fritz before merging

ingest/muxer.go
kris-watts-gravwell: resolved, fixed the bad comment

ingest/muxer.go
kris-watts-gravwell: added merge utility, please validate that i got the logic right.

david-fritz-gravwell (approved):
LGTM

PR #1491, opened 2024-12-19
add release of file lock for cache

kris-watts-gravwell:
This PR is more work on cleaning up abnormal shutdown and entry recovery in the case of errors.

It does the following:

1. When committing/closing the cache we also release the lock on the file (bug)
2. When grabbing outstanding entries on shutdown or failure we also purge the entryConfBuffer
3. Added a new ejectOutstandingEntries() call to ingestConnections and buffers and stuff


david-fritz-gravwell (approved):
LGTM

PR #1522, opened 2025-01-22
change the RemoteAI stuff to just LogbotAI 

kris-watts-gravwell:
reduce ambiguity on where the AI processors live

This PR addresses  no specific issue.

*NOTE* there was a govet issue in the `client/mfa.go` file, so I fixed that too.

### DIFF
--- a/client/types/cbac.go
+++ b/client/types/cbac.go
@@ -76,7 +76,7 @@ const (
 	SecretWrite     Capability = 48
 	AlertRead       Capability = 49
 	AlertWrite      Capability = 50
-	RemoteAIService Capability = 51
+	LogbotAIService Capability = 51
 	_maxCap         Capability = 52 //REMINDER - when adding capabilities, make sure to expand this number
 )
 
@@ -101,7 +101,7 @@ const (
 	UsersAndGroupsCat  = `Users and Groups`
 	SystemAndStatsCat  = `System and Stats`
 	SecretsCat         = `Secrets`
-	RemoteAIServiceCat = `Remote AI Service APIs`
+	LogbotAIServiceCat = `Logbot AI Service APIs`
 )
 
 const (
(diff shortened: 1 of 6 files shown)


ashnwade (commented):
We discussed possibly calling this `LogbotChat`.
I think that is a more user friendly name for it and frontend displays what the backend provides. I would consider renaming this and even updating the description 
from: "User can submit requests to the Logbot AI service APIs"
to: "User can submit requests to the Logbot Chat service"

client/types/cbac.go
```diff
@@ -101,7 +101,7 @@ const (
 	UsersAndGroupsCat  = `Users and Groups`
 	SystemAndStatsCat  = `System and Stats`
 	SecretsCat         = `Secrets`
-	RemoteAIServiceCat = `Remote AI Service APIs`
+	LogbotAIServiceCat = `Logbot AI Service APIs`
```
ashnwade, line 104: Can we just say "Logbot AI" ? I don't think it adds value to mention APIs here.

client/types/cbac.go
```diff
@@ -792,8 +792,8 @@ func (c Capability) Description() string {
 		return `User can read and access alerts`
 	case AlertWrite:
 		return `User can create, update, and delete alerts`
-	case RemoteAIService:
-		return `User can submit requests to the remote AI service APIs`
+	case LogbotAIService:
+		return `User can submit requests to the Logbot AI service APIs`
```
ashnwade, line 796: Do we need to say service APIs here?

client/types/cbac.go
kris-watts-gravwell: agreed, fixed

michael-wisely-gravwell (commented):
This looks good to me! I'd stamp it with an approve but I'd like @ashnwade 's opinion too

ashnwade (approved):
LGTM

PR #1529, opened 2025-02-04
SimpleRelay add stale data flush on RFC TCP streams

kris-watts-gravwell:
adding ability to force out data on an RFC5424 or RFC3164 stream that doesn't have a follow up header block

This PR addresses no specific issue, but tries to make SimpleRelay detect stale data on a TCP connection that doesn't have a follow up RFC header block.

Might be worth porting into the other RFC reader type and maybe regex reader types too.

I did some limited speed tests using the generator on my workstation and saw no difference whatsoever in syslog throughput.

### DIFF
--- a/.github/workflows/golang-testing.yml
+++ b/.github/workflows/golang-testing.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v4
       with:
-        go-version: 1.23.5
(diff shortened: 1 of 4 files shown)


david-fritz-gravwell (approved):
LGTM

ashnwade (approved):
Already reviewed and approved by Fritz. Approving just the go tidy

Issue #1416, opened 2024-11-11
timegrinder: Tweak defaults of where we look & how we handle zero dates


john-floren-gravwell:
Unless a specific timestamp format is set, timegrinder should probably only look so far into the data by default. Most of the time, the "most relevant" timestamp will be near the start, but timegrinder walks through a set of date formats in order; if one of the "earlier" formats is found way deep in the entry, we'll use that, even if there's a perfectly valid timestamp right at the start. It makes sense to have a setting (either by default or caller-specified) 

We should also, if we find a valid timestamp that happens to be precisely the Unix epoch, walk through all the rest of the available formats to see if we can get a non-zero date first.

kris-watts-gravwell:
I think the default should be any timestamps older than the year 1970 be ignored and set to NOW, this will catch zero timestamps AND anything chucking a zero or near zero EPOCH timestamp which we commonly see with broken data sources.  Maybe anything before Jan 2 1070?

Make timegrinder accept a config parameter to override this though.

john-floren-gravwell:
We've implemented the requisite functionality in timegrinder, now we just need to throw this into all the ingesters:

```
			var window timegrinder.TimestampWindow
			window, err = cfg.GlobalTimestampWindow()
			if err != nil {
				lg.Error("Failed to get global timestamp window", log.KVErr(err))
				return
			}
			if hcfg.tg, err = timegrinder.New(timegrinder.Config{TSWindow: window}); err != nil {
```

john-floren-gravwell:
The following ingesters require a little more work to get the changes implemented properly:

* filefollower (must push it down into filewatch library)
* kafka_consumer
* simple relay (instantiated in many many places, need to chase down each one)

kris-watts-gravwell:
HTTP ingester does NOT respect the timeoverrides on the HEC compatible ingester when using the proper HEC format with a timestamp.

kris-watts-gravwell:
LGTM 5.8.0-beta3

PR #1535, opened 2025-02-10
Add TimestampCutoff to timegrinder

john-floren-gravwell:
If you set TimestampCutoff in timegrinder.Config, any extracted timestamps prior to that date will be ignored -- this avoids extracting random big numbers as Unix timestamps from the 80s, for instance.

Part of gravwell/gravwell#1416


david-fritz-gravwell (dismissed):
LGTM

ingesters/HttpIngester/config.go
```diff
@@ -66,6 +67,7 @@ type lst struct {
 	Assume_Local_Timezone     bool
 	Timezone_Override         string
 	Timestamp_Format_Override string //override the timestamp format
+	Timestamp_Cutoff          string
```
kris-watts-gravwell, line 70: this isn't documented yet, I think this should be a duration NOT a timestamp.

The timestamp will just mean you have to continually update it or pick something that won't ever exactly fit.

I would rather say nothing older than -2 years or +2d to control outright bad timestamps making it in.

ingesters/HttpIngester/hec_config.go
```diff
@@ -39,6 +40,7 @@ type hecCompatible struct {
 	Tag_Name                  string   //the tag to assign to the request
 	Ignore_Timestamps         bool
 	Timestamp_Format_Override string //override the timestamp format (only used for raw)
+	Timestamp_Cutoff          string
```
kris-watts-gravwell, line 43: same, i say durations not timestamps.  and maybe it should be global, not per listener

timegrinder/processors.go
```diff
@@ -180,6 +180,7 @@ type Processor interface {
 	ToString(time.Time) string
 	ExtractionRegex() string
 	Name() string
+	SetCutoff(time.Time)
```
kris-watts-gravwell, line 183: i think we nee durations not time.Time everywhere, a floating "window around now" not some concrete timestamp.

kris-watts-gravwell (changes requested):
I vote for durations not time.Time  so that we can create a window.

Basically a global ingester config that lets me define a window around NOW where if bad timestamps land we force them to now.

ingesters/HttpIngester/config.go
```diff
@@ -37,13 +39,15 @@ const (
...
-	Max_Connections         int
-	Max_Concurrent_Requests int
+	Bind                       string
+	Max_Body                   int
+	TLS_Certificate_File       string
+	TLS_Key_File               string
+	Health_Check_URL           string
+	Max_Connections            int
+	Max_Concurrent_Requests    int
+	Timestamp_Max_Past_Delta   string
```
kris-watts-gravwell, line 49: I don't believe we should have one offs in these, I think it should be added to the `config.IngestConfig` so that we can use them in all ingesters more easily.

As is we are going to have to make changes to every single of our ingesters and every single listener.

We already have the timegrinder override stuff that timegrinder can pick up using our abstracted functions, I think this should go to that same place using that same methodology

ingesters/HttpIngester/config.go
```diff
@@ -132,6 +136,18 @@ func (c *cfgType) Verify() error {
 	if hc, ok := c.HealthCheck(); ok {
 		urls[newRoute(http.MethodGet, hc)] = `health check`
 	}
+
+	if len(c.Timestamp_Max_Past_Delta) > 0 {
```
kris-watts-gravwell, line 140: same

ingesters/HttpIngester/config.go
```diff
@@ -254,6 +270,20 @@ func (c *cfgType) MaxBody() int {
 	return c.Max_Body
 }
 
+func (c *cfgType) GlobalTimestampWindow() (w timegrinder.TimestampWindow, err error) {
```
kris-watts-gravwell, line 273: these needs to check for negative durations, people will get confused and put those in.

kris-watts-gravwell: also move this into the IngestConfig so all ingesters can use it

kris-watts-gravwell (approved):
LGTM

Issue #1536, opened 2025-02-10
generator: can panic generatior by providing single entry count and a duration


kris-watts-gravwell:
# Bug Report

## What are the steps to reproduce this issue?

1. Run the gravwellGenerator with the following args:
```
./gravwellGenerator -clear-conns=172.19.0.3,172.19.0.4 -type=syslog -start-time "2025-02-09T23:00:00Z" -duration=1h -entry-count=1 -tag-name=default
```

## What happens?

Panic

## What were you expecting to happen?

not panic

## Any logs, error output, etc?

```
panic: invalid argument to Intn

goroutine 1 [running]:
math/rand.(*Rand).Intn(0xc0001a18d0?, 0x444?)
	/opt/golang/go/src/math/rand/rand.go:180 +0x4c
math/rand.Intn(0x0)
	/opt/golang/go/src/math/rand/rand.go:453 +0x25
main.getHost(...)
	/home/kris/githubwork/gravwell/generators/gravwellGenerator/consts.go:302
main.genDataSyslog({0xc00007c120?, 0x0?, 0x0?})
	/home/kris/githubwork/gravwell/generators/gravwellGenerator/sysloggen.go:24 +0xae
github.com/gravwell/gravwell/v3/generators/base.OneShot({0x99e3b8, 0xc00028e508}, 0x0, {_, _, _}, {0x1, 0x0, 0x0, 0x0, ...}, ...)
	/home/kris/githubwork/gravwell/generators/base/base.go:366 +0x52c
main.main()
	/home/kris/githubwork/gravwell/generators/gravwellGenerator/main.go:81 +0x5ba

```

## Any other comments?

Lookos like we just have a naive call to `rand.Intn()`

## What versions of software are you using?

5.7.1

john-floren-gravwell:
lgtm on 5.7.3-beta1

PR #1551, opened 2025-02-18
fix consts count overrides so we don't panic when seeding, issue #1536

kris-watts-gravwell:
This PR addresses #1536


david-fritz-gravwell (approved):
LGTM

PR #1554, opened 2025-02-18
Add special cases to catch ShardInfo too far in the future.

john-floren-gravwell:
encoding/json won't handle years greater than 9999, so define a max timestamp and set the ShardInfo's Start/End fields to that in MarshalJSON if they're too big.

### DIFF
--- a/client/types/encoding_test.go
+++ b/client/types/encoding_test.go
@@ -1,0 +1,39 @@
+package types
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+)
+
+func TestJsonMaxDate(t *testing.T) {
+	si := ShardInfo{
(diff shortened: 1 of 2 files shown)


client/types/encoding_test.go
```diff
@@ -0,0 +1,39 @@
...
+		t.Fatal(err)
+	} else if x.Start != maxJsonTimestamp || x.End != maxJsonTimestamp {
+		t.Fatalf("Invalid timestamps on future shard: %v, %v", x.Start, x.End)
+	}
+
+	si = ShardInfo{
+		Start: time.Now(),
+		End:   time.Now().Add(1 * time.Hour),
+	}
+	b, err = json.Marshal(si)
```
kris-watts-gravwell, line 29: Lgtm

kris-watts-gravwell (approved):
Lgtm

PR #1588, opened 2025-03-06
always URL decode s3 keys

david-fritz-gravwell:


kris-watts-gravwell (approved):
looks incredibly benign, so LGTM, but I would like to hand/show this to the requesting party and see if they can validate too.

I don't have a data stream that does this to test.

kris-watts-gravwell (approved):
still LGTM

PR #1618, opened 2025-03-28
`next-major` -> `preview/major/search-store`

kris-watts-gravwell:
This PR addresses no specific issue.

I also bumped some dependencies and bumped the go runtime to 1.24.1 so that we can get some tests going on a non-release ready branch with that new run time in case something explodes.


ashnwade (approved):
LGTM

PR #1620, opened 2025-04-02
bump some versions and replace JWT package due to vulns in old one

kris-watts-gravwell:
just replacing/upgrading a dependency

This PR addresses no specific issue


ashnwade (approved):
LGTM

PR #1628, opened 2025-04-10
make HEC health API call WillBlock instead of trying to figure out health by hand

kris-watts-gravwell:
This PR addresses no specific issue, but the HEC health API was not calling WillBlock (which checks cache levels and all that) and was instead just returning based on whether we had upstream ingest connections.


david-fritz-gravwell (approved):
LGTM

PR #1631, opened 2025-04-14
a little abstraction on Debug handlers so we can use them in other components

kris-watts-gravwell:
This PR addresses no specific issue


ashnwade (approved):
LGTM

PR #1679, opened 2025-05-20
fix some dumb inverted logic and do the syscallN dance correctly

kris-watts-gravwell:
Fix bad syscall conversions from the deprecated API to the new syscallN interface


david-fritz-gravwell (approved):
LGTM

PR #1682, opened 2025-05-22
fix kafka consumer default tag handling with multiple consumers and no transport tags defined in kafka messages

kris-watts-gravwell:


ingesters/kafka_consumer/consumer.go
```diff
@@ -392,14 +393,18 @@ func (kc *kafkaConsumer) resolveSourceAndTag(m *sarama.ConsumerMessage) (tag ent
 			if tag, tagHit, err = kc.resolveTag(string(rh.Value)); err != nil {
 				return
 			}
+			fmt.Printf("%s -  %q %q %q %v %v %v %v\n", kc.name,
```
david-fritz-gravwell, line 396: is this a debug statement?

kris-watts-gravwell: good catch.  PTAL

david-fritz-gravwell (approved):
LGTM

PR #1694, opened 2025-05-28
add Message type and example IDs

david-fritz-gravwell:
@evan-gravwell @michael-wisely-gravwell 

Started to implement the new Message type and I have new questions. I've left an example implementation below to illustrate.

My main concern here is that the ID, while being unique, also carries an implied unique severity. This specification though allows mismatched severity/ID pairs, which feels odd.

Thoughts?

### DIFF
--- a/client/types/message.go
+++ b/client/types/message.go
@@ -1,0 +1,32 @@
+/*************************************************************************
+ * Copyright 2025 Gravwell, Inc. All rights reserved.
+ * Contact: <legal@gravwell.io>
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD 2-clause license. See the LICENSE file for details.
+ **************************************************************************/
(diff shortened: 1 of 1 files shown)


david-fritz-gravwell:
```
ID = LiterallyCaughtFire
Severity = "info" // no need to call the fire department, this is fine
```

evan-gravwell:
Would it ever make sense to use the same message ID with different level of severity? Maybe something like launching a search with an end date after the start date would be `fatal` while zooming on an existing, healthy search with an end date after the start date would be an error, as the search wouldn't be terminated.

@david-fritz-gravwell

david-fritz-gravwell:
Well that's a good point. Nevermind!

evan-gravwell (approved):
LGTM!

PR #1710, opened 2025-06-09
FEAT(#1138): gwcli: implement mfa login

rory-landau-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/1138.

This PR is focused on implementing MFA handling, allowing users to input TOTP and recovery codes while logging in (if applicable). The MFA view is barebones and probably could do with a makeover, but it works:

mfa

The TOTP and recovery prompts will only appear after a user has supplied a valid username & password, if the backend responds that MFA is required.

API Tokens are now accepted as well, allowing script usage to circumvent having to incorporate TOTP generators.

> [!IMPORTANT]
> When MFA is enabled on an account, `--api` is the only acceptable way to log in when `--script` is supplied.

A significant reworking of the login system is included in order to accommodate these enhancements, fix some pre-existing bugs, and improve predictability and maintainability of the login system.

This PR also introduces the basics of TUI-interactive testing, primarily via the teatest experimental package. Automated testing now exists for the mini prompts used to collect credentials and MFA information. Future PRs will begin to incorporate more interactive (direct TUI-usage) tests for better code coverage (rather than just relying on integration-style testing).

> [!WARNING]
> This PR is also the discovery point of a behavior change made in Bubbles that breaks our input validation across the board: https://github.com/charmbracelet/bubbles/pull/185. In a nutshell, the maintainers accepted a PR that alters `textinput.Validate` behavior to **not** undo the last character input if an error was returned, reducing the usefulness of validation on rune-restricted inputs (such as our cron validator). An issue (https://github.com/charmbracelet/bubbles/issues/691) has been opened, but no acknowledgment has been made. This behavior change could be mitigated by manually validating inputs, but this is a silly, external workaround for a behavior that used to be internal.


gwcli/connection/connection.go
```diff
@@ -77,76 +138,167 @@ func Initialize(conn string, UseHttps, InsecureNoEnforceCerts bool, restLogPath
...
+		if err := loginWithCredentials(username, password, scriptMode); err != nil {
+			return err
+		}
+	} else { // a username was given, but no password/passfile
+		// in script mode, fail out
+		if scriptMode {
+			return ErrCredentialsOrAPITokenRequired
+		}
+		// in interactive mode, throw up a prompt and pre-populate username
+		mfa, err := promptForInput(username)
```
john-floren-gravwell, line 174: I think this function should be renamed; it's not prompting for input, it's getting username AND password and logging in if possible.

rory-landau-gravwell: promptForInput calls the .Collect() subroutines in `credprompt` and `mfaprompt`, both of which are self-contained/self-destructive TUIs which prompt for input. We can certainly rename `promptForInput()`, but the name was intended to alert developers that it will wait on stdin

john-floren-gravwell: When I read this code at first, I thought "ok, we're going to prompt for the username, pre-populating it if set. Now where does it prompt for the password, and MFA if needed?", due to the name. Inspection of the function itself showed what it does, and being internal it's not a huge deal, but it definitely confused me the first time around.

rory-landau-gravwell: Gotcha. Well in that case, let me come up with clearer name

PR #1727, opened 2025-06-12
Updated state tracker to use raw json message as key

mike-wade-gravwell:
- Changed statetracker key to use json instead of string to make the state tracker more flexible 
- Fixed an issue with thinkst that was causing the state tracker to not update the state file resulting in multiple copies of logs


experiments/gravwell_fetcher/shodan.go
```diff
@@ -140,13 +144,21 @@ func buildShodanHandlerConfig(cfg *cfgType, src net.IP, ot *objectTracker, lg *l
 		// check if there is a statetracker object for each config
 		_, ok := ot.Get("shodan", k)
 		if !ok {
+			query := shodanQueryResponse{
+				Query: v.Query,
+			}
+			keyBytes, err := json.Marshal(query)
+			if err != nil {
+				lg.Error("Failed to marshal last entry key", log.KVErr(err))
+				// Handle error appropriately
```
david-fritz-gravwell, line 153: what is the error handling supposed to be here?

experiments/gravwell_fetcher/shodan.go
```diff
@@ -304,11 +338,20 @@ func getShodanHostData(h *shodanHandlerConfig, latestTS string, src net.IP, rl *
 		lg.Error("failed to send entry", log.KVErr(err))
 	}
 
+	query := shodanQueryResponse{
+		Query: latestTS,
+	}
+	keyBytes, err := json.Marshal(query)
+	if err != nil {
+		lg.Error("Failed to marshal last entry key", log.KVErr(err))
+		// Handle error appropriately
```
david-fritz-gravwell, line 347: what is the error handling supposed to be here?

experiments/gravwell_fetcher/shodan.go
```diff
@@ -374,11 +417,20 @@ func getShodanSearchData(h *shodanHandlerConfig, latestTS string, src net.IP, rl
 	}
 	lg.Info(fmt.Sprintf("ingested %v entries in tag %v", len(searchResp.Matches), h.tag))
 
+	query := shodanQueryResponse{
+		Query: latestTS,
+	}
+	keyBytes, err := json.Marshal(query)
+	if err != nil {
+		lg.Error("Failed to marshal last entry key", log.KVErr(err))
+		// Handle error appropriately
```
david-fritz-gravwell, line 426: see above

experiments/gravwell_fetcher/shodan.go
```diff
@@ -449,11 +501,20 @@ func getShodanCountData(h *shodanHandlerConfig, latestTS string, src net.IP, rl
...
 	}
 
+	query := shodanQueryResponse{
+		Query: latestTS,
+	}
+
+	keyBytes, err := json.Marshal(query)
+	if err != nil {
+		lg.Error("Failed to marshal last entry key", log.KVErr(err))
+		// Handle error appropriately
```
david-fritz-gravwell, line 511: see above

experiments/gravwell_fetcher/thinkst.go
```diff
@@ -336,12 +356,21 @@ func getThinkstIncidentLogs(cli *http.Client, latestTS string, src net.IP, rl *r
 					lg.Error("failed to send entry", log.KVErr(err))
 				}
 			}
+			newKey := thinkstLastEntryKey{
+				LastEntryKey: lastEntryKey,
+			}
+			keyBytes, err := json.Marshal(newKey)
+			if err != nil {
+				lg.Error("Failed to marshal last entry key", log.KVErr(err))
+				// Handle error appropriately
```
david-fritz-gravwell, line 365: see above

experiments/gravwell_fetcher/thinkst.go
```diff
@@ -306,7 +325,8 @@ func getThinkstIncidentLogs(cli *http.Client, latestTS string, src net.IP, rl *r
 					lg.Info("could not find ts")
 				} else {
 					entryTime = r
-					if strconv.Itoa(t.UpdatedID) > lastEntryKey {
+					i, _ := strconv.Atoi(lastEntryKey)
```
david-fritz-gravwell, line 328: we should check the error here -- the returned integer will just be the zero value otherwise. If that is intended, please add a comment to that effect.

david-fritz-gravwell (changes requested):

experiments/gravwell_fetcher/shodan.go
mike-wade-gravwell: If the data coming back from query, which is sourced from the config file, can't be formed into JSON it would error before it tries to submit it to the object tracker.

experiments/gravwell_fetcher/thinkst.go
mike-wade-gravwell: I will explicitly set that.

experiments/gravwell_fetcher/shodan.go
mike-wade-gravwell: I completely removed the need for it.

experiments/gravwell_fetcher/thinkst.go
```diff
@@ -306,7 +325,11 @@ func getThinkstIncidentLogs(cli *http.Client, latestTS string, src net.IP, rl *r
 					lg.Info("could not find ts")
 				} else {
 					entryTime = r
-					if strconv.Itoa(t.UpdatedID) > lastEntryKey {
+					i, err := strconv.Atoi(lastEntryKey)
+					if err != nil {
+						i = 0
```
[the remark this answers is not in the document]
mike-wade-gravwell: Setting i=0 sets the floor for the updateID to compare against. Each thinkst incident comes with an updatedID int that shows the version of the event. In this case if i get any data back from the api it would take the latest updateID as the new key to be stored in the state tracker. 

This check is only meant to log if there was an issue with the state file

PR #1732, opened 2025-06-17
FEAT(#1138): Implement direct file ingestion

rory-landau-gravwell:
This PR addresses #1138.

This PR sees the implementation of file ingestion, allowing users to upload files from their local computer. Script mode supports an arbitrary number of files and is intelligent enough to match each file against a tag at the same index or apply the same tag to all files (if only a single tag was given). Interactive mode can only consume a single file at a time, but comes with a file browser to enable users to traverse their local drives interactively.

image

image

This PR also includes a structural rework of how gwcli handles colors/themes. The new format should be much clearer in how/where colors are being used and supports defining your own color schemes. The latter currently requires altering the source code, but this rework opens up the possibility of reading theme information from a configuration file in the future. It will also make implementing [NOCOLOR](https://no-color.org/) (via --no-color) much simpler.

>[!NOTE]
> A lot of files were touched in this PR, but the vast majority are non-functional changes stemming from the restructuring of the stylesheet package (forcing other packages to not rely on hardcoded colors/styles). A couple other bug fixes are included, too.


rory-landau-gravwell:
I am taking this PR back over as a draft in order to rework ingest more heavily to better support directories and Gravwell JSON files. Should make the action both more user-friendly and more maintainable.

rory-landau-gravwell:
The ingest mechanisms have been completely rewritten, with special focus on autoingest(). It should be much easier to use and more robust. This version also includes more granular testing.

kris-watts-gravwell:
can we move the TODO.MD file into an epic rather than an embedded file?

It is far easier to track that progress and link commits back to them via an EPIC and issues rather than a TODO file in the source.

john-floren-gravwell (changes requested):
When ingesting a file, we *must* sample the file and see if it contains JSON-formatted types.StringTagEntry objects. If so, *do not* require the user to enter a tag. We have to include this functionality to bring us on par with the GUI.

john-floren-gravwell:
> > When ingesting a file, we _must_ sample the file and see if it contains JSON-formatted types.StringTagEntry objects. If so, _do not_ require the user to enter a tag. We have to include this functionality to bring us on par with the GUI.
> 
> This must be in reference to interactive use, right? That is the current behavior in non-interactive (via `autoingest()`), but it looks like it is not in interactive mode.

Yep, we need to have it in interactive mode too. Can you just invoke the same code to sample the file and see if we're good to go?

rory-landau-gravwell:
> > > When ingesting a file, we _must_ sample the file and see if it contains JSON-formatted types.StringTagEntry objects. If so, _do not_ require the user to enter a tag. We have to include this functionality to bring us on par with the GUI.
> > 
> > 
> > This must be in reference to interactive use, right? That is the current behavior in non-interactive (via `autoingest()`), but it looks like it is not in interactive mode.
> 
> Yep, we need to have it in interactive mode too. Can you just invoke the same code to sample the file and see if we're good to go?

Aye, I see where `actor.go` is checking for an empty tag prior to sampling. Replacing that with determineTag() now so it benefits from the same handling as non-interactive.

rory-landau-gravwell:
I've reworked the View portions of ingest to correct a bug related to incorrect heights. Basically, some file names would wrap, causing the height counter to overflow its limit and exceed the bounds of the terminal.

This changes the horizontal alignment of the file picker. We cannot properly center align the filepicker until #1733 is completed (at least, not without causing graphical issues in certain edge cases). This is because the basic filepicker bubble treats the whole view as a single block, disallowing us from properly building ratios into each subsection. Currently, center aligning will align each line individually (causing jagged centering).

When #1733 is completed, we can compose the left (permissions+sizes), a divider (1, empty cell), and right (filenames) as 1 cell margin : left : 1 cell divider : right : 1 cell margin. Left and right should each be (TotalWidth-3)/2.

This will work perfectly well until then; it is fully functional, but less visually appealing.

john-floren-gravwell (approved):

PR #1738, opened 2025-07-02
add Messages fields to updated types

david-fritz-gravwell:
Per API changes for next-major, add `Messages` to several types. We currently only have a single warning we send from backend, so populate that as well.


kris-watts-gravwell (approved):
LGTM

PR #1743, opened 2025-07-09
FEAT(#1138): system metrics and scaffoldlist upgrades

rory-landau-gravwell:
This PR addresses #1138.

image

image

image

image

image

This PR is focused on implementing a series of actions for querying instance state. The new Systems nav is composed of 2 subnavs and 7 actions:
- `hardware`: a user-friendly overview of the instance's machinery (cpu/mem usage, storage amounts, disk and network IO, and hardware specifics).
    - Hardware does not implement list semantics (json, csv, table, columns, etc) because this data is already encapsulated (with list semantics) in the `indexer get` action.
    - This action follows the GUI's lead and ignores 'webserver'. Webserver's information could be included after deduplication, if desirable. We would just need to match on hardware hash or UUID to avoid duplicate statistics.
-  `storage`: displays hot/cold data ingestion info by disk. Implements list semantics.
- `ingesters` nav
    - `list`: displays basic information about all ingesters
    - `get`: fetches all information relevant to the matching ingesters. Prefixes matches on hostname, UUID, and name.
- `indexers` nav
    - `list`: display basic information about all indexers
    - `get`: display all information about the named indexer. Exact matches exclusively on name. It can return *a lot* of info.
    - `calendar`: return day-by-day ingestion amounts for a named indexer.

As part of implementing these actions, scaffoldlist received substantial upgrades, improving its functionality, flexibility, and robustness. scaffoldlist now accepts an options struct, which allows developers to modify its behavior. It can be fed a Pretty function (to pretty-print the data instead of returning tabular), have additional flags bolted on, and support arbitrary flag/arg validation external to its main dataFunc. scaffoldlist now also validates its parameters and can detect bad column names (rather than just returning no data for that column). A variety of other bugs and minor annoyances have been phased out.

### DIFF
--- a/gwcli/tree/systems/indexers/get.go
+++ b/gwcli/tree/systems/indexers/get.go
@@ -13,7 +13,6 @@ import (
 	"github.com/gravwell/gravwell/v4/gwcli/connection"
 	"github.com/gravwell/gravwell/v4/gwcli/stylesheet"
 	"github.com/gravwell/gravwell/v4/gwcli/utilities/scaffold/scaffoldlist"
-	"github.com/gravwell/gravwell/v4/utils/weave"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
@@ -34,26 +33,6 @@ func get() action.Pair {
 		example = fmt.Sprintf("%v 127.0.0.1:9404", use)
 	)
 
-	// we want to get *most* columns by default so it is easier to grab all of them and remove ones we do not want
-	var defaultColumns []string
-	if dcols, err := weave.StructFields(deepIndexerInfo{}, true); err != nil {
-		panic("failed to delve dii: " + err.Error())
-	} else {
-		// add columns not in exclude to the set of default columns
-		var defaultExclude = map[string]bool{
-			"Ingest.EntriesHourTail":   true,
-			"Ingest.EntriesMinuteTail": true,
-			"Ingest.BytesHourTail":     true,
-			"Ingest.BytesMinuteTail":   true,
-		} // hashset
-		for i := range dcols {
-			if _, found := defaultExclude[dcols[i]]; !found {
-				defaultColumns = append(defaultColumns, dcols[i])
-			}
-		}
-
-	}
-
 	return scaffoldlist.NewListAction(short, long, deepIndexerInfo{},
 		func(fs *pflag.FlagSet) ([]deepIndexerInfo, error) {
 			dii := deepIndexerInfo{Name: strings.TrimSpace(fs.Arg(0))}
@@ -66,8 +45,13 @@ func get() action.Pair {
 			return []deepIndexerInfo{dii}, nil
 		},
 		scaffoldlist.Options{
-			Use:            use,
-			DefaultColumns: defaultColumns,
+			Use: use,
+			ExcludeColumnsFromDefault: []string{
+				"Ingest.EntriesHourTail",
+				"Ingest.EntriesMinuteTail",
+				"Ingest.BytesHourTail",
+				"Ingest.BytesMinuteTail",
+			},
 			CmdMods: func(c *cobra.Command) {
 				c.Example = example
 			},
@@ -82,6 +66,14 @@ func get() action.Pair {
 					return "did not find indexer '" + fs.Arg(0) + "'", nil
 				}
 				return "", nil
+			},
+			ColumnAliases: map[string]string{
+				"Storage.DataIngestedHot":  "Storage.Hot.Ingested",
+				"Storage.DataIngestedCold": "Storage.Cold.Ingested",
+				"Storage.DataStoredHot":    "Storage.Hot.Stored",
+				"Storage.DataStoredCold":   "Storage.Cold.Stored",
+				"Storage.EntryCountHot":    "Storage.Hot.Count",
+				"Storage.EntryCountCold":   "Storage.Cold.Count",
 			},
 		})
 }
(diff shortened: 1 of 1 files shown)


rory-landau-gravwell:
This PR includes *some* changes from #1732, so I will leave it in draft until that is merged.

john-floren-gravwell (changes requested):
It's very challenging to tell, in the `systems hardware` command, where one indexer ends and another begins:

image

Either a newline, some sort of horizontal divider, a change in color, or a Unicode icon of some point would help make it more obvious.

We might also consider moving the summary from the top of the output to the bottom, so you don't have to scroll way back up if you have lots of indexers.

Also, and this might be challenging to solve due to Bubbletea, but if I'm using xterm and I scroll back up, every time the cursor blinks it automatically scrolls me back to the bottom again. (works fine in rxvt)

Would it be possible to wire up tab-completion for indexer names? I know this would involve a round trip but indexer "names" are often just an IP:port pair, which is a hassle to type.

`indexers calendar <indexer>`  shows nothing, and `indexers calendar` just shows me 0 entries/0 bytes. Intended, or an oversight?

In `storage` and `indexers list`, the column headers get truncated to uselessness:

image

I know one of the `Stats.DataSto` columns corresponds to hot and one to cold, but I can't tell which. Do we have a way to either override the width or somehow trim down/override the column header string so it'll fit?

For the `ingesters get` command, I suggest that we might trim down the default column set a bit... MetadataJSON and Configuration, for instance, could probably both be omitted.

rory-landau-gravwell:
> Either a newline, some sort of horizontal divider, a change in color, or a Unicode icon of some point would help make it more obvious.

Ezpz. I'll play with a couple ideas, see what works best.

> We might also consider moving the summary from the top of the output to the bottom, so you don't have to scroll way back up if you have lots of indexers.

Ezpz.
 
> Also, and this might be challenging to solve due to Bubbletea, but if I'm using xterm and I scroll back up, every time the cursor blinks it automatically scrolls me back to the bottom again. (works fine in rxvt)

That's a new one to me; have not experienced it in Terminator. Sounds really annoying, but that it occurs in xterm and not rxvt/terminator concerns me for the feasibility of a fix. I'll poke around.

> Would it be possible to wire up tab-completion for indexer names? I know this would involve a round trip but indexer "names" are often just an IP:port pair, which is a hassle to type.

Doable? Absolutely. It will be a heavier lift, though.

For completions in your terminal, Cobra supports suggestions (via Carapace, iirc), but gwcli doesn't *really* take advantage of that yet. This also relies on your shell actually sourcing the completions.

For completions at the gwcli prompt.... It would be rad to have and I'd love to support it, but there is currently no mechanism for actions to be able to register completions with Mother. I can develop this mechanism, but it will be a larger scale change as we'd need to add the receiver mechanism to Mother's build/boot process and then figure out how to bolt these context-aware completions to the current set of completions (which is only mildly context aware). It would be cool as hell though.

> `indexers calendar <indexer>` shows nothing, and `indexers calendar` just shows me 0 entries/0 bytes. Intended, or an oversight?

Calling calendar without `--start`/`--end` should set start to the beginning of today and end to the end of today for your local tz. However, the API returns data by day, so I'm not 100% sure it actually reads timestamp and therefore could (theoretically) be returning from now to now. Aka, I'm not actually clear if this is intended or a bug on my implementation.

> In `storage` and `indexers list`, the column headers get truncated to uselessness:
> I know one of the `Stats.DataSto` columns corresponds to hot and one to cold, but I can't tell which. Do we have a way to either override the width or somehow trim down/override the column header string so it'll fit?

Yes and no. This is actually a bigger issue for all table outputs. Weave does not support overriding widths or column renaming, but both are on the EPIC. Column truncation occurs because, at least when we started using lipgloss's table module (which was still considered bleeding edge), long column names broke the table, particularly when it wrapped.

> For the `ingesters get` command, I suggest that we might trim down the default column set a bit... MetadataJSON and Configuration, for instance, could probably both be omitted.

Ezpz.

rory-landau-gravwell:
full example of new divider:

image

rory-landau-gravwell:
The last batch of commits addresses some of the prior issues and adds new functionality:

1. Cleaned up the `systems hardware` cmd, moving summary to the foot and adding more explicit division between each section/indexer.
2. Trimmed default columns across most list actions.
3. Scaffoldlist can now build the default columns via whitelisting or blacklisting.
4. Weave (and scaffoldlist) now supports column aliasing. Column aliases should match across all column input (--show-columns information, --all-columns, --columns=<>, and even --json and --csv headers). This is the current fix for column header truncation (see Note 1 below).
5. `indexers calendar` actually works now and matches the data available in the GUI.
6. Some other minor bug fixes for other list actions.

There are still some issues floating in the ether:

1. Tables scale on text and do not extend or compress based on the terminal width, causing them to frequently wrap. This is an artefact of how we pass information from Mother to child action and should be addressed with a larger scale change (having Mother pass terminal size information on child invocation).
2. Completions are in a similar boat; they are better addressed as a more structural change. They have been put off for the time being.
3. Tangentially related: `queries past` (formerly `queries history`) does not return data without a debugger attached. Presumably this is a race condition somewhere, but it seems that the API itself is not returning the data (returns a 400: No Results). Unrelated to the current fixes, but I uncovered it while upgrading a swatch of actions utilizing scaffold list. I will throw it on the EPIC for further review.

Note 1: It is not possible to wrap column headers in lipgloss tables. This is because the table renderer has an explicit `.SetHeight(1)`, so any and all text after a newline (wrapped or manual) will be dropped. This may be circumvent-able by decapitating the table s.t. the headers are actual a separate table's rows (as rows can wrap, but headers cannot). Problems for future Rory.
Note 2: xterm still scrolls down on blink/gwcli output. This is an xterm setting and beyond scope of gwcli. A note has been added to the README.

PR #1753, opened 2025-07-21
FEAT(#1138): mac support

rory-landau-gravwell:
This PR addresses #1138.

The primary intention is to better support Mac users by switching away from alt+enter keybinds. These keybinds were primarily used in query, scaffoldedit, and scaffoldcreate. There are now submit buttons instead. Query can also be submitted via the EOF signal (ctrl+D).

image

image

Upgrading scaffoldedit and scaffoldcreate to support a submit button was a larger change than expected, so both received a handful of upgrades and tests in the process.

As is tradition, a number of bugs and background upgrades are included.
- BasicActions got another pass, for greater flexibility.
- DataScope received a number of style fixes; apparently it was broken when the stylesheet upgrade occurred. Tests have been installed to (hopefully) catch this in the future. These tests rely on TeaTest and therefore use Golden files. Golden file support has been shoehorned into the magefile.


david-fritz-gravwell (approved):
Holy cow

LGTM

PR #1769, opened 2025-08-01
FEAT(#1138): input parser rework

rory-landau-gravwell:
This PR addresses #1138.

This PR rewrites the walk() subroutine, which is used to parse user input. The changes began while fixing `-h` not working on navs but evolved into a full rewrite of the parser. It now has fewer external dependencies, more deterministic output, full test coverage, and more flexible results.

The PR also includes a couple new actions (`alerts list` and `resource delete`), fixes some small bugs, and actually implements admin mode.

`resource delete` is straightforward and works out of the box.

`alerts list` works in its bare form, but also supports --consumer and --dispatcher (which call `GetAlertsByConsumer` and `GetAlertsByDispatcher`, respectively). However, these two flags do not turn back data. Unfortunately, I am out of time to hunt down where exactly the issues lies.

`user admin` has a bare form and a toggle. Calling `user admin` bare displays whether the user is an administrator and if they have admin mode enabled. Calling `user admin -t` toggles admin mode.
Because gwcli does not have persistence (other than the JWT), admin mode only applies to the current session. This also means that admin mode makes no sense outside of a session context (aka outside of when Mother is running); thus, it is hidden until the user enters an interactive prompt.

Please note: admin mode has had basic testing, but needs automated tests and a more thorough investigation.
Relatedly, I am not sure that it currently serves much utility. It appends ?admin=true to API calls (which some calls, like `GetAlerts()`), but many API calls already have GetAll* equivalents which pull back all data iff you are an administrator. In the long term, we should seek to consolidate down to a single method of exercising admin features. Either remove admin mode and rely solely on `--all` for calls that have admin counterparts or hide `--all` and imply it when admin mode is enabled.


rory-landau-gravwell:
Thank you!

john-floren-gravwell (approved):

PR #1794, opened 2025-08-11
Registry macros

john-floren-gravwell:
Implements the new macro types in the registry. Also brings in the new users and groups.


david-fritz-gravwell (approved):
LGTM

PR #1812, opened 2025-08-28
Use the new User and Group types everywhere

john-floren-gravwell:
All APIs should now return the new types, not the old types.


david-fritz-gravwell (approved):
LGTM

PR #1813, opened 2025-09-02
Add SearchReference to PackedScheduledSearch type

john-floren-gravwell:
Addresses gravwell/issues#1559 by allowing kits to include scheduled searches which reference the search library.


david-fritz-gravwell (approved):
LGTM

PR #1839, opened 2025-09-10
Add StrictDashboard type

john-floren-gravwell:
This defines the full dashboard structure rather than punting it into a json blob.


david-fritz-gravwell (approved):
LGTM

PR #1861, opened 2025-09-25
Make a list request on macro endpoints return a MacroListResponse.

john-floren-gravwell:
We're going to do this for all assets, because it makes pagination possible.


kris-watts-gravwell (approved):
Lgtm

PR #1881, opened 2025-10-16
Search library -> saved queries for registry

john-floren-gravwell:
"Search library entries" was really clunky. They're "saved queries" now.

We still talk about SearchLibraries in kits, at least for now, for compatibility reasons, same reason they're still called Pivots in kits.


david-fritz-gravwell (approved):
LGTM

Issue #1900, opened 2025-10-29
CI: Improve local build checks cross platform compatibility


nic-osullivan-gravwell:
# Task

Before opening a task, please complete the following checklist:
- [x] Check to see if this is a duplicate

## What needs to be done?
`run_local_build_checks.sh` makes a fair number of assumptions about the environment we need to make these a bit more explicit. This is mainly adding `GOOS` and `GOARCH ` as needed, and explicitly starting with `CGO_ENABLED=0` and only enabling cgo when needed.

## Why is this task needed? 
DevEx and consistency in testing/checks

PR #1901, opened 2025-10-29
ci (#1900): be more specific about build platforms

nic-osullivan-gravwell:
This PR addresses #1900


ingesters/utils/caps/caps.go
```diff
@@ -0,0 +1,346 @@
...
+const (
+	All    Capabilities = 0xffffffffffffffff
+	minCap              = CHOWN
+	maxCap              = CHECKPOINT_RESTORE
+)
+
+// stolen directly from: https://pkg.go.dev/kernel.org/pub/linux/libs/security/libcap/cap#Value
+// we are choosing to treat this a BSD 3-clause as laid out in the license note:
+// https://pkg.go.dev/kernel.org/pub/linux/libs/security/libcap/cap#section-readme
+const (
```
nic-osullivan-gravwell: This is likely the largest change, need these defined on all platforms so things at least build. This doesn't mean they'll actually work, but at least build.

PR #1932, opened 2025-11-07
add regex replace preprocessor

### LINKED ISSUE #1735: Preprocessor: Regex replace
# Feature

## What is the feature to be added?

A preprocessor that takes a regex and a replacement string.  The preprocessor would replace any matches from the regex with the string.

Some data sources coming in have huge chunks of data that are not usable and can fill up the system. We ran into an issue with a use case where some endpoint data included the entire file contents on an event which could extend the length of the entry to over 500KB or accidental passwords exposed in commands.

Something like this could allow us to cut out those sections and save the platform from having to process them and spare the user the storage and cleanup when the log generator could not clean the data.

1. This would be helpful for customers that don't have a system to modify their data before it hits gravwell.  This is useful in cases where the event has a field or set of fields that are really big that they don't want logged. This would have been helpful in a use case where someone needed to modify their script logs.

2. Redaction. Some logs need to be redacted. It would be a nice feature to be able to look for known patterns and redact them.

### Question

1. Should it match and replace all instances or just the first?
2. How are subsequent matches handled if there is overlap with a replace?
e.g. Should i call the regex, replace the first instance, call it again, replace if it hits again, etc...

For example if your Regex is `foo-[a-z]+-bar` and your replace is `stuff` and your data is `foo-and-bar-foo-fooand-bar` the subsequent find and replaces means that the underlying data changes and so later calls will have to be re-aligned.

We could park it in a loop and just keep looping while the regex continues to match, but then a regex of `foo-[a-z]+-bar` and a replace of `foo-heyoooo-bar` will spin forever.

### Answer

1. I think it should be an option on the preprocessor.
2. Each preprocessor should get one shot in order. The result of the first should affect the second but it should not allow it to rerun.

david-fritz-gravwell:
Fixes #1735

### DIFF
--- a/ingest/processors/processors.go
+++ b/ingest/processors/processors.go
@@ -77,6 +77,7 @@ func CheckProcessor(id string) error {
 	case CorelightProcessor:
 	case SyslogRouterProcessor:
 	case TagSrcRouterProcessor:
+	case RegexReplaceProcessor:
 	default:
 		return checkProcessorOS(id)
 	}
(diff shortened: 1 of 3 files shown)


kris-watts-gravwell (approved):
LGTM

PR #1962, opened 2025-11-25
Registry resources

john-floren-gravwell:
Update the API for resources in the registry


kris-watts-gravwell (changes requested):
I think the monorepo import is wrong, builds are failing

kris-watts-gravwell (changes requested):
builds are still failing

PR #1920, opened 2025-11-05
feat(#1209): Quarantine cache files that cannot be opened

daniel-moreno-levy-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/1209

PR #1964, opened 2025-11-27
Chancacher was breaking internal tests.

john-floren-gravwell:
You need to stat on the directory, not the files. This PR reverts the merge-checking & locking behavior to how it was prior to #1920


kris-watts-gravwell (approved):
LGTM

Issue #1967, opened 2025-11-27
chancacher: harden tests for `chancacher` merge logic


daniel-moreno-levy-gravwell:
# Task

Before opening a task, please complete the following checklist:
- [ ] Check to see if this is a duplicate

## What needs to be done?

Investigate why https://github.com/gravwell/gravwell/pull/1920 exploded our internal tests without breaking the monorepo unit tests, then, create unit tests that cover this weak point.

## Why is this task needed? 

Because it will inevitably happen again.

## Any other comments?

See:
- https://github.com/gravwell/gravwell/pull/1964

Issue #1209, opened 2024-08-27
chancacher: quarantine caches that cannot be opened


david-fritz-gravwell:

daniel-moreno-levy-gravwell:
Some useful context from mattermost:

>  Trying to understand this issue: https://github.com/gravwell/gravwell/issues/1209 From what I've been able to gather,  looks like this quarantine should happen here?  https://github.com/gravwell/gravwell/blob/8f858d5e33b9dcfdf54e4f17f7fd78be1fa9f0cf/chancacher/chancacher.go#L165-L174 ? 

> yes that's probably where we'd quarantine and quarantine here just means to move the file off to a <filename>.1, .2, .3... better to never delete a cache but also try to get to a running state

daniel-moreno-levy-gravwell:
Reopening, see https://github.com/gravwell/gravwell/issues/1967#issuecomment-3603122167

PR #1975, opened 2025-12-03
feat(#1967,#1209): Quarantine ingester cache files when faulty

daniel-moreno-levy-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/1967 and https://github.com/gravwell/gravwell/issues/1209

### DIFF
--- a/chancacher/chancacher_test.go
+++ b/chancacher/chancacher_test.go
@@ -818,7 +818,7 @@ func Test_validateCache(t *testing.T) {
 	}
 
 	// Verify original file is gone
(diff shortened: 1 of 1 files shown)


daniel-moreno-levy-gravwell:
Main additions:

- Added a unit test that mimics a faulty muxer trying to send entries repeatedly. This is a very similar setup to the internal tests that broke.
- Moved the file locks BEFORE trying to open the cache files.
- openCache -> validateCache, and as the name implies, only validates that the cache is readable with the give permissions and can be decoded. It no longer tries to return a file handle to the final cache file.
- Modified other unit tests to accommodate changes.

PR #1986, opened 2025-12-09
types: add binning fields related to the results API

david-fritz-gravwell:
This PR:

- Adds binning related fields (see https://github.com/gravwell/api-docs/pull/512) for the results API

### DIFF
--- a/client/types/render.go
+++ b/client/types/render.go
@@ -100,8 +100,6 @@ const (
 
 	MetadataTypeRaw    string = `raw`
 	MetadataTypeNumber string = `number`
-
-	OPTIONS_BINS string = `bins`
 )
 
 const (


client/types/search.go
```diff
@@ -296,6 +305,9 @@ type SearchInfo struct {
 	Error string `json:",omitempty"`
 
 	LaunchInfo SearchLaunchInfo // information about how a search was launched
+
+	// Binning describes whether search results can be condensed (binned) and, if not, why.
+	Binning BinningStatus `json:",omitempty"`
 }
```
evan-gravwell: I think this field was added to the wrong corresponding api-docs type? IE this seems like the type corresponding to api-docs' `SearchCtrlStatus` (which does not have a Binning field spec'd in api-docs).

evan-gravwell (changes requested):
Two comments.

evan-gravwell (approved):
We talked over MM. I was confused how two types from api-docs were being mapped to the single type here. LGTM.

Issue #784, opened 2023-07-31
Global Attach functionality in a preprocessor


kris-watts-gravwell:
The global attach logic is cool and all, but having that same functionality in a preprocessor would also add some extra options.

should be an extremely small pre-processor

kris-watts-gravwell:
bumping, this is useful

kris-watts-gravwell:
Another bump

PR #2007, opened 2025-12-17
Preprocessor 784: Added new attach preprocessor

mike-wade-gravwell:
This PR addresses [ISSUE_LINK](https://github.com/gravwell/gravwell/issues/784)

## This PR proposes to add a new preprocessor to use the attach functionality at the listener level. 

- Adds a new preprocessor called attach that attaches values at the listener level instead of the ingester level.
- Uses the same code from attach
Note: The $UUID functionality is not available at the listener level and is replaced by a UUID of all 0's

## PR Tasks

- [X] e2e and/or unit tests included. If not, please provide an explanation.

## Reviewer Tasks


ingest/processors/attach.go
```diff
@@ -0,0 +1,106 @@
...
+	c.Vals = make(map[gcfg.Idx]*[]string, len(vc.Vals))
+	for k, v := range vc.Vals {
+		c.Vals[gcfg.Idx(k)] = v
+	}
+
+	// Filter out the "type" key which is used for preprocessor selection
+	// but should not be attached as an enumerated value
+	delete(c.Vals, c.Idx("type"))
+
+	err = c.Verify()
```
kris-watts-gravwell: right here, if err == nil, swing through the map of c.Vals and check for a key of `$UUID`, if you see it, throw an error.

This way we can warn the user rather than push in useless bytes.

ingest/processors/attach.go
```diff
@@ -0,0 +1,106 @@
...
+	return
+}
+
+// NewAttachProcessor creates a new attach processor
+func NewAttachProcessor(cfg attach.AttachConfig) (*AttachProc, error) {
+	if err := cfg.Verify(); err != nil {
+		return nil, err
+	}
+	// Use zero UUID - $UUID is not available in the preprocessor version
```
kris-watts-gravwell: duplicate comments?

ingest/processors/attach_test.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+
+	if val, ok := result[0].GetEnumeratedValue("host"); !ok {
+		t.Error("Expected 'host' enumerated value")
+	} else if s, ok := val.(string); !ok || s != expectedHostname {
+		t.Errorf("Expected host='%s', got %v", expectedHostname, val)
+	}
+}
+
+func TestAttachProcessorUUID(t *testing.T) {
```
kris-watts-gravwell: make sure this just fails at config time.

kris-watts-gravwell (changes requested):
update copyright headers and make preprocessor fail on attempts to use UUID rather than stuff in useless bytes on each entry

ingest/processors/attach_test.go
mike-wade-gravwell: Two tests were addressed. One for a standard config and one where the UUID was injected into a valid config

PR #2012, opened 2025-12-17
client: Fix up some other functions on /api/users/<id>

john-floren-gravwell:
Getting/Setting default search groups and updating user info needs to take into account that the /api/users/<id> endpoint is now admin-only


client/admin.go
```diff
@@ -138,10 +138,21 @@ func (c *Client) UserChangePass(id int32, orig, pass string) error {
...
-	udet := types.User{}
-	if err := c.methodStaticURL(http.MethodGet, usersInfoUrl(uid), &udet); err != nil {
+	me, err := c.MyInfo()
+	if err != nil {
 		return err
 	}
+	udet := me
+	if uid != me.ID {
+		if !me.Admin {
+			return errors.New("Only admins can change another user's default search groups")
```
kris-watts-gravwell: probably drop this into a variable rather than repeat it 3 times

PR #2056, opened 2026-01-27
feat: Okta Ingester

nic-osullivan-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/2074

Most of this was done by @kris-watts-gravwell, but I've added a bit and cleaned some parts up. 

## This PR proposes...
The creation of a baseline hosted ingester that can run as a normal ingester. This is intended as a platform for future ingesters that primarily poll for info rather than receive a stream of information. This is helpful for audit apis (eg Okta, mimecast) that don't support any form of event streaming. 

Also included is a change to allow any supported log level as a value for `Log-Level` giving us some visibility into debug level logs (3rd party apis are tricky)

## PR Tasks

## Reviewer Tasks

For reviewers there are 3 main sections of this PR, The runner `ingesters/hosted/runner`, the okta fetcher `ingesters/hosted/plugin/okta`, and the runtime interface `ingesters/hosted`. There are small number of changes in the core `ingest` libs mostly to support the other work. 

The runner is the main piece that enables running fetchers as a "normal" ingester. The hosted interface and native runtime are the main pieces that will be reused in the future with further integrations. Ideally only a new fetcher needs to be written and minimal modifications to core behavior, expect for adding configs, will be needed. 

Fetchers should not rely on how they get their config structures handed in (eg they may come from a file, or a api call) in order to keep coupling low.   

### DIFF
--- a/ingesters/hosted/storage/bolt.go
+++ b/ingesters/hosted/storage/bolt.go
@@ -261,10 +261,3 @@ func retslice(v []byte) (r []byte) {
 	}
 	return
 }
-
-// parseInt64 converts a byte slice to an int64
-func parseInt64(v []byte) (int64, error) {
-	var value int64
-	_, err := fmt.Sscanf(string(v), "%d", &value)
-	return value, err
-}


ingesters/hosted/native.go
```diff
@@ -0,0 +1,333 @@
...
+
+// Close stops the running routine, collects the error and returns
+func (nr *NativeRunner) Close() (err error) {
+	if nr == nil || nr.rt == nil {
+		return errors.New("not ready")
+	}
+	nr.cf()
+	// wait for routine to exit TODO FIXME - add a timeout on this wait
+	// WaitGroup probably isn't the right tool here
+	nr.wg.Wait()
```
daniel-moreno-levy-gravwell: This *feels* like you want to use https://pkg.go.dev/golang.org/x/sync/errgroup instead... Might be easier since you can pass a context with a deadline to it.

ingesters/hosted/plugins/tester/tester.go
```diff
@@ -0,0 +1,101 @@
...
+const (
+	defaultInterval = time.Second // by default we fire an entry every second
+)
+
+type Config struct {
+	Ingester_UUID string // set the UUID for the ingester
+	Interval      string // how often to send an entry, this should be a string parsable by time.ParseDuration
+}
+
+func (c *Config) Verify() (err error) {
```
daniel-moreno-levy-gravwell: I'm seeing `Verify` and `UUID` as part of the config structs in both plugin ingesters... Might be worth adding an interface with these in `interface.go`?

ingesters/hosted/plugins/config.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/google/uuid"
+	"github.com/gravwell/gravwell/v3/ingesters/hosted"
+
+	// include all the native hosted ingesters
+	"github.com/gravwell/gravwell/v3/ingesters/hosted/plugins/okta"
+	"github.com/gravwell/gravwell/v3/ingesters/hosted/plugins/tester"
+)
+
+type Configs struct {
```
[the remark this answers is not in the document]
daniel-moreno-levy-gravwell: Ignore this, I re-read the PR and is clear that is not the plan.

ingesters/hosted/native.go
nic-osullivan-gravwell: That is what I wanted for sure, cleaned up some mutex handling nicely as well. All around cleaner.

ingesters/hosted/storage/bolt.go
```diff
@@ -0,0 +1,270 @@
+/*************************************************************************
```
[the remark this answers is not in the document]
nic-osullivan-gravwell: moved to `storage/`, but we still have some semi direct coupling to this implementation, but I noted that in the config structure as that's the only user facing coupling. The coupling is small enough likely not an issue for now though.

ingesters/hosted/storage/bolt.go
```diff
@@ -0,0 +1,270 @@
...
+		NoSync:  !sync,
+		Timeout: time.Second,
+	}
+
+	info, err := os.Stat(pth)
+	exists := !os.IsNotExist(err)
+	if err != nil && exists { // if there is no file, bolt.Open will create it
+		return
+	}
+	if exists && info.Mode().Perm()&os.ModePerm&0600 != 0 { // exists && !writable
```
daniel-moreno-levy-gravwell: I think os.ModePerm is not necessary here. I checked the value out of curiosity and it is 0777. 

```
	ModePerm = fs.ModePerm // Unix permission bits, 0o777
```

So doing `info.Mode().Perm() & 0o600`  will already narrow the bits even further than 0o777.

ingesters/hosted/runner/main.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	// ingesters exist, fire them up
+	if err = rm.startIngesters(); err != nil {
+		rm.stop()  // best effort close
+		sh.Close() // ignore return, but no writes should have occurred
+		ib.Logger.FatalCode(2, "failed to start ingesters", log.KVErr(err))
+	}
+
+	//TODO FIXME - register child ingesters on our base ingester for reporting
```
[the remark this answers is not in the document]
nic-osullivan-gravwell: I removed the todo, but created a follow up. https://github.com/gravwell/gravwell/issues/2073 I think this would require a fair bit of dedicated work to make happen (mostly around how the forEachIngester methods work), and as we are just going to a preview this should be fine to not have yet.

daniel-moreno-levy-gravwell (approved):
LGTM!

PR #2078, opened 2026-02-09
FIX(N/A): gwcli "missing/unknown column" panic on startup 

rory-landau-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/1138.

## TL;DR

Struct changes caused certain actions to panic at startup. This PR fixes those actions and updates the build process to loudly fail should this reoccur. The design could use a sniff test.

---

With the introduction of common fields and similar struct changes in registry, several actions broke with "invalid/missing column" panics. Actions operate on generic "scaffolding". To select which columns (fields) they should show, the action developer provides a _dot-qualified_ path to the field so the scaffolding can walk the structs and extract data. If the struct changes, the paths change. If the paths change, the scaffolding fails to extract data and panics.

This PR updates DQ'd paths and adds guardrails to detect changes.
Package `weave` has been upgraded to generate mapping for arbitrarily-nested struct fields; should the fields change, so too will these mappings. Instead of providing the path as a string, developers can guarantee existence at compile time using the mappings weave generates.

These mappings are capable of supporting anonymous, embedded, and nested structs and have controls for mapping exported vs all fields. This has been integrated into the build process so compilation will fail if it is not addressed. Currently, all mappings are stored in `/internal/typemap`.

**NOTE**: I am not sold on the current design and would really appreciate a second opinion on it. Further, [typemap](https://github.com/rory-landau-gravwell/gravwell/blob/c936255d74e6fb4f98b22bd98859a5040b30d83a/gwcli/internal/typemap/typemap.go) is prone to balloon in size. Disk space is cheap, but still something to consider. 

## Reviewer Tasks


daniel-moreno-levy-gravwell:
I agree with @rory-landau-gravwell . I understand @john-floren-gravwell point that adding a new tool is adding complexity, but the benefit of not depending on developer attention (self-checking system instead) is worth it. The point about having to remember to run `mage build` is valid, but I wonder if we can just make a github action job to check the code is in "the correct state" (like `mage build && git diff --exit-code`)?

nic-osullivan-gravwell:
I have no objection to mage (and like it even), but I think this should be done via go generate if possible. Even having mage build call go generate and go build would be fine. I don't think we want the _only_ way to do something to be in mage. I think it should be closer to make and glue commands together rather than be core behavior. 

I've been thinking we likely need _something_ in a precommit hook as a standard across contributors.

PR #2097, opened 2026-02-12
feat: Hosted Mimecast ingester

nic-osullivan-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/1998

## This PR proposes...

- Adding a Mimecast Hosted Ingester. Much like the Okta ingester before we want to make it easier for people to ingest data from 3rd party systems.
- This makes a few adjustments to the base hosted runner for a few things that were missed in the initial implementation. 

## PR Tasks

## Reviewer Tasks

This does likely need a thumbs up from CS before merging as they wrote the initial experimental mimecast fetcher. 
- [ ] CS Approval

### DIFF
--- a/tools/actions/kitmanager/kitwork.go
+++ b/tools/actions/kitmanager/kitwork.go
@@ -1,5 +1,5 @@
 /*************************************************************************
- * Copyright 2025 Gravwell, Inc. All rights reserved.
+ * Copyright 2026 Gravwell, Inc. All rights reserved.
  * Contact: <legal@gravwell.io>
  *
  * This software may be modified and distributed under the terms of the
@@ -9,15 +9,18 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
 
 	"github.com/google/uuid"
 	"github.com/gravwell/gravwell/v3/client"
 	"github.com/gravwell/gravwell/v3/client/types"
+	"github.com/gravwell/gravwell/v3/client/types/kits"
 )
 
 // pullKit reaches out to the remote Gravwell instance and performs a kit build using the existing kit build request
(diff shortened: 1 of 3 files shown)


ingesters/hosted/storage/bolt.go
```diff
@@ -56,7 +56,7 @@ func OpenBoltHandler(pth string, sync bool) (sh *BoltHandler, err error) {
 	if err != nil && exists { // if there is no file, bolt.Open will create it
 		return
 	}
-	if exists && info.Mode().Perm()&0600 != 0 { // exists && !writable
+	if exists && info.Mode().Perm()&0o200 == 0 { // exists && !writable
```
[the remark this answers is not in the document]
nic-osullivan-gravwell: same... that's what I get for trusting the test Kris wrote I guess.

ingesters/hosted/plugins/config.go
daniel-moreno-levy-gravwell: I'm seeing a bit of repetition and error proneness here. To add a new config, you need to add the implementation to `Tags`, `IngesterCounts` and `ForEachIngester`. The pattern is super repeatable, which is good, but it is spread out through out the file, so it is easy to forget one part.

These whole setup runs only once at boot, it might be worth to see if we could do some reflection to for loop through the `Configs` struct to get the the tags, counts and for each ingesters. For example:

```golang
     func (c Configs) IngesterCount() int {                                                                                     
       v := reflect.ValueOf(c)                                                                                                  
       count := 0                                                                                                               
       for i := 0; i < v.NumField(); i++ {                                                                                      
         count += v.Field(i).Len()                                                                                              
       }                                                                                                                        
       return count                                                                                                             
     }                                                                                                                          
```

I'm sure the others can be changed around for this. Not a blocker so if you want to delegate to another ticket that is fine.

ingesters/hosted/runner/example.conf
```diff
@@ -23,10 +23,21 @@
...
 	#Request-Burst=20 #optionally override request burst (leaky bucket limited), default is 10
+
+#[Mimecast "mimecast"]
+	#Ingester-UUID="99000000-0000-0000-0000-000000000000" #example UUID, remember to set
+	#Lookback=48h
+	#Client-Id="id"
+	#Client-Secret="secret"
+	#Api=audit
+	#Api=mta-delivery
+	#Tag-Prefix="mimecast"
```
daniel-moreno-levy-gravwell: Should include Tag-Name here too, and disclaimer that i Tag-Name or Tag-Prefix, not both together

ingesters/hosted/plugins/mimecast/example.conf
```diff
@@ -0,0 +1,44 @@
+[Global]
```
daniel-moreno-levy-gravwell: Why this file compared to runner/example.conf?

daniel-moreno-levy-gravwell (changes requested):
Some changes requested, other comments are more clarifying questions rather changes. Generally looks good and it is very close to a LGTM.

ingesters/hosted/plugins/mimecast/example.conf
nic-osullivan-gravwell: runner/example.conf would include an example from every ingester, this is just for mimecast and is more used for local testing.

ingesters/hosted/runner/example.conf
```diff
@@ -23,10 +23,21 @@
...
+	#Interval=5s
+
```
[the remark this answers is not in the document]
daniel-moreno-levy-gravwell: Ok, as long as we are keeping track of it and point to it in the README I'll be happy.

PR #2170, opened 2026-03-04
fix: duplicated mimecast events

nic-osullivan-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/1998

## This PR proposes...

Adjust timestamp and timerange handling for SIEM api requests, as they are quite strict on proper validation. 

## PR Tasks

## Reviewer Tasks


daniel-moreno-levy-gravwell (dismissed):
LGTM... A few nags but honestly, no show stoppers, up to you if you want to address them.

nic-osullivan-gravwell:
@daniel-moreno-levy-gravwell this is good to go again.

Main changes since last review
- Everything using timerange
- updating how we progress forward and use tokens/cursors (apparently on the batch api we can be caught up and still need the token 🤷 )
- removed some unused code
- updated the hosted logger implementation to use the `*WithDepth` methods to properly report call sites in logs
- made the e2e test connection timeout for ingesters configurable from a single point rather than per test. 
- when shutdown with e2e.Terminate we now capture container stdout and write that to a file for debugging.

ingesters/hosted/plugins/mimecast/mimecast.go
```diff
@@ -204,41 +202,50 @@ func (m *Mimecast) mtaEvent(ctx context.Context, rt hosted.Runtime, api Api) err
...
-			rt.Debug("got another page of events", log.KV("api", api))
-			rt.PutString(m.cursor(api), r.NextPage)
 		}
 	}
 
 	return nil
 }
 
-func (m *Mimecast) handleMtaPage(rt hosted.Runtime, tag entry.EntryTag, page []json.RawMessage) (*time.Time, error) {
+func (m *Mimecast) handleMtaPage(rt hosted.Runtime, tag entry.EntryTag, page []json.RawMessage, api Api) (*time.Time, error) {
```
daniel-moreno-levy-gravwell: (Put this in the wrong func, I meant to put it in HandleMtaBatch, though the same criticism applies to this for returning time pointer.)

Not sure why we return a `time` pointer here... Seeing how it is used in:

```golang
		if last == nil || last.IsZero() { // there were no events in the range
			last = &tr.End
		}
```

If this returned a value, we could just simply check with `last.IsZero`, unless I am missing something, seems like making it a pointer is just making it more complicated for no obvious gain.

daniel-moreno-levy-gravwell: Additionally I would not pass `timerange` as a pointer either, since it isn't modified, and doesn't to be used for anything other than debug logs.

daniel-moreno-levy-gravwell (changes requested):
Small nits around function arguments, otherwise LGTM

ingesters/hosted/plugins/mimecast/mimecast.go
nic-osullivan-gravwell: Dropped the raw `time.Time` pointers, but left the `TimeRange` one as wanted to avoid having to copy that on every call for every page, it's not a small amount of calls that will hit this in some scenarios.

daniel-moreno-levy-gravwell (approved):
LGTM

PR #2231, opened 2026-03-24
Implement APIs and types for automations in the registry

john-floren-gravwell:
Scheduled searches, scheduled scripts, and flows are now their own types with their own separate *results* types.

### DIFF
--- a/client/types/kits.go
+++ b/client/types/kits.go
@@ -259,47 +259,47 @@ func (pbr *KitBuildRequest) Validate() error {
 	}
 	for i := range pbr.ScheduledSearches {
 		if pbr.ScheduledSearches[i] == "" {
-			return fmt.Errorf("Invalid scheduled search ID %s", pbr.ScheduledSearches[i])
+			return fmt.Errorf("empty scheduled search ID")
 		}
 	}
 	for i := range pbr.ScheduledScripts {
 		if pbr.ScheduledScripts[i] == "" {
-			return fmt.Errorf("Invalid scheduled script ID %s", pbr.ScheduledScripts[i])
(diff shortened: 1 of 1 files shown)


client/types/api.go
```diff
@@ -322,13 +322,13 @@ type SearchAgentConfig struct {
...
 	Disable_Self_Ingest              bool // disables ingesting search agent logs to indexers
 	HTTP_Proxy                       string
-}
 
-type SearchAgentCheckin struct {
-	LastCheckin     string
-	Warning         bool
-	SearchAgentUUID string
-	SearchAgentIP   string
+	Search_Rate  int64 // searches launched per second
```
rory-landau-gravwell: Why are these Pascal Snake? It matches with the convention used in SearchAgentConfig, so feel free to disregard.

client/types/kits.go
```diff
@@ -257,13 +258,18 @@ func (pbr *KitBuildRequest) Validate() error {
 		pbr.Resources[i] = strings.TrimSpace(pbr.Resources[i]) //clean it
 	}
 	for i := range pbr.ScheduledSearches {
-		if pbr.ScheduledSearches[i] <= 0 {
-			return fmt.Errorf("Invalid scheduled search/script ID %d", pbr.ScheduledSearches[i])
+		if pbr.ScheduledSearches[i] == "" {
+			return fmt.Errorf("Invalid scheduled search ID %s", pbr.ScheduledSearches[i])
```
rory-landau-gravwell: This error message will always read "Invalid scheduled search ID ", as you throw the error iff the element is empty.

rory-landau-gravwell (approved):
Comments addressed.

LGTM

PR #2281, opened 2026-04-07
Make user & group APIs look more like other registry assets

john-floren-gravwell:
Biggest change is that listing now works the same as for other assets, so you can paginate.


client/users.go
```diff
@@ -0,0 +1,143 @@
...
+func (c *Client) ListAllUsers(opts *types.QueryOptions) (ret types.UserListResponse, err error) {
+	if opts == nil {
+		opts = &types.QueryOptions{}
+	}
+	opts.AdminMode = true // we'll reject this if the user isn't actually an admin
+	err = c.postStaticURL(USERS_LIST_URL, opts, &ret)
+	return
+}
+
+// GetUserMap returns a map of UID to username for every user on the system.
```
rory-landau-gravwell: Probably should mark this as admin-only as it uses ListAllUsers under the hood.

client/users.go
```diff
@@ -0,0 +1,143 @@
...
+	return c.methodStaticPushURL(http.MethodPut, usersInfoUrl(id), req, nil, nil, nil)
+}
+
+// CleanupUsers (admin-only) purges all deleted users for all users.
+func (c *Client) CleanupUsers() error {
+	return c.deleteStaticURL(USERS_URL, nil)
+}
+
+// LookupUser looks up a User object given a username
+// if the username is not found, ErrNotFound is returned
```
rory-landau-gravwell: Also admin-only?

rory-landau-gravwell (commented):
Changes look fine, but is it worth marking subroutines that rely on ListAll* (as noted below for Users, but also for groups) as admin-only? *We* know that ListAll will turn back normal List results for a non-admin, but that probably isn't useful for mapping commands as users and groups are mostly admin-only.

john-floren-gravwell:
Actually, users and groups are kinda different... if you have the List[Users|Groups] CBAC capability, you're allowed to list users and groups. There's no difference with the admin mode set or unset. To that end, I'm wondering if I should just remove ListAll[Users|Groups] entirely, since they should make zero difference.

rory-landau-gravwell:
> Actually, users and groups are kinda different... if you have the List[Users|Groups] CBAC capability, you're allowed to list users and groups. There's no difference with the admin mode set or unset. To that end, I'm wondering if I should just remove ListAll[Users|Groups] entirely, since they should make zero difference.

I don't want users getting results they don't understand the limiting factor of; sucks to call UserMap and get a map of 1 (the caller). Files/resources/macros/etc have a clear difference between "I want mine" vs "I want all". "I want mine" makes less sense for users/groups (outside of "get me the groups I am in").

If we can say "List always turns back either you or all users|groups, check CBAC", then *let's junk ListAll* for users/groups.

Lastly, is this influenced in any way by community editions that do not have access to CBAC?

john-floren-gravwell:
> > Actually, users and groups are kinda different... if you have the List[Users|Groups] CBAC capability, you're allowed to list users and groups. There's no difference with the admin mode set or unset. To that end, I'm wondering if I should just remove ListAll[Users|Groups] entirely, since they should make zero difference.
> 
> I don't want users getting results they don't understand the limiting factor of; sucks to call UserMap and get a map of 1 (the caller). Files/resources/macros/etc have a clear difference between "I want mine" vs "I want all". "I want mine" makes less sense for users/groups (outside of "get me the groups I am in").
>
> If we can say "List always turns back either you or all users|groups, check CBAC", then _let's junk ListAll_ for users/groups.
> 
> Lastly, is this influenced in any way by community editions that do not have access to CBAC?

You'll actually get a map of nothing if you don't have the ListUsers CBAC capability, as implemented. The entire endpoint is gated behind that cap.

When CBAC isn't enabled, everybody is allowed to list all users.

rory-landau-gravwell:
> You'll actually get a map of nothing if you don't have the ListUsers CBAC capability, as implemented. The entire endpoint is gated behind that cap.
> 
> When CBAC isn't enabled, everybody is allowed to list all users.

In that case, I want to throw away ListAll and make these behaviors explicit in List; I would have expected a map of only myself (or a permission error) and I'm sure I can't be the only fool.

john-floren-gravwell:
> > You'll actually get a map of nothing if you don't have the ListUsers CBAC capability, as implemented. The entire endpoint is gated behind that cap.
> > When CBAC isn't enabled, everybody is allowed to list all users.
> 
> In that case, I want to throw away ListAll and make these behaviors explicit in List; I would have expected a map of only myself (or a permission error) and I'm sure I can't be the only fool.

Actually you're right, you'll get an error if you call List without having the appropriate capability.

rory-landau-gravwell:
> > > You'll actually get a map of nothing if you don't have the ListUsers CBAC capability, as implemented. The entire endpoint is gated behind that cap.
> > > When CBAC isn't enabled, everybody is allowed to list all users.
> > 
> > 
> > In that case, I want to throw away ListAll and make these behaviors explicit in List; I would have expected a map of only myself (or a permission error) and I'm sure I can't be the only fool.
> 
> Actually you're right, you'll get an error if you call List without having the appropriate capability.

Even better. Then I'm less concerned about documenting the behavior

rory-landau-gravwell (dismissed):
LGTM!

rory-landau-gravwell (approved):
Good catch. Double-checked the repo for any other ListAllUsers/ListAllGroups; found none. LGTM

PR #2289, opened 2026-04-09
refactor (2099): improve hosted runner startup

nic-osullivan-gravwell:
This PR addresses gravwell/issues#1965

## This PR proposes...

Refactoring the creation and startup of hosted ingesters/runners to reduce the use of callbacks in favor of iterators for a more flexible and "linear" logical startup. The individual callbacks were passed around and it wasn't really clear what was actually responsible for what. Now the plugins are responsible for providing a consistent interface for building and the manager handles everything else. 

## PR Tasks

## Reviewer Tasks

### DIFF
--- a/ingesters/hosted/plugins/builder.go
+++ b/ingesters/hosted/plugins/builder.go
@@ -1,3 +1,11 @@
+/*************************************************************************
+ * Copyright 2026 Gravwell, Inc. All rights reserved.
+ * Contact: <legal@gravwell.io>
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD 2-clause license. See the LICENSE file for details.
+ **************************************************************************/
+
 package plugins
 
 import (
--- a/ingesters/hosted/plugins/builder.go
+++ b/ingesters/hosted/plugins/builder.go
@@ -1,0 +1,97 @@
+package plugins
+
+import (
+	"github.com/google/uuid"
+	"github.com/gravwell/gravwell/v3/ingesters/hosted"
+	"github.com/gravwell/gravwell/v3/ingesters/hosted/plugins/mimecast"
+	"github.com/gravwell/gravwell/v3/ingesters/hosted/plugins/okta"
+	"github.com/gravwell/gravwell/v3/ingesters/hosted/plugins/tester"
+)
+
+type BuilderConfig interface {
+	UUID() uuid.UUID
+}
+
+type Builder[T BuilderConfig] struct {
+	config  T
+	kind    string
+	id      string
+	version string
+}
+
+func (b *Builder[T]) Config() any {
+	return b.config
+}
+
+func (b *Builder[T]) Kind() string {
+	return b.kind
+}
+
+func (b *Builder[T]) ID() string {
+	return b.id
+}
+
+func (b *Builder[T]) Version() string {
+	return b.version
+}
+
+func (b *Builder[T]) UUID() uuid.UUID {
+	return b.config.UUID()
+}
+
+type TesterBuilder struct {
+	Builder[*tester.Config]
+}
+
+func (tb *TesterBuilder) Build(tn hosted.TagNegotiator) (hosted.Ingester, error) {
+	return tester.NewTesterIngester(*(tb.config), tn)
+}
+
+func NewTesterBuilder(config *tester.Config, kind, id, version string) *TesterBuilder {
+	return &TesterBuilder{
+		Builder[*tester.Config]{
+			config:  config,
+			kind:    kind,
+			id:      id,
+			version: version,
+		},
+	}
+}
+
+type OktaBuilder struct {
+	Builder[*okta.Config]
+}
+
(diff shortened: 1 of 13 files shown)


ingesters/hosted/plugins/builder.go
```diff
@@ -0,0 +1,97 @@
+package plugins
```
rory-landau-gravwell, line 9: Please add file header
```
/*************************************************************************
 * Copyright 2026 Gravwell, Inc. All rights reserved.
 * Contact: <legal@gravwell.io>
 *
 * This software may be modified and distributed under the terms of the
 * BSD 2-clause license. See the LICENSE file for details.
 **************************************************************************/
```

ingesters/hosted/plugins/builder.go
rory-landau-gravwell, line 1: Maybe also a couple comments attached to the BuilderConfig interface and/or generic struct; the code is reasonably self explanatory, but comments are a nicety we can afford. At least on the major components.

ingesters/hosted/plugins/config.go
```diff
@@ -73,73 +74,31 @@ func (c Configs) IngesterCount() (count int) {
 type NewIngesterCallback func(id, name string, runner hosted.Runner) error
```
rory-landau-gravwell: These two don't appear to be used anymore; are we expecting them to be used in the future or can we junk them?

ingesters/hosted/plugins/config.go
```diff
@@ -73,73 +74,31 @@ func (c Configs) IngesterCount() (count int) {
...
+type IngesterBuilder interface {
+	UUID() uuid.UUID
+	Kind() string
+	ID() string
+	Version() string
+	Build(hosted.TagNegotiator) (hosted.Ingester, error)
+	Config() any
+}
+
+func (c Configs) Builders() iter.Seq2[string, IngesterBuilder] {
```
rory-landau-gravwell, line 96: Probably could also use a header comment. Though it is a common pattern, noting that False kills the iter would be nice.

ingesters/hosted/runner/config.go
```diff
@@ -64,18 +61,3 @@ func (c cfgType) AttachConfig() attach.AttachConfig {
 func (c cfgType) IngestBaseConfig() config.IngestConfig {
 	return c.Global
 }
-
-func (c cfgType) forEachIngester(tn hosted.TagNegotiator, nrt plugins.NewRuntimeCallback, cb plugins.NewIngesterCallback) (err error) {
```
rory-landau-gravwell, line 68: This doesn't have a replacement. It was internal only. Is this just junking relic code?

ingesters/hosted/runner/manager.go
```diff
@@ -70,6 +70,37 @@ func (rm *runtimeManager) stop() (err error) {
 	return
 }
 
+func (rm *runtimeManager) createRunners(c *cfgType, ib base.IngesterBase) (err error) {
+	for name, builder := range c.Builders() {
```
rory-landau-gravwell, line 77: c isn't nil checked and could panic. builder isn't either but c.Builders always returns new builders so I'm not particularly concerned.

rory-landau-gravwell (changes requested):
Code looks good, clean, smart. Tests pass. Below points are super minor.

ingesters/hosted/runner/config.go
nic-osullivan-gravwell: `createRunners` calling `Builders()` is the replacement. The small layer provided by the builders allows that to be a bit more generic. rather than having to pass multiple callbacks down to the configs. 

Basically with this the config doesn't need to know about the runtime. 

The old flow was manager `createIngesters` > runner config `forEachIngester` > plugin config `forEachIngester` > plugin config `buildIngesters` now it's just `createRunners` > plugin config `Builders()` and we just iterate loop the list directly. Removed some layers of indirection so it's easier to follow what is happening in what order.

nic-osullivan-gravwell:
@rory-landau-gravwell good catches on the unused code though I got it all, but alas. Got the docs added as well so that should help clear things up.

ingesters/hosted/runner/manager.go
```diff
@@ -71,6 +71,9 @@ func (rm *runtimeManager) stop() (err error) {
 }
 
 func (rm *runtimeManager) createRunners(c *cfgType, ib base.IngesterBase) (err error) {
+	if rm == nil {
```
rory-landau-gravwell: er, I meant nil checking `c` prior to invoking a method on it (`c.Builders`), not `rm`; I don't believe a nil `rm` could get this far

rory-landau-gravwell (changes requested):
Minor nil check miscommunication; once that is fixed, this PR LGTM

ingesters/hosted/runner/manager.go
nic-osullivan-gravwell: Ah right right, done!

rory-landau-gravwell (approved):
LGTM!

PR #2291, opened 2026-04-09
Remove the permissions-specific filters from QueryOptions struct

john-floren-gravwell:
Small tweaks on the backend allow you to filter via the Filters slice just like other fields.


evan-gravwell (approved):
LGTM!

PR #2310, opened 2026-04-17
Implement changes for alerts in the registry

john-floren-gravwell:
Implements types and APIs for the new-style alerts in the registry.

Better listing means we can get rid of special functions to fetch all alerts with a specific consumer/dispatcher, thankfully.


rory-landau-gravwell (approved):
LGTM. And thank you for updating gwcli; you didn't have to do that, but I appreciate it.

PR #2321, opened 2026-04-21
Implement actionable (pivot) contents

john-floren-gravwell:
Getting these types into next-patch so they're available in a v3 import on the registry work.


ashnwade:
You don't have to do this, but you can re-target this PR to `next-minor` if you want to avoid the pubsub error. The change would still go into the product before 6.0.0

john-floren-gravwell:
Actually I really think @evan-gravwell needs to review this, it's more of a frontend thing

david-fritz-gravwell (dismissed):
LGTM

client/types/actionables.go
```diff
@@ -0,0 +1,112 @@
...
+// StrictPivot is a Pivot which actually decodes the Contents field,
+// rather than treating it as a RawObject. This is for migration to
+// the registry.
+type StrictPivot struct {
+	GUID        uuid.UUID
+	Name        string
+	Description string
+	Contents    ActionableContent
+	Labels      []string
+	Disabled    bool
```
evan-gravwell: js-client has some other fields. I'm unsure if they're necessary for migration purposes.

ThingUUID
UID
GIDs
Global
Updated - timestamp

evan-gravwell (changes requested):
Just a Q. Otherwise, LGTM.

client/types/actionables.go
john-floren-gravwell: Good catch, got lost in the shuffle

john-floren-gravwell: On reflection, we don't actually need to define a StrictPivot at all... we're only using these types here in 6.0 as something to unpack existing pivots into, and then we'll upgrade to the tweaked new types. See latest commit

evan-gravwell (approved):
LGTM!

Issue #2306, opened 2026-04-15
S3/SQS ingester does not report WHICH queue does not exists


kris-watts-gravwell:
# Bug Report

Before opening a bug report, please complete the following checklist:
- [ ] Check the documentation for intended behavior, notes, warnings, and edge cases
- [ ] Check open AND closed issues to see if this is a duplicate
	- A closed duplicate issue may exist if this is a wontfix

## What are the steps to reproduce this issue?

1. configure SQS/S3 with valid creds but at a queue that does not exist

## What happens?

you get the handy dandy error
```
sqs receive message error error="AWS.SimpleQueueService.NonExistentQueue: The specified queue does not exist."
```

so.... which queue?  if i have multiple configured, which one doesn't exist?

## What were you expecting to happen?

Give me the configuration block name and/or queue path

## Any logs, error output, etc?

see above

## What versions of software are you using?

5.8.12

kris-watts-gravwell:
This doesn't feel quite right, Rather than the SQS queue path I would like to see the listener block

consider this config
```
[SQS-S3-Listener "sqs-testing-draw"]
	Region="us-west-1"
	Tag-Name="cloudtrail"
	Queue-URL="URL"
	ID="<My ID>"
	Secret="<MY SECRET>"
	Credentials-Type=static
	Reader="cloudtrail"

```


I would like to see "sqs-testing-draw" NOT the giant URL

kris-watts-gravwell:
v5.9.0-beta7 LGTM

PR #2340, opened 2026-04-27
feat(2306): add Endpoint to SQS config and add queue to error message

tyler-elton-gravwell:
This PR addresses [2306](https://github.com/gravwell/gravwell/issues/2306)

## This PR proposes...

- Adding the `Endpoint` configuration value to SQS to support SQS-compatible APIs outside of AWS (ex: ElasticMQ)
- Wraps all SQS errors with the queue to better isolate which queue is returning the error (ex: queue doesn't exist)
- Adds some tests to `sqs_common`, `ingesters/s3Ingester`, and `ingesters/sqsIngester`

## PR Tasks

## Reviewer Tasks


sqs_common/sqs.go
```diff
@@ -77,7 +83,7 @@ func (s *SQS) GetMessages() ([]*sqs.Message, error) {
 	for out == nil || len(out.Messages) == 0 {
 		out, err = s.svc.ReceiveMessage(req)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("queue '%s': %w", s.Queue(), err)
```
nic-osullivan-gravwell: I think it's worth expanding this error message a bit, as these make it to users. Something like:

```suggestion
			return nil, fmt.Errorf("error getting messages on queue %q: %w", s.Queue(), err)
```

or this if we don't want to add the extra detail

```suggestion
			return nil, fmt.Errorf("error: %w, on queue %q", err, s.Queue())
```

sqs_common/sqs.go
```diff
@@ -107,6 +113,11 @@ func (s *SQS) DeleteMessages(m []*sqs.Message, lg *log.Logger) error {
 			lg.Error("deleting messages retry failed, objects will likely be duplicated", log.KVErr(err))
 		}
 	}
+
+	if err != nil {
+		err = fmt.Errorf("queue '%s': %w", s.Queue(), err)
```
nic-osullivan-gravwell: same note as in `GetMessages`

nic-osullivan-gravwell (changes requested):
Overall looks good, just a small nit on the error verbiage.

nic-osullivan-gravwell (approved):
lgtm

PR #2377, opened 2026-05-01
Implement playbook API changes

john-floren-gravwell:


client/types/kits/types.go
```diff
@@ -464,3 +464,76 @@ func (pd *PackedDashboard) JSONMetadata() (json.RawMessage, error) {
...
+		AuthorURL:     pb.AuthorURL,
+		Labels:        pb.Labels,
+	}
+	return
+}
+
+// Validate checks the fields of the PackedPlaybook.
+func (pp *PackedPlaybook) Validate() error {
+	if pp.Name == `` {
+		return fmt.Errorf("Missing playbook name")
```
tyler-elton-gravwell:
```suggestion
		return fmt.Errorf("missing playbook name")
```

client/types/kits/types.go
```diff
@@ -464,3 +464,76 @@ func (pd *PackedDashboard) JSONMetadata() (json.RawMessage, error) {
...
+	pb.AuthorName = pp.AuthorName
+	pb.AuthorEmail = pp.AuthorEmail
+	pb.AuthorCompany = pp.AuthorCompany
+	pb.AuthorURL = pp.AuthorURL
+	return
+}
+
+// JSONMetadata returns additional info about the PackedPlaybook in JSON format.
+func (pp *PackedPlaybook) JSONMetadata() (json.RawMessage, error) {
+	b, err := json.Marshal(&struct {
```
tyler-elton-gravwell, line 529: might just be my preference, but i'd read this easier if it were a named/unexported struct instead of inline. nothing structurally wrong though.

client/admin.go
```diff
@@ -830,13 +830,13 @@ func (c *Client) PurgeUser(id int32) error {
...
-	} else if len(pbs) > 0 {
-		for _, pb := range pbs {
-			if pb.UID == id {
-				if err := nc.DeletePlaybook(pb.GUID); err != nil {
-					return fmt.Errorf("Failed to purge user playbook %v %w", pb.GUID, err)
+	} else if len(pbs.Results) > 0 {
+		for _, pb := range pbs.Results {
+			if pb.OwnerID == id {
+				if err := nc.PurgePlaybook(pb.ID); err != nil {
+					return fmt.Errorf("Failed to purge user playbook %v %w", pb.ID, err)
```
tyler-elton-gravwell:
```suggestion
					return fmt.Errorf("purge user playbook %v %w", pb.ID, err)
```

client/playbooks.go
```diff
@@ -9,61 +9,72 @@
...
+	return
+}
+
+// ListAllPlaybooks (admin-only) returns all playbooks on the system.
+func (c *Client) ListAllPlaybooks(opts *types.QueryOptions) (ret types.PlaybookListResponse, err error) {
+	if opts == nil {
+		opts = &types.QueryOptions{}
 	}
-	return pbs, nil
+	opts.AdminMode = true // we'll reject this if the user isn't actually an admin
```
tyler-elton-gravwell, line 31: is this supposed to be here? the comment mentioned rejecting

tyler-elton-gravwell (changes requested):
mostly lgtm minus a few nitpick-y things plus just a question on the admin setting (more for my own learning)

client/playbooks.go
rory-landau-gravwell: Yeah; that is standard practice for the ListAll* calls. ListAll and List are the same, but ListAll force-sets admin mode so the user requests ALL items. The backend rejects the calls if the user is not actually an admin.

rory-landau-gravwell (dismissed):
LGTM

tyler-elton-gravwell (approved):
lgtm

PR #2468, opened 2026-06-15
feat(issues#2365): Include ContentType and FileExtension in Files 

rory-landau-gravwell:
This PR addresses [issues#2365](https://github.com/gravwell/issues/issues/2365)

Sister PR to https://github.com/gravwell/backend/pull/3974

## This PR makes the Resources API conform to that of most other assets

Resource update calls (Metadata or Populates) now return the updated metadata, like most other asset types do. This didn't actually require a change to backend as backend was always shipping over the updated metadata and we just weren't giving it back to the caller.

## This PR enables file extensions to be set into the HTTP request header

Currently, we *always* set 'file' as the file name in the http header for Resources. When backend tried to read the extension from this header, it would always find nothing. Now, both Files and Resources set 'file'+extension. We don't actually care about the name here, only reading the extension off the header. While we could read the extension verbatim from the metadata.... (see next point)

## This PR adds Populate*FromPath functions

These functions are expected to be the primary entry points for populating Files/Resources and set the extension automatically, rather than taking it as a param.

## PR Tasks

  - tests are included in the backend's testing suite, which has been updated to use the new semantics

## Reviewer Tasks

### DIFF
--- a/client/file.go
+++ b/client/file.go
@@ -16,7 +16,7 @@ import (
 	"io"
 	"net/http"
 	"os"
-	"path"
+	"path/filepath"
 	"strconv"
 
 	"github.com/gravwell/gravwell/v4/client/types"
@@ -110,8 +110,8 @@ func (c *Client) PopulateFile(id string, extension string, data []byte) (types.F
 // Extension is taken verbatim from the path.
 //
 // Returns the metadata of the populated/updated file.
-func (c *Client) PopulateFileFromPath(id string, filepath string) (types.File, error) {
-	f, err := os.Open(filepath)
+func (c *Client) PopulateFileFromPath(id string, pth string) (types.File, error) {
+	f, err := os.Open(pth)
 	if err != nil {
 		return types.File{}, err
 	}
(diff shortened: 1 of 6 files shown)


rory-landau-gravwell:
This is R4R, but backend is hard blocked, so I am holding off assigning for review until backend is also ready.

rory-landau-gravwell:
Okay, we should be set. I believe Backend is ready to go as soon as this merges.

client/resources.go
```diff
@@ -86,10 +93,25 @@ func (mpw *mpWriter) Close() (err error) {
...
+// Extension is taken verbatim from the path.
+//
+// Returns the metadata of the populated/updated resource.
+func (c *Client) PopulateResourceFromPath(id string, filepath string) (types.Resource, error) {
+	f, err := os.Open(filepath)
+	if err != nil {
+		return types.Resource{}, err
+	}
+
+	return c.PopulateResourceFromReader(id, path.Ext(filepath), f)
```
john-floren-gravwell: use filepath.Ext here

client/file.go
```diff
@@ -94,20 +96,42 @@ func (c *Client) GetFileMetadata(id string) (types.File, error) {
...
+	}
 
+	// sanity check the length
+	if fi, err := f.Stat(); err != nil {
+		return types.File{}, err
+	} else if uint64(fi.Size()) > maxFileSize {
+		return types.File{}, ErrOversizedFile
+	}
+
+	return c.PopulateFileFromReader(id, path.Ext(filepath), f)
```
john-floren-gravwell: use filepath

john-floren-gravwell (changes requested):

client/file.go
rory-landau-gravwell: Good catches

john-floren-gravwell (approved):

PR #2469, opened 2026-06-16
Registry kits

john-floren-gravwell:
Move kits into the registry.

Big rework to how items get packed into kits, too: everything now has a "Packed" struct counterpart, so Alert -> PackedAlert etc. KitItem now contains the information that once lived in AdditionalInfo, which has thankfully been killed off along with those terrible JSONMetadata functions.


kris-watts-gravwell (approved):
YOLO LGTM

PR #2471, opened 2026-06-18
fix: hosted ingester logs not writing to file

nic-osullivan-gravwell:
This PR addresses no issue. 

## This PR proposes...

When cloning a logger include the writers. Despite the initial comment this doesn't appear to create major blocks. Though there is a slight chance of a race where log data in the log file is mis ordered. In practice I don't think this is going to be an issue. This only occurred ~.5% to 1% of the time. This was also observed during normal ingester startup logs before running the testers. This testing did reveal a race condition in our loggers where we write a line and the newline in separate calls. Adjusting this to write them together resolves this. 

Even with one tester running at a 1ms interval we couldn't really keep up logging to a file at that interval. We were able to get about 166-178 entries written a second. With all the below configs that dropped to 110-124, but the intervals 30ms and above logged about as expected per second. 

image

I used the following config for the concurrency tests:
```
[Global]
	Ingester-UUID="553bb802-1882-4091-b6ff-cffb99afbb73"
	Ingest-Secret = "IngestSecrets"
	Connection-Timeout = 0
	Insecure-Skip-TLS-Verify=false

	Cleartext-Backend-Target=127.0.0.1:4023 #example of adding a cleartext connection
	#Cleartext-Backend-Target=127.1.0.1:4023 #example of adding another cleartext connection
	#Encrypted-Backend-Target=127.1.1.1:4024 #example of adding an encrypted connection

	##a named pipe connection, this should be used when ingester is on the same machine as a backend
	#Pipe-Backend-Target=/opt/gravwell/comms/pipe

	# always enable the cache for hosted ingesters
	Ingest-Cache-Path=/opt/gravwell/cache/hosted_runner.cache
	Max-Ingest-Cache=1024 #Number of MB to store, localcache will only store 1GB before stopping.  This is a safety net

	# log levels and output files
	Log-Level=INFO
	Log-File=/opt/gravwell/log/hosted_runner.log

# configure the state file for hosted ingesters
[State]
	Path="/opt/gravwell/etc/hosted_runner.state"
	Sync=false #do NOT flush to disk on every sync (sync=true mode can be expensive if disks are slow)

[Tester "hosted-100"]
    Ingester-UUID=8d838099-707e-4b6e-9921-8b8d0979f143
    Interval=100ms

[Tester "hosted-10"]
    Ingester-UUID=1777c95e-16f9-43d6-88fd-d4c91b22e3ce
    Interval=10ms

[Tester "hosted-30"]
    Ingester-UUID=c372e628-efe3-442a-a9f0-ef96ce89bb29
    Interval=30ms

[Tester "hosted-40"]
    Ingester-UUID=2488326c-3dee-4f38-b49f-a6bfbc968af7
    Interval=40ms

[Tester "hosted-1"]
    Ingester-UUID=34cfd3ee-c0b3-4718-87d2-ac9baa72f252
    Interval=1ms
```

Finally this also address a bug found with the KVLogger in that it doesn't append the kvs on all log calls. 

## PR Tasks

## Reviewer Tasks

### DIFF
--- a/ingest/log/logging.go
+++ b/ingest/log/logging.go
@@ -169,10 +169,8 @@ func NewDiscardLogger() *Logger {
 }
 
 // Clone creates a new logger based on an existing one, but allows
-// overriding the hostname and appname values
-// embedded WriteClosers are NOT cloned as this will just create a huge race
-// condition were we mux writes in really dumb ways.  This should really only be used in ingesters
-// that are writing up to the host logger.
+// overriding the hostname and appname values. This should really only be used
+// in ingesters that are writing up to the host logger.
 func (l *Logger) Clone(hostname, appname string) (r *Logger, err error) {
 	l.mtx.Lock()
 	defer l.mtx.Unlock()
@@ -539,9 +537,7 @@ func (l *Logger) writeOutput(lvl Level, ts time.Time, ln, msg string) (err error
 	l.mtx.Lock()
 	if err = l.ready(); err == nil {
 		for _, w := range l.wtrs {
-			if _, lerr := io.WriteString(w, ln); lerr != nil {
-				err = lerr
-			} else if _, lerr = io.WriteString(w, "\n"); lerr != nil {
+			if _, lerr := io.WriteString(w, ln+"\n"); lerr != nil {
 				err = lerr
 			}
 		}
--- a/ingest/log/logging.go
+++ b/ingest/log/logging.go
@@ -178,6 +178,7 @@ func (l *Logger) Clone(hostname, appname string) (r *Logger, err error) {
 	defer l.mtx.Unlock()
 	r = &Logger{
 		metadata: l.metadata,
+		wtrs:     make([]io.WriteCloser, len(l.wtrs)),
 		rls:      make([]Relay, len(l.rls)),
 		lrls:     make([]LevelRelay, len(l.lrls)),
 		mtx:      sync.Mutex{},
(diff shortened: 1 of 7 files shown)


ingest/log/logging.go
```diff
@@ -537,9 +537,7 @@ func (l *Logger) writeOutput(lvl Level, ts time.Time, ln, msg string) (err error
 	l.mtx.Lock()
 	if err = l.ready(); err == nil {
 		for _, w := range l.wtrs {
-			if _, lerr := io.WriteString(w, ln); lerr != nil {
-				err = lerr
-			} else if _, lerr = io.WriteString(w, "\n"); lerr != nil {
+			if _, lerr := io.WriteString(w, ln+"\n"); lerr != nil {
```
nic-osullivan-gravwell, line 540: This does technically create a new string here, but I don't think the cost is worth the race condition. Writing twice was _probably_ a micro-optimization.

ingest/log/logging_test.go
```diff
@@ -100,7 +100,7 @@ func TestRawValue(t *testing.T) {
 		t.Fatal(err)
 	}
 	lgr.raw = true
-	testOutputs(t, lgr)
```
nic-osullivan-gravwell, line 103: technically testOutputs never asserted against `testraw.log` and only passed when run with `TestValue`

kris-watts-gravwell (approved):
LGTM

PR #2472, opened 2026-06-18
feat(issues#2528): include bytes param for previewing contents

rory-landau-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/2528.

Sister PR to https://github.com/gravwell/backend/pull/3985.

## This PR adds the previewBytes param

Files and Resources can now take a previewBytes param to limit the number of bytes of content returned.

## This PR adds GetResourcesEx

>[!NOTE]
>I am not married to the API (just adding previewBytes as a param). If we forsee more parameters, we should switch to an Options style (ex: `types.GetResourceOptions` with fields `PreviewBytes` and `IncludeDeleted`/QO). I am open to other ideas.

## PR Tasks

## Reviewer Tasks

### DIFF
--- a/client/resources.go
+++ b/client/resources.go
@@ -12,8 +12,6 @@ import (
 	"bufio"
 	"bytes"
 	"encoding/json"
-	"errors"
-	"fmt"
 	"io"
 	"mime/multipart"
 	"net/http"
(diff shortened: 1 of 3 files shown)


client/resources.go
```diff
@@ -201,13 +202,22 @@ func (c *Client) GetResourceMetadata(id string) (types.Resource, error) {
 // 2. Resources shared with a group to which the user belongs are next
 // 3. Global resources are the lowest priority
 func (c *Client) GetResource(name string) ([]byte, error) {
+	return c.GetResourceEx(name, 0)
+}
+
+// GetResourceEx returns the contents of the resource with the specified name, up to previewBytes (if 0, everything is returned).
+//
+// Follows the name/ID logic of GetResource.
+func (c *Client) GetResourceEx(name string, previewBytes uint64) ([]byte, error) {
```
john-floren-gravwell: Can you add include_deleted here too and implement it in your backend PR? Looks like an oversight.

john-floren-gravwell (changes requested):

client/resources.go
rory-landau-gravwell: I certainly can, but doing so will require a change to lookup as well (unless I'm misreading it and lookup includes deleted by default). The lookup endpoint is not included in the JSON spec and it does not (currently) take QOs. ~~Want me to update lookup's logic as well?~~

john-floren-gravwell (approved):

PR #2481, opened 2026-06-22
Make automation results embed as pointers

john-floren-gravwell:
This makes it clearer if there's any results at all or not.


rory-landau-gravwell (approved):
LGTM

PR #2493, opened 2026-06-29
feat(#2517): Add AI embeddings preprocessor

david-fritz-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/2517

## This PR proposes...

- Adding an AI embeddings preprocessor that attaches an intrinsic EV with the vector. 
- This is intended to be used with an upcoming semantic search module.


ingest/processors/vector.go
```diff
@@ -0,0 +1,281 @@
...
+	VectorProcessor string = `vector`
+
+	defaultVectorTimeout       = 30 // seconds
+	defaultVectorRetryAttempts = 3
+	vectorRetryBaseDelay       = 250 * time.Millisecond
+	vectorRetryMaxDelay        = 5 * time.Second
+)
+
+var (
+	ErrMissingModel    = errors.New("Missing Model in vector config")
```
daniel-moreno-levy-gravwell: In my editor, staticheck gives me the error:

`1. ST1005: error strings should not be capitalized [default]`

I realize this is everywhere in the codebase already, but let's try no to add to it.

ingest/processors/vector.go
```diff
@@ -0,0 +1,281 @@
...
+		return ErrMissingEndpoint
+	}
+	if u, err := url.Parse(c.Endpoint); err != nil || (u.Scheme != "http" && u.Scheme != "https") || u.Host == "" {
+		return ErrInvalidEndpoint
+	}
+	return nil
+}
+
+type VectorProc struct {
+	nocloser
```
daniel-moreno-levy-gravwell, line 81: `nop_closer` is a bit clearer IMO, I had to look up what this was to understand, while `nop_closer` is pretty obvious.

ingest/processors/vector.go
```diff
@@ -0,0 +1,281 @@
...
+	// Collect the non-nil entries and their payloads so the whole batch can be
+	// embedded in a single request.
+	idx := make([]int, 0, len(ents))
+	inputs := make([]string, 0, len(ents))
+	for i, ent := range ents {
+		if ent == nil {
+			continue
+		}
+		idx = append(idx, i)
+		inputs = append(inputs, string(ent.Data))
```
daniel-moreno-levy-gravwell, line 134: What about binary data? I realize it would be stupid of someone to try to get vec embeddings of non printable string data, but It could happen... Is there anything we could do? What does the model do?

ingest/processors/vector.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	rset = ents[:0]
```
daniel-moreno-levy-gravwell, line 137: Could you put a comment explaining this line and the intention behind it? We had a whole discussion on the backend channel for this, which is a good indication it should at least be explained for posterity

ingest/processors/vector.go
```diff
@@ -0,0 +1,281 @@
...
+		return nil, false, ErrEmbeddingCount
+	}
+
+	// The API may return data out of order; place each by its index.
+	embeddings := make([][]float64, want)
+	for _, d := range embResp.Data {
+		if d.Index < 0 || d.Index >= want {
+			return nil, false, ErrEmbeddingCount
+		}
+		if len(d.Embedding) == 0 {
```
daniel-moreno-levy-gravwell, line 265: This is probably the answer to: https://github.com/gravwell/gravwell/pull/2493/changes#r3498987540

However... do we want to stop all embeddings if one of the entries was unprocessable? Shouldn't we try to embed what we can?

ingest/processors/vector_test.go
```diff
@@ -0,0 +1,424 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(rset) != 1 {
+		t.Fatalf("expected success after retries, got %d results", len(rset))
+	}
+	if got := atomic.LoadInt32(&calls); got != 3 {
+		t.Errorf("expected 3 attempts (2 rate-limited + 1 success), got %d", got)
+	}
+}
```
daniel-moreno-levy-gravwell, line 425: I'm missing some test cases here, like the error paths of `doEmbeddingRequest` (empty embeddings / wrong count)

daniel-moreno-levy-gravwell (changes requested):

ingest/processors/vector.go
daniel-moreno-levy-gravwell: Additionally, it would be nice to leave a comment here as to why we these methods are necessary... It took me some digging to see that `Processor` interface requires it.

david-fritz-gravwell: The `nocloser` type comes from the processors definition. Won't fix/change here.

ingest/processors/vector.go
david-fritz-gravwell: Strings only -- and this should be documented for sure.

ingest/processors/vector.go
```diff
@@ -0,0 +1,281 @@
...
+	embeddings, eerr := vp.getEmbeddings(inputs)
+	if eerr != nil {
+		// In passthrough mode we keep going quietly and emit the originals
+		// untouched. Otherwise surface the error so the ingester logs it; the
+		// batch is dropped.
+		if vp.Passthrough_On_Error {
+			return ents, nil
+		}
+		return nil, eerr
+	}
```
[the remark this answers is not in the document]
david-fritz-gravwell: We have to filter nils on success because that's how embeddings work, but nil entries are otherwise valid.

ingest/processors/vector_test.go
david-fritz-gravwell: Done.

ingest/processors/vector.go
david-fritz-gravwell: Done

ingest/processors/vector.go
david-fritz-gravwell: Done.

ingest/processors/vector.go
david-fritz-gravwell: I'm not sure what the right thing to do here is. I don't know if it's possible for a model to respond with "some but not all" embeddings, so for now at least I think this is the behavior I want...

ingest/processors/vector.go
```diff
@@ -0,0 +1,286 @@
...
+	vectorRetryMaxDelay        = 5 * time.Second
+	maxEmbeddingResponseSize   = 1 << 20 // 1 MB — guard against unbounded response bodies
+)
+
+var (
+	ErrMissingModel    = errors.New("missing model in vector config")
+	ErrMissingEndpoint = errors.New("missing endpoint in vector config")
+	ErrInvalidEndpoint = errors.New("invalid endpoint in vector config")
+	ErrEmptyEmbedding  = errors.New("empty embedding response")
+	ErrEmbeddingCount  = errors.New("Embedding response count does not match request")
```
daniel-moreno-levy-gravwell: Initial uppercase error..

daniel-moreno-levy-gravwell:
```suggestion
	ErrEmbeddingCount  = errors.New("embedding response count does not match request")
```

daniel-moreno-levy-gravwell (changes requested):
Nitpik, everything else was either addressed or given a good reason not to. After the small change request, it LGTM.

daniel-moreno-levy-gravwell (approved):

PR #2495, opened 2026-06-30
feat(#2540): Client types for RendererSettings

daniel-moreno-levy-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/2540

> [!IMPORTANT]
> Please review https://github.com/gravwell/backend/pull/4061 along with this PR. They only make sense together. Only this PR will be flipped to IS READY, due to needing to update the replace go.mod statement when this one is merged, but they NEED to be reviewed together. Leave comments in both, dear reviewer.

## This PR proposes...

- Adding the client types for RendererSettings, matching the api-docs schemas in gravwell/api-docs#731, gravwell/api-docs#732, gravwell/api-docs#736, gravwell/api-docs#737 and gravwell/api-docs#740.
- Carrying no JSON name annotations, per https://github.com/gravwell/issues/issues/2585, so these types are PascalCase on the wire from the outset. `omitempty` stays where a channel is optional (`Temporal`, `Min`, `Max`, `Color`), since the settings say "this renderer has no such channel" by omitting it. api-docs sibling: gravwell/api-docs#747.
- Guarantee the required arrays (`tooltip`, `columns`) marshal as `[]` rather than `null`. `encoding/gob` doesn't preserve empty-vs-nil slices, so a search round-tripping through the archive or remote search info would otherwise come back as `"tooltip": null` and break the schema contract.
- Fix TimeFrame struct so it can be validated as a pointer (and update the one call site in `client/searchCtrl.go`).
- Transitive dep bump that came along with `go mod tidy`: google.golang.org/grpc -> 1.83.0

## PR Tasks

- [x] ~~**Bug fixes only:** minimal repro steps included on the issue (for PR QA + Release QA).~~

## Reviewer Tasks


client/types/search.go
```diff
@@ -353,6 +567,7 @@ type SearchCtrlStatus struct {
 type SearchDownloadRequest struct {
 	Format    string         `json:"format"`
 	Rows      []RowSelection `json:"rows,omitempty"`
+	// TODO  This should be a pointer... Fix here?
```
daniel-moreno-levy-gravwell: I should probably take care of this here, document it in this PR, the issue and the backend PR.

client/types/search.go
```diff
@@ -317,12 +319,249 @@ type SearchLaunchInfo struct {
...
+		}
+		rs.Tabular = &tabular
+	case "fdg":
+		var fdg RSFdg
+		if err := json.Unmarshal(data, &fdg); err != nil {
+			return err
+		}
+		rs.Fdg = &fdg
+	default:
+		return fmt.Errorf("unknown renderer %s", temp.Renderer)
```
daniel-moreno-levy-gravwell, line 425: **Unknown `renderer` aborts the entire enclosing decode, not just this field.**

Returning an error here propagates out of everything that carries a `SearchInfo`, so one search using a renderer this build doesn't know takes down the whole payload:

- `SearchInfo()` (`client/searchCtrl.go:57`) and `ListSearchDetails()` (`:124`) → error plus zero/truncated data
- `StartSearchEx` (`:240`) and `AttachSearch` (`:291`) via `LaunchResponse.Info` → cannot launch or attach *any* search
- `BaseResponse.SearchInfo` (`client/types/render.go:247`) → every renderer response frame

Verified against this branch:

```
body = [{"ID":"s1",...},{"ID":"s2","RendererSettings":{"renderer":"sankey","channels":{}}},{"ID":"s3",...}]
err  = unknown renderer sankey     // s1 and s3 are lost as well
```

`{"RendererSettings":{}}` and `{"renderer":""}` fail the same way with the unactionable message `unknown renderer ` (trailing space).

This is the wrong direction for a published SDK — an older client against a newer server has to degrade, not fail. Nothing else in this package hard-fails on an unrecognised value: `SearchStatus`, `SearchState.Status` and `LaunchResponse.RenderModule` are plain strings, and an unknown `LaunchInfo.method` decodes cleanly (verified).

Suggest leaving every variant nil for an unrecognised renderer (optionally keeping the raw bytes in a `json.RawMessage`) rather than erroring.

client/types/search.go
```diff
@@ -317,12 +319,249 @@ type SearchLaunchInfo struct {
...
+			if !field.IsNil() {
+				if active != nil {
+					return nil, fmt.Errorf("multiple render settings options specified in RenderSettings, only one can be active at a time")
+				}
+				active = field.Interface()
+			}
+		}
+	}
+	if active == nil {
+		return nil, fmt.Errorf("no render settings option specified in RenderSettings")
```
daniel-moreno-levy-gravwell: **A non-nil-but-empty `RendererSettings` makes the whole enclosing document unencodable — zero bytes out.**

`json:",omitempty"` on `SearchInfo.RendererSettings` tests the *pointer*, not the pointee, so `&RendererSettings{}` is never skipped and this error escapes to the top of the encode. Verified blast radius:

```
json.Marshal([]SearchInfo{ok, ok, {RendererSettings: &RendererSettings{}}, ok})
  → bytes=0, err=json: error calling MarshalJSON for type types.SearchInfo: ... no render settings option specified
```

That propagates through `LaunchResponse` (whose `Info` is a **value** field, so it cannot even be omitted), `AttachSearchResponse`, and all 11 renderer marshallers that do `json.Marshal(x.BaseResponse)` (`render.go:744`, `maps.go:210/229/248`, `table.go:112/131/150`, `chart.go:259`, `text.go:42`, `stackgraph.go:43`, `fdg.go:52`). In the usual handler shape the 200 status line is already committed, so the client sees HTTP 200 with an empty body and no error signal. The message also names no element index.

And this shape survives the exact round trip this PR is written for — `gob` **preserves** a non-nil pointer to an all-nil struct:

```
gob encode/decode SearchInfo{RendererSettings: &RendererSettings{}}
  → RendererSettings after gob = &{<nil> ... <nil>}   (nil? false)
  → json.Marshal(out) = error "no render settings option specified in RenderSettings"
```

For a display-only hint field this is a very expensive failure mode. Emitting `null` (and letting `omitempty` do its job) degrades gracefully; validate at the boundary that accepts the value instead.

client/types/search.go
```diff
@@ -317,12 +319,249 @@ type SearchLaunchInfo struct {
...
+	return json.Marshal(a)
+}
+
+func (rs *RendererSettings) UnmarshalJSON(data []byte) error {
+	var temp struct {
+		Renderer string `json:"renderer"`
+	}
+	if err := json.Unmarshal(data, &temp); err != nil {
+		return err
+	}
```
daniel-moreno-levy-gravwell: **`UnmarshalJSON` is missing two guards: it never resets the receiver, and it is not a no-op for `null`.**

**1. No receiver reset.** `encoding/json` reuses existing storage, so decoding twice into the same value leaves two live variants — a state `MarshalJSON` then rejects forever:

```
var rs RendererSettings
json.Unmarshal(chartJSON, &rs); json.Unmarshal(fdgJSON, &rs)
  → Chart!=nil=true Fdg!=nil=true
  → json.Marshal(rs) = error "multiple render settings options specified"
```

Confirmed for a reused value, a reused `[]RendererSettings` element, and a `json.Decoder` streaming loop (the normal pattern) — frame 1 encodes fine, frame 2 poisons the value permanently. The diagnostic then points at the *encoder*, far from the cause.

**2. `null` is not a no-op.** The docs say "to approximate the behavior of `Unmarshal` itself, Unmarshalers implement `UnmarshalJSON([]byte(\"null\"))` as a no-op." Here `null` leaves `temp.Renderer == ""` and falls to `default:` → `unknown renderer `. Confirmed for `json.Unmarshal([]byte("null"), &rs)`, a value field from `{"RS":null}`, and `[]RendererSettings` from `[null]`. Only the current `*RendererSettings` field shape escapes it, because `encoding/json` short-circuits `null` for a settable pointer without consulting the Unmarshaler — so this is latent today and fires the first time the type is used in any other shape.

`RowSelection.UnmarshalJSON` in this same file gets the reset right (`*rs = RowSelection(v)`, `:626`).

```suggestion
func (rs *RendererSettings) UnmarshalJSON(data []byte) error {
	if string(data) == `null` {
		return nil
	}
	*rs = RendererSettings{}

	var temp struct {
		Renderer string `json:"renderer"`
	}
	if err := json.Unmarshal(data, &temp); err != nil {
		return err
	}
```

client/types/search.go
```diff
@@ -317,12 +319,249 @@ type SearchLaunchInfo struct {
...
+					return nil, fmt.Errorf("multiple render settings options specified in RenderSettings, only one can be active at a time")
+				}
+				active = field.Interface()
+			}
+		}
+	}
+	if active == nil {
+		return nil, fmt.Errorf("no render settings option specified in RenderSettings")
+	}
+	return json.Marshal(active)
```
daniel-moreno-levy-gravwell, line 353: **The `renderer` discriminator is never set or validated on marshal, so `Marshal` emits documents its own `Unmarshal` rejects — and a wrong value silently changes the variant.**

`MarshalJSON` picks the variant by non-nil pointer and never touches `Renderer`. Two confirmed consequences:

```
json.Marshal(RendererSettings{Chart: &RSChart{Channels: {Category:"c", Nominal:"n"}}})
  → {"renderer":"","channels":{"category":"c","nominal":"n"}}      // marshal succeeds
json.Unmarshal(thatOutput, &rs)
  → err = unknown renderer                                          // its own output is unreadable
```

```
json.Marshal(RendererSettings{Chart: &RSChart{Renderer: "fdg", Channels: {Category:"c", Nominal:"n"}}})
  → {"renderer":"fdg","channels":{"category":"c","nominal":"n"}}
round trip → Fdg set, Chart nil, all three chart channels gone. No error at any layer.
```

The Go type gives a caller no hint that the magic string has to be duplicated by hand, and because `numbercard`/`gauge` collapse to `RSNumber` and `table`/`hex`/`pcap`/`raw`/`text` collapse to `RSTabular`, this unvalidated string is the **only** carrier of variant identity for 7 of the 14 discriminators.

Deriving `renderer` from the active variant in `MarshalJSON` (or verifying it matches) removes the whole class. A typed `RendererKind string` with a declared constant set would make the drift unrepresentable.

client/types/search.go
daniel-moreno-levy-gravwell, line 1: **Pre-existing bug in `SearchInfo.UnmarshalJSON` (`client/types/search.go:773-781`) — `Duration` is always lost on decode.** Flagging it because this PR adds `TestSearchInfoMarshal`, which round-trips a `SearchInfo` and would have caught it.

```go
if len(v.Duration) > 0 {
	dur, err := time.ParseDuration(v.Duration)
	if err != nil {
		return err
	}
	si.Duration = dur      // set here...
}
*si = SearchInfo(v.aalias)     // ...and immediately clobbered
```

`v.aalias.Duration` is always the zero value, because the outer `Duration string` field shadows the embedded `time.Duration` for JSON purposes. Verified on this branch:

```
marshalled: {...,"ItemCount":42,...,"Duration":"1m30s"}
Duration in=1m30s out=0s   |   ItemCount in=42 out=42
```

So every consumer that reads how long a search took — `SearchInfo()`, `ListSearchDetails()`, `AttachSearch`, `BaseResponse.SearchInfo` — gets `0s`. The fix is to swap the two statements (assign `*si` first, then apply the parsed duration).

client/types/search.go
```diff
@@ -317,12 +319,249 @@ type SearchLaunchInfo struct {
...
+
+func (rs *RendererSettings) UnmarshalJSON(data []byte) error {
+	var temp struct {
+		Renderer string `json:"renderer"`
+	}
+	if err := json.Unmarshal(data, &temp); err != nil {
+		return err
+	}
+
+	switch temp.Renderer {
```
daniel-moreno-levy-gravwell, line 369: **Decoding into closed structs makes read-modify-write lossy: unknown channels and options on a *known* renderer are silently dropped.**

Every branch does `json.Unmarshal(data, &variant)` into a concrete struct with no overflow capture, so anything this build doesn't know about is gone on re-encode:

```
in : {"renderer":"chart","channels":{"category":"c","nominal":"n","size":"Bytes"},"stacked":true}
out: {"renderer":"chart","channels":{"category":"c","nominal":"n"}}
```

A client that fetches a `SearchInfo`, changes `Name`, and PUTs it back permanently destroys a `size` channel or a `stacked` option a newer server added. Combined with the hard error on unknown *renderers*, the codec is fragile in both directions at once: it fails loudly on renderers it doesn't recognise and fails silently on fields it doesn't recognise.

Either keep a `json.RawMessage` passthrough per variant, or document `RendererSettings` as read-only for clients and never echoed back.

client/types/search.go
```diff
@@ -353,7 +592,7 @@ type SearchCtrlStatus struct {
 type SearchDownloadRequest struct {
 	Format    string         `json:"format"`
 	Rows      []RowSelection `json:"rows,omitempty"`
-	Timeframe Timeframe      `json:"timeframe,omitempty"`
+	Timeframe *Timeframe     `json:"timeframe,omitempty"`
```
daniel-moreno-levy-gravwell: **Value → pointer keeps compiling for field reads but turns every zero-value read into a nil dereference, and it buys no tri-state.**

Go auto-derefs, so existing consumer code that does `req.Timeframe.Start` / `.End` still builds with zero diagnostics (`go vet` cannot see it) and now panics on any request that omits `timeframe` — which is precisely the case `omitempty` newly enables. Only `IsEmpty()` was nil-hardened; no other access path was. Verified:

```
json.Unmarshal([]byte(`{"format":"json"}`), &sdr)   → ok, sdr.Timeframe == nil
sdr.Timeframe.IsEmpty()                            → true      (safe)
sdr.Timeframe.Start                                → panic: invalid memory address or nil pointer dereference
```

That makes the new guard a trap rather than a fix: `if !sdr.Timeframe.IsEmpty()` is safe, a bare `.Start` crashes, and the two look equally correct at the call site.

Also worth noting the change delivers less than it looks like. `Timeframe`'s own fields carry no `omitempty` (`:658-659`), so `&Timeframe{}` serialises **byte-identically** to the old always-present zero object, and `IsEmpty()` collapses `nil` and `&Timeframe{}` to the same answer. There is no "absent vs explicitly zero" distinction gained — only the nil hazard.

This is also a source break for construction (`Timeframe: types.Timeframe{...}` no longer compiles) on a published SDK type, and the doc comments at `client/staticActions.go:315` and `:333` still describe it as "an optional [types.Timeframe]" with no note of the change. `client/types/render.go:838-856` shows the in-package idiom for omitting a zero struct without touching the public type (`tsPointer` + shadow struct).

client/types/search.go
```diff
@@ -420,7 +659,11 @@ type Timeframe struct {
...
 }
 
-func (tf Timeframe) IsEmpty() bool {
+func (tf *Timeframe) IsEmpty() bool {
+	if tf == nil {
+		return true
+	}
+
 	return tf.Start.IsZero() && tf.End.IsZero()
 }
```
daniel-moreno-levy-gravwell, line 641: **Switching to a pointer receiver removes `IsEmpty` from `Timeframe`'s value method set — a source break on a published SDK type — and may defeat its own purpose.**

```
$ go vet   # on a probe calling IsEmpty on a non-addressable Timeframe
vet: cannot call pointer method IsEmpty on Timeframe
```

So `types.Timeframe{}.IsEmpty()`, `m[k].IsEmpty()` and `f().IsEmpty()` all stop compiling for anyone on `github.com/gravwell/gravwell/v4/client/types`. All four forms compile against `origin/next-major`. There are **zero** callers of `Timeframe.IsEmpty` anywhere in this repo, so CI cannot catch the break — only downstream builds will.

The stated goal was to let it "be validated as a pointer", but the change makes a `Timeframe` **value** stop satisfying `interface{ IsEmpty() bool }`, while its two siblings still satisfy it:

```
Timeframe  (value) → false      *Timeframe → true
TimeRange  (value) → true       (render.go:542, value receiver)
PointmapKV (value) → true       (maps.go:88,  value receiver)
```

A validator that type-asserts on values — the common reflective pattern — therefore silently stops matching `Timeframe` while continuing to match the other two. Worth confirming the validator actually addresses the pointer (`reflect.Value.Addr`, or asserting on both `T` and `*T`) before mutating a published type's method set for it.

Separately: folding `nil` into `IsEmpty` erases the only distinction the `*Timeframe` field was introduced to express — server code can no longer tell "client sent no timeframe" from "client sent an all-zero timeframe". A nil-tolerant free function would give you the nil check without either cost.

client/types/search_test.go
```diff
@@ -80,3 +80,242 @@ func TestRowSelectionUnmarshal(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("Marshal got failed: %v", err)
+			}
+			b2, err := json.Marshal(tt.want)
+			if err != nil {
+				t.Fatalf("Marshal want failed: %v", err)
+			}
+			if string(b1) != string(b2) {
+				t.Errorf("Mismatch:\nGot:  %s\nWant: %s", string(b1), string(b2))
+			}
```
daniel-moreno-levy-gravwell, line 313: **These assertions cannot observe the fix this PR exists to make. Proven: deleting all three `*Channels.MarshalJSON` methods leaves the entire suite green.**

Run in an isolated worktree at `53c0197` with `RSP2PChannels.MarshalJSON`, `RSPointmapChannels.MarshalJSON` and `RSTabularChannels.MarshalJSON` removed:

```
$ go test ./client/types/
ok  github.com/gravwell/gravwell/v4/client/types  0.199s     # all 10 subtests PASS
```

The fix demonstrably *does* change the wire format, so the blindness is in the oracle, not the fix:

```
fix removed: {"renderer":"table","channels":{"columns":null}}   ..."tooltip":null}
fix present: {"renderer":"table","channels":{"columns":[]}}     ..."tooltip":[]}
this test reports equal=true in BOTH builds
```

Root cause: `b1` and `b2` both pass through the same normalizing `MarshalJSON`, so anything the marshaller normalizes is invisible by construction. The `point2point` case feeds JSON with no `tooltip`, `want.Tooltip` is nil, and both sides render `[]`.

The same blindness hides variant flips, because `MarshalJSON` emits only the active variant:

```
RendererSettings{Chart: &RSChart{Renderer: "stackgraph", Channels: {Category:"c", Nominal:"n"}}}
  → round-trips into back.StackGraph (Chart nil) and re-marshals byte-identically
  → the assertion here AND the roundtrip assertion below both pass
```

Nothing in the file asserts a single decoded Go field (no `reflect.DeepEqual`; imports are only `encoding/json` and `testing`), and `b3` at `:246` is a byte-identical recomputation of `b2`.

Suggest a `wantJSON string` column asserted against `string(b1)` for at least the nil-`Tooltip` and nil-`Columns` cases, plus `reflect.DeepEqual(got, tt.want)` for the decode direction.

client/types/search.go
```diff
@@ -317,12 +319,249 @@ type SearchLaunchInfo struct {
...
+	Number     *RSNumber
+	Heatmap    *RSHeatmap
+	Pointmap   *RSPointmap
+	StackGraph *RSStackGraph
+	WordCloud  *RSWordCloud
+	Tabular    *RSTabular
+	Fdg        *RSFdg
+}
+
+func (rs RendererSettings) MarshalJSON() ([]byte, error) {
```
daniel-moreno-levy-gravwell, line 336: **The union validates only the *number* of active variants, never the required channels — so structurally meaningless settings round-trip silently.**

Every required channel field was deliberately given no `omitempty` (`location`, `magnitude`, `from`, `to`, `category`, `nominal`, `weight`, `label`, `value`, `name`), but nothing enforces them, so "required" degrades to "always present, possibly empty":

```
in : {"renderer":"heatmap","channels":{}}          → decodes with no error
out: {"renderer":"heatmap","channels":{"location":"","magnitude":""}}

{"renderer":"chart"}      → {"renderer":"chart","channels":{"category":"","nominal":""}}
{"renderer":"gauge",...}  → {"renderer":"gauge","channels":{"label":"","value":""}}   // min/max are omitempty, so a gauge can ship with no range while a table can never ship without columns
```

A GUI receiving that has a schema-valid object naming no enumerated values: it renders nothing and has no error to report. Symmetrically, `&RendererSettings{Heatmap: &RSHeatmap{Renderer: "heatmap"}}` built in Go marshals cleanly and PUTs a broken saved search.

`RowSelection` 275 lines below is the in-file precedent that gets this right — one `validate()` called from **both** `MarshalJSON` (`:612`) and `UnmarshalJSON` (`:623`), so its accepted set is identical in both directions:

```
RowSelection      marshal err=unknown row selection kind: "bogus"   unmarshal err=unknown row selection kind: "bogus"
RendererSettings  marshal err=<nil>                                 unmarshal err=unknown renderer bogus
```

A single symmetric `validate()` here is also the one change that would fix the empty-discriminator, mismatched-discriminator and two-live-variants problems at their shared root.

client/types/search.go
```diff
@@ -317,12 +319,249 @@ type SearchLaunchInfo struct {
...
+			return err
+		}
+		rs.StackGraph = &stackgraph
+	case "wordcloud":
+		var wordcloud RSWordCloud
+		if err := json.Unmarshal(data, &wordcloud); err != nil {
+			return err
+		}
+		rs.WordCloud = &wordcloud
+	case "table", "hex", "pcap", "raw", "text":
```
daniel-moreno-levy-gravwell: **Folding `hex`/`pcap`/`raw`/`text` into `RSTabular` contradicts how the rest of the client models these renderers, and forces a meaningless `"columns":[]` onto all four.**

The repo keeps these as two disjoint families:

- `table` → `TableValueSet{Columns []string; Rows TableRowSet}` (`client/types/table.go:22-25`) — the only type in the package with a `Columns` field
- `raw`/`hex`/`pcap`/`text` → `TextResponse`/`RawResponse` with `Entries []SearchEntry` and no columns at all (`client/types/text.go:20-32`)

And routes them separately: `GetEntries` sends `pcap`/`raw`/`hex`/`text` to `getStringTagTextEntries` and only `table` to `getStringTagTableEntries` (`client/searchCtrl.go:349-360`); `checkRender` (`:484-509`) allows `hex`↔`text` and `text`/`pcap`↔`raw` substitutions but never anything↔`table`.

With the new `RSTabularChannels.MarshalJSON` the four non-table renderers now always emit a required, permanently-empty column list:

```
hex   → {"renderer":"hex","channels":{"columns":[]}}
pcap  → {"renderer":"pcap","channels":{"columns":[]}}
raw   → {"renderer":"raw","channels":{"columns":[]}}
text  → {"renderer":"text","channels":{"columns":[]}}
```

So a client switching on the union tag (`if rs.Tabular != nil { renderColumns(...) }`) will try to render a column table for a `raw` search, or has to re-parse `rs.Tabular.Renderer` to recover the distinction the union was supposed to encode. A strict schema consumer is also told `columns` is required-and-always-present for renderers whose payload has no columns.

If the api-docs schema really does group all five, a comment saying so would help; otherwise these want separate variants.

client/types/search.go
```diff
@@ -317,12 +319,249 @@ type SearchLaunchInfo struct {
...
+	for _, field := range val.Fields() {
+		if field.Kind() == reflect.Pointer {
+			if !field.IsNil() {
+				if active != nil {
+					return nil, fmt.Errorf("multiple render settings options specified in RenderSettings, only one can be active at a time")
+				}
+				active = field.Interface()
+			}
+		}
+	}
```
daniel-moreno-levy-gravwell, line 348: **The reflection loop discards the guard it needs, and its pointer filter fails open rather than closed.**

`reflect.Value.Fields()` yields `iter.Seq2[StructField, Value]`; `for _, field := range` throws away the `StructField` whose `IsExported()` is exactly the guard for the next line. `field.Interface()` **panics** on an unexported field — it does not return an error:

```
// with an unexported pointer field added to RendererSettings
recovered PANIC: reflect.Value.Interface: cannot return value obtained from unexported field or method
```

Inside a JSON encoder in an HTTP handler that kills the goroutine and truncates the response. Two more ways the loop fails open:

- A future variant declared by value (`Sankey RSSankey` rather than `*RSSankey`) is silently skipped, and marshalling it returns `no render settings option specified` — an error that actively misdirects, since a setting *is* set and its data is dropped.
- Any non-variant pointer field (a `*string` label, a shared `*Options`) becomes a phantom variant and breaks marshalling with `multiple render settings options`.

The `field.Kind() == reflect.Pointer` check is also dead today — all nine fields are pointers.

The comment says reflection is here "to accommodate future renderers", but `UnmarshalJSON` still needs a hand-written `case` and struct per renderer, so nothing is saved — and the asymmetry the comment claims to prevent is what it produces: the marshal side stays silent when the decode side was forgotten. This package already has the non-reflective idiom for exactly this union shape at `client/types/render.go:858-867`:

```go
func (rr ResultsResponse) MarshalJSON() ([]byte, error) {
	if rr.Table != nil {
		return json.Marshal(rr.Table)
	}
	if rr.Graph != nil {
		return json.Marshal(rr.Graph)
	}
	return nil, ErrResultsResponseInvalid
}
```

Note it also returns a package-level sentinel (`render.go:118`) instead of an inline `fmt.Errorf`, so callers can `errors.Is` it — with the current strings nobody can tell "none set" from "multiple set" without matching text. Measured A/B through `json.Marshal`: reflection `963 ns / 745 B / 12 allocs` vs switch `748 ns / 496 B / 5 allocs`.

(Minor, same lines: both messages name `RenderSettings`, which is not a declared type — it's `RendererSettings`.)

client/types/search.go
```diff
@@ -317,12 +319,249 @@ type SearchLaunchInfo struct {
...
+// For table, Columns must always be defined as well
+
+func (r RSTabularChannels) MarshalJSON() ([]byte, error) {
+	type alias RSTabularChannels
+	a := alias(r)
+	if a.Columns == nil {
+		a.Columns = []string{}
+	}
+	return json.Marshal(a)
+}
```
daniel-moreno-levy-gravwell: **These three methods re-implement `emptyStrings`, which this package already has for exactly this job — and they're ~3x more expensive.**

`client/types/api.go:354`:

```go
type emptyStrings []string

func (es emptyStrings) MarshalJSON() ([]byte, error) {
	if len(es) == 0 {
		return emptyList, nil
	}
	return json.Marshal([]string(es))
}
```

It's used at `render.go:583`, `render.go:608`, `chart.go:229`, `cbac.go:136/816`, `api.go:424` — and the canonical example nil-guards a field literally named `Columns`, via the alias + shadow-struct pattern (`client/types/table.go:90-99`):

```go
func (t TableValueSet) MarshalJSON() ([]byte, error) {
	type alias TableValueSet
	return json.Marshal(&struct {
		alias
		Columns emptyStrings
	}{alias: alias(t), Columns: emptyStrings(t.Columns)})
}
```

Output is byte-identical and it is measurably cheaper — nil-tooltip case `110 ns / 64 B / 1 alloc` vs `304 ns / 208 B / 3 allocs` here, because the current form copies the whole struct and runs a nested `json.Marshal`. Declaring the three fields as `emptyStrings` deletes all 28 lines; `[]string{...}` literals still assign to them, and decode is unaffected since `emptyStrings` has no `UnmarshalJSON`.

**Altitude, and the reason this matters beyond tidiness:** the invariant currently lives in prose plus three methods a future author has to remember to write. It happens to be complete today — reflection over all nine `RS*Channels` structs finds exactly three slice fields and all three are patched — but it's unenforced. Add a required array to `RSChartChannels` and it emits `null`: no method, no failure, no test. Typing the field moves the invariant into the type so it cannot be forgotten.

Also worth noting the patch is at the wrong layer for the stated cause. `gob` loses the empty-vs-nil distinction *on decode*, and this only repairs the JSON output — the in-memory struct still holds `nil` (verified), so non-JSON consumers of a gob-decoded value stay broken. Normalising once after gob decode would fix the layer where the information is actually lost.

client/types/search.go
```diff
@@ -317,12 +319,249 @@ type SearchLaunchInfo struct {
...
+func (rs *RendererSettings) UnmarshalJSON(data []byte) error {
+	var temp struct {
+		Renderer string `json:"renderer"`
+	}
+	if err := json.Unmarshal(data, &temp); err != nil {
+		return err
+	}
+
+	switch temp.Renderer {
+	case "chart":
```
daniel-moreno-levy-gravwell: **All 14 discriminator strings are hardcoded here, though this package already exports a constant for every one of them.**

`client/types/render.go:84-97` — a perfect 1:1 overlap with this switch, referenced zero times by it:

```go
RenderNameRaw        string = `raw`        RenderNameChart      string = `chart`
RenderNameHex        string = `hex`        RenderNameFdg        string = `fdg`
RenderNameText       string = `text`       RenderNameStackGraph string = `stackgraph`
RenderNamePcap       string = `pcap`       RenderNamePointmap   string = `pointmap`
RenderNameTable      string = `table`      RenderNameHeatmap    string = `heatmap`
RenderNameGauge      string = `gauge`      RenderNameP2P        string = `point2point`
RenderNameNumbercard string = `numbercard` RenderNameWordcloud  string = `wordcloud`
```

These are already consumed as constants elsewhere in the client — `client/searchCtrl.go:349-361` and `checkRender` at `:484-509`. Using them here matters for more than style: a rename in `render.go` leaves this switch compiling cleanly while silently no longer matching, and `grep RenderNameNumbercard` won't find the decoder. The `// Value "chart"` comments on the variant structs are a third hand-maintained copy of the same strings.

Compounding it, `hex`, `pcap`, `raw` and `text` have **no test case** (10 of the 14 discriminators are covered), and the `default:` branch is never exercised — so nothing at all pins those four strings. They all route to the already-covered `RSTabular`, so it's four table rows.

Also untested: the new `nil`-receiver branch of `Timeframe.IsEmpty()` (never called from any test, before or after this PR) and both halves of the `SearchDownloadRequest.Timeframe` change.

client/searchCtrl.go
```diff
@@ -1065,7 +1065,7 @@ func (c *Client) DownloadSearch(ctx context.Context, sid string, tr types.TimeRa
 	var sdr types.SearchDownloadResponse
 	sdr, err = c.SearchDownloadRequestWithContext(ctx, sid, types.SearchDownloadRequest{
 		Format: format,
-		Timeframe: types.Timeframe{
+		Timeframe: &types.Timeframe{
```
daniel-moreno-levy-gravwell: **The only producer in the repo always allocates a non-nil `Timeframe`, so the nil path this PR just introduced is unreachable from the SDK — and the nil-safe `IsEmpty()` added alongside it goes unused.**

`entry.Timestamp.StandardTime()` returns `time.Time{}` for a zero timestamp (`ingest/entry/time.go:83`), so a zero `TimeRange` still ships a year-1 window rather than omitting the field:

```
tr = types.TimeRange{}   →   tr.IsEmpty() == true
wire = {"format":"json","timeframe":{"end":"0001-01-01T00:00:00Z","start":"0001-01-01T00:00:00Z"}}
```

A zero `TimeRange` is an established in-repo idiom for "download everything": `gwcli/utilities/querysupport/querysupport.go:251` and `gwcli/script_test.go:551` both pass `types.TimeRange{}`. So a server updated in tandem to branch on `Timeframe == nil` rather than `Timeframe.IsEmpty()` reads this as a bounded download at year 1 and returns nothing.

The two halves of the PR disagree about how "no timeframe" is spelled: the struct now has two encodings for it, and this — the sole construction site — only ever emits the non-nil one. Guarding the field with the `IsEmpty()` this PR just made nil-safe is the missing piece:

```go
req := types.SearchDownloadRequest{Format: format}
if !tr.IsEmpty() {
	req.Timeframe = &types.Timeframe{
		Start: tr.StartTS.StandardTime(),
		End:   tr.EndTS.StandardTime(),
	}
}
sdr, err = c.SearchDownloadRequestWithContext(ctx, sid, req)
```

client/types/search.go
daniel-moreno-levy-gravwell: Not changing — this is a spec-defined `oneOf`, not an open enum. And degrading silently is worse than failing: unknown renderer → empty struct → re-encodes to `null`, so a fetch/modify/PUT would quietly wipe the settings. The error prevents that write.

client/types/search.go
daniel-moreno-levy-gravwell: Fixed — empty now marshals to `null`. Kept the multiple-variants error; it is programmer error and unreachable from decode now that `UnmarshalJSON` resets the receiver.

client/types/search.go
daniel-moreno-levy-gravwell: Fixed — both guards added, with tests for the reused-value, reused-slice-element and `json.Decoder` cases.

client/types/search.go
daniel-moreno-levy-gravwell: Not changing — matches the merged api-docs grouping, and changing it means another spec round. Nit: `ResultsTable` (`render.go:191`) and `dashboard.go:150` also have `Columns` fields.

client/types/search.go
daniel-moreno-levy-gravwell: Taking the `emptyStrings` half — filing it. Not the "wrong layer" half: nil and empty slices are identical for `len`/`range`/`append`, and the contract here is the JSON schema.

client/types/search.go
daniel-moreno-levy-gravwell: Real break — you are right, and I was wrong in another thread claiming the type was new here. It is on `next-major` with a value receiver.

Accepting it anyway: next-major branch, internal client, zero callers of `IsEmpty`, and every construction site already uses the pointer form. But "no tri-state gained" is wrong — nil now omits the field, which the value shape could not. Filing the `.Start` nil-panic and the missing tests.

client/types/search.go
daniel-moreno-levy-gravwell: Same — real break, accepted for the same reasons. The `TimeRange`/`PointmapKV` asymmetry is the part worth revisiting if a reflective validator ever depends on it.

client/types/search_test.go
daniel-moreno-levy-gravwell: Confirmed — deleted the three methods and the suite stayed green. Rewritten: literal `wantJSON` per case, `reflect.DeepEqual` on the decoded value, and a dedicated required-arrays test verified by deleting each method individually.

client/types/search.go
daniel-moreno-levy-gravwell: Done — this was the `Timeframe` should-be-a-pointer TODO, and that change landed in this PR. Resolving.

client/types/search.go
daniel-moreno-levy-gravwell: Not changing. This needs someone hand-constructing a `RendererSettings` with a discriminator that disagrees with the variant, and the only producer is `buildRenderSettings` on the backend, which sets `Renderer` from the same switch variable it picks the variant with — they cannot diverge. On decode the string chooses the variant, so they always agree there too. Settings are built once at launch and immutable afterwards.

client/types/search.go
daniel-moreno-levy-gravwell: Same reasoning as the discriminator thread — not changing. Required channels are always filled by the single backend producer, and the settings are immutable after launch, so a `validate()` would guard an unreachable state. I drafted it and its first effect was to reject a legitimate existing test fixture, which is a fair signal that it is protecting nothing real here.

client/types/search.go
daniel-moreno-levy-gravwell: Not changing — there is no read-modify-write path. `SearchInfo` is read-only in the client (`SearchInfo()`, `ListSearchDetails()`, and the launch/attach responses); nothing sends it back to the server, and renderer settings are generated once at launch and immutable. Unknown channels being dropped on decode therefore cannot round-trip back and destroy anything.

client/types/search.go
daniel-moreno-levy-gravwell: Leaving the reflection loop. Every failure mode here is conditional on a future change — an unexported field, a value-typed variant, a non-variant pointer — and today the struct is nine exported pointers. Fair catch that the `Kind() == reflect.Pointer` check is dead; filing that along with the `emptyStrings` simplification rather than reworking the marshaller in this PR.

client/types/search.go
daniel-moreno-levy-gravwell: Correction to my own reply above: I cited a fetch/modify/PUT path as the risk, and that path does not exist — `SearchInfo` is read-only in the client and renderer settings are built once at launch. The decision stands on the simpler ground: this is a closed, spec-defined union rather than an open enum.

client/types/search.go
daniel-moreno-levy-gravwell: Fixed — `*si = SearchInfo(v.aalias)` now runs before the duration is applied, so it is no longer clobbered. Added `TestSearchInfoDurationRoundTrip`, verified by reverting the fix and watching it fail.

client/types/search.go
daniel-moreno-levy-gravwell: Fixed — all 14 cases now use the `RenderName*` constants from `render.go`. Also added the missing `hex`/`pcap`/`raw`/`text` coverage and tests for the `Timeframe` change.

client/searchCtrl.go
daniel-moreno-levy-gravwell: Fixed — `DownloadSearch` now only sets `Timeframe` when `tr` is non-empty, so a zero `TimeRange` omits the field instead of sending a year-1 window. Added `TestDownloadSearchOmitsZeroTimeframe`. Good catch that the nil path was unreachable while the nil-safe `IsEmpty()` sat unused.

david-fritz-gravwell (approved):
LGTM

PR #2496, opened 2026-06-30
kits: Add LicenseText field to KitItem struct & other tweaks

john-floren-gravwell:
This lets us embed the text of licenses directly into the KitItem,
making it easier to review later *or* rebuild a kit from an
already-deployed kit.

Also cleans up some junk from KitBuildRequest and KitMetadata that no longer apply.

Addresses gravwell/issues#2555


rory-landau-gravwell (approved):
LGTM

PR #2479, opened 2026-06-18
fix: log http debug info even in error cases

nic-osullivan-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/2525

## This PR proposes...

Refactoring our `Debug-Posts` handling into defered functions so they are run in more cases. This info is helpful even in error scenarios. 

Given the scope of these changes I also added a number of e2e tests across the http ingester. All listeners but otel metrics have improved coverage at least to ensure something is ingested. HEC got much more coverage as the focus of a lot of our efforts and debugging. 

## Repro

Using any stanza with a `Debug-Posts=true` and running with `-v` you should see a lot more output. I provided samples of the expected output in the linked issue. 

## PR Tasks

- [x] e2e and/or unit tests included. If not, please provide an explanation.
- [x] **Bug fixes only:** minimal repro steps included on the issue (for PR QA + Release QA).

## Reviewer Tasks

- [ ] e2e or unit tests are present to test the proposed changes.
- [ ] Code is sufficiently documented.
- [ ] Code meets quality and correctness expectations.

PR #2497, opened 2026-06-30
fix(2525): allow targeted http listener debugging

nic-osullivan-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/2525

## This PR proposes...

This build on https://github.com/gravwell/gravwell/pull/2479 by adjusting where logging actually happens and making the new logging only occur for errors. This should make it easier to troubleshoot when there are a large number of listeners by only enabling debug logging on one at a time. This is done via a new flag `-debug=[listener name]` matching the name from the config stanza.

This also adds a number of tests around authed http listeners as this got a little more invasive on the changes to individual listeners. 

A few places have been updated to explicitly write a `200` as relying on the default behavior means we don't see the code in the debug logs. 

## PR Tasks

## Reviewer Tasks


tyler-elton-gravwell (approved):
👍

PR #2498, opened 2026-06-30
fix(2553): allow hosted runner config validation when db is locked

nic-osullivan-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/2553

## This PR proposes...

Allow the config validation to pass for the Hosted Runner when another instance is running. This required a little extra work as bolt hides this specific scenario from us behind a vague timeout error.

## PR Tasks

### Repro
- Start an instance of the runner, eg: `go run ./hosted/runner -config-file ./hosted/plugins/tester/example.conf -v`
- Run the validate `go run ./hosted/runner -config-file ./hosted/plugins/tester/example.conf -validate`
- config should be valid. 

## Reviewer Tasks

### DIFF
--- a/hosted/storage/bolt.go
+++ b/hosted/storage/bolt.go
@@ -12,9 +12,14 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"syscall"
 	"time"
 
 	bolt "go.etcd.io/bbolt"
+)
+
+var (
+	ErrFileLocked = errors.New("file is locked")
 )
 
 type BoltConfig struct {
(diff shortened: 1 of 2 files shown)


hosted/storage/bolt.go
```diff
@@ -35,16 +40,26 @@ func (s *BoltConfig) Verify() (err error) {
...
 	}
 
-	//go attempt to open and close the state file
+	// go attempt to open and close the state file
 	var sh *BoltHandler
 	if sh, err = OpenBoltHandler(s.Path, s.Sync); err != nil {
-		err = fmt.Errorf("failed to open state file %w", err)
+		// If the file is locked that likely means we're already running.
+		// The config is most likely pointing to a valid db so config is valid.
+		if errors.Is(err, ErrFileLocked) {
```
daniel-moreno-levy-gravwell, line 48: This fix doesn't sit well with me. The biggest issue for me is that, although unlikely, it is susceptible to race conditions, seemingly open to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) attacks. I also don't like making assumptions that are "almost always" correct, leaving room up for unforeseen errors or bad actors.

I would like to see some deterministic mechanism for which:

- New process can request to take hold of the DB handle, and confirm old process has a handle on it.
- Old process let's go of the handle gracefully, and let's new process takes hold of it.

I have some ideas bouncing around, but I also want to be pragmatic and not over engineer... Let's discuss them and if nothing is convincing, or I am exaggerating the risks, then we can move on with this as is.

daniel-moreno-levy-gravwell (changes requested):

hosted/storage/bolt.go
daniel-moreno-levy-gravwell: To play devil's advocate... I missed that this code path is only triggered for the `-validate` flag... Not sure if that is how we default the deployments of this package. If it isn't, then maybe this scenario is too niche to even bother.

daniel-moreno-levy-gravwell: This idea is pretty simple actually:

- The runner writes a PID file when it starts. `-validate` checks if the PID file exists and the PID is alive and only then skips the DB open. This ties the lock to a specific process identity.

nic-osullivan-gravwell: Talked offline. Since `-validate` only checks the configs and exist, it does not start up the hosted runner we don't need anything more than this. A PID check might be worth it in the future, but for now this is good.

hosted/storage/bolt.go
```diff
@@ -68,6 +88,27 @@ func OpenBoltHandler(pth string, sync bool) (sh *BoltHandler, err error) {
...
+		return
+	}
+	defer file.Close()
+	fd := file.Fd()
+	if err = syscall.Flock(int(fd), syscall.LOCK_NB|syscall.LOCK_SH); err != nil {
+		if errors.Is(err, syscall.EWOULDBLOCK) {
+			return true, nil
+		}
+		return false, fmt.Errorf("lock check failed: %w", err)
+	}
```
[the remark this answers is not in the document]
nic-osullivan-gravwell: This was a good catch from how the lib acquires the lock.

daniel-moreno-levy-gravwell (approved):
LGTM!

PR #2499, opened 2026-06-30
Clean up some kit types & return values

john-floren-gravwell:
KitBuildResponse fields changed.

Listing remote kits now returns a ListResponse like many other list endpoints.


rory-landau-gravwell (approved):
LGTM

PR #2504, opened 2026-07-01
update fsnotify package

david-fritz-gravwell:
Fixes https://github.com/gravwell/issues/issues/2560 via (at least) https://github.com/fsnotify/fsnotify/pull/740

## This PR proposes...

- Bumping the release of fsnotify to 1.10.1 

## Reviewer Tasks

- This is a race in the 1.7.0 release of fsnotify. Easily triggered with something like:

```
touch auth.log; sleep 1; rm auth.log; sleep 1; lsof -nP | grep auth.log
```

When watching for "auth.log", the FD will often be leaked (you can issue this command several times and see it leak more and more FDs.

- Please test that the FDs are no longer leaked using the same approach.


kris-watts-gravwell (approved):
LGTM

PR #2511, opened 2026-07-02
fix(issues#2483): wrap bad status codes with equivalent errors

rory-landau-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/2483.

Sister PR to https://github.com/gravwell/backend/pull/4060.

## This PR adds proper error aliases to bad status code

Requests that result in unexpected, non-200 status codes now return the correspondong error, instead of a string error (or *ClientError with the error as a string in the body). This should make errors.Is against our custom errors much more reliable.

This is implemented somewhat unevenly due to the volume of different helper functions for issuing requests. The most popular helpers (like `Client.staticRequest()` and `Client.methodStaticPushRawURL()`), those that return only an error,  automatically check the status response code for the caller. Other helpers, those that return the response and an error, expect the caller to do. This can be tidied up again in the future, as we trim out the fat.

## PR Tasks

  - Backend tests have been added in the sister PR.

## Reviewer Tasks

### DIFF
--- a/client/staticActions.go
+++ b/client/staticActions.go
@@ -155,18 +155,9 @@ func (c *Client) staticRequest(req *http.Request, obj interface{}, okResponses [
 		return errors.New("Invalid response")
 	}
 	defer drainResponse(resp)
-	if resp.StatusCode == http.StatusUnauthorized {
-		c.state = STATE_LOGGED_OFF
-		return ErrNotAuthed
-	} else if resp.StatusCode == http.StatusNotFound {
-		return ErrNotFound
-	}
-
-	statOk := respOk(resp.StatusCode, okResponses...)
-	//either its in the list, or the list is empty and StatusOK is implied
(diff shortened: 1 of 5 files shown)


client/staticActions.go
```diff
@@ -431,7 +414,9 @@ func (c *Client) methodRequestURL(method, url, contentType string, body io.Reade
 	return
 }
 
-func (c *Client) methodParamRequestURL(method, uri string, params map[string]string, body io.Writer) (resp *http.Response, err error) {
+// methodParamRequestURL builds and submits a request against the specified uri.
+// Returns an error iff the request failed. You must check the resp's status code yourself.
+func (c *Client) methodParamRequestURL(method, uri string, params map[string]string) (resp *http.Response, err error) {
```
rory-landau-gravwell, line 419: This change just removes an unused parameter.

john-floren-gravwell (approved):

PR #2513, opened 2026-07-06
feat(2578): set a default ingester uuid in hosted runner plugins

tyler-elton-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/2578

## This PR proposes...

- updating all hosted runner plugins to have a default ingester uuid that will be used if one isn't explicitly provided in the config

## PR Tasks

## Reviewer Tasks

image


hosted/config.go
```diff
@@ -28,6 +33,20 @@ func (b *BaseConfig) UUID() uuid.UUID {
 	return ParseUUID(b.Ingester_UUID)
 }
 
+// ApplyDefaultIngesterUUID applies a default ingester UUID if none is provided.
+func (b *BaseConfig) ApplyDefaultIngesterUUID(defaultIngesterUUID string) {
```
nic-osullivan-gravwell, line 37: Should this just be `ApplyDefaults(BaseConfig)` to keep it more flexible if we add more "base" options? Passing in the entire base config seems like overkill, but more flexible to adding things. This might be yagni though on my part. Though this does pull it closer to an interface we are impliciting creating for these embedded configs.

hosted/config.go
```diff
@@ -28,6 +33,20 @@ func (b *BaseConfig) UUID() uuid.UUID {
...
+	b.Ingester_UUID = cmp.Or(b.Ingester_UUID, defaultIngesterUUID)
+}
+
+// VerifyIngesterUUID validates the Ingester-UUID field is a valid UUID.
+// Returns ErrInvalidIngesterUUID if we could not parse Ingester-UUID.
+func (b *BaseConfig) VerifyIngesterUUID() error {
```
nic-osullivan-gravwell: I think this should be just `Verify()` and called like `c.BaseConfig.Verify()` again to handle more than just uuid. This also keeps it closer to the implicit interface we have for ingester validation.

nic-osullivan-gravwell (changes requested):

nic-osullivan-gravwell:
Oh can we also get an e2e test to cover this? I've actually been bit by this pretty much every time I add a stanza to the hosted e2e tests. Nothing fancy just a test stanza without a uuid and verify it ingested something.

hosted/config.go
tyler-elton-gravwell: To me, this feels like yagni. If `BaseConfig` ever gets a second field that gets a default value we can refactor it then.

hosted/config.go
tyler-elton-gravwell: addressed in [e34a3ab](https://github.com/gravwell/gravwell/pull/2513/commits/e34a3abd6ad367e07bd31d3ba6221d76479e60df)

nic-osullivan-gravwell (approved):
lgtm

PR #2514, opened 2026-07-06
Adjustments to Tokens APIs

john-floren-gravwell:
CreateToken now takes a types.Token and UpdateToken takes just types.Token, no separate ID argument. This brings the API in line with all our other assets.

Also removes JSON lowercasing on struct fields, because we want that in general.

Addresses gravwell/issues#2322 (with a corresponding backend PR)


rory-landau-gravwell (approved):
LGTM

PR #2518, opened 2026-07-07
feat(#2516): LLM Proxy Ingester

david-fritz-gravwell:
Fixes gravwell/issues#2516

## This PR proposes...

- Adding an LLM proxy/ingester
- Support for ingesting message deltas, user messages, or full conversations
- Best-effort session tracking via intrinsic EVs
- Support for client and upstream auth
- Support for multiple proxies


ingesters/llm_ingester/config.go
```diff
@@ -0,0 +1,244 @@
...
+	Upstream_Authorization            string
+	Session_TTL                       string
+	Max_Body                          int
+	TLS_Certificate_File              string
+	TLS_Key_File                      string
+	Insecure_Skip_TLS_Verify_Upstream bool
+	Preprocessor                      []string
+
+	// derived during Verify
+	sessionTTL         time.Duration
```
daniel-moreno-levy-gravwell, line 104: What happens to these? Are they also overridable by user cfg? Or does or parser not try to populate them?

ingesters/llm_ingester/config.go
```diff
@@ -0,0 +1,244 @@
...
+
+type cfgReadType struct {
+	Global       gbl
+	Attach       attach.AttachConfig
+	Listener     map[string]*listener
+	Preprocessor processors.ProcessorConfig
+}
+
+type listener struct {
+	Bind                              string
```
daniel-moreno-levy-gravwell: I know we will document these in the wiki, but I really wish these were documented here as well, not just for review sake, but whoever needs to change these next.

daniel-moreno-levy-gravwell: Never mind... I saw that it is documented in the example config. Maybe a small `// See gravwell_llm_ingester.conf` would help here.

daniel-moreno-levy-gravwell:
More of a general comment, but for the wiki docs, we better tell folks somewhere that both [codex](https://developers.openai.com/codex/config-advanced) and [claude code](https://code.claude.com/docs/en/network-config) support proxy settings, and like to their docs, so they quickly have available how to setup this proxy with their client.

ingesters/llm_ingester/ingest.go
```diff
@@ -0,0 +1,178 @@
...
+	durationMs   int64
+	lg           *log.Logger
+}
+
+// emitRequestEvents writes the request-side events that the listener's
+// Log_Mode dictates.
+func emitRequestEvents(ec *emitCtx, evs []protocol.Event) {
+	switch ec.logMode {
+	case logModeUserOnly:
+		// Only the most recent user message.
```
daniel-moreno-levy-gravwell, line 47: Some agents, like opencode, let you interrupt your session at any point and then resume it with another user message. There are instances where I have interrupted a message right after pressing enter to add additional context. I worry that this simple heuristic will skip messages for this case, if it so happens that under a slow connection, someone interrupts the message, adds another one, and only the latest one is captured.

daniel-moreno-levy-gravwell: I'm wondering if having a digest of each message for each session is the way to go instead. `map[uuid]bool` instead. You can still do the reverse order lookup of the messages of the request, but at least you do know for sure what the real "last" message you've seen is, and use it as the stop condition... This applies specially to the delta algo below.

ingesters/llm_ingester/protocol/protocol.go
```diff
@@ -0,0 +1,123 @@
...
+// Package protocol defines an LLM provider so the proxy core can parse
+// requests/responses without knowing the vendor.
+package protocol
+
+import (
+	"fmt"
+	"sync"
+)
+
+// EventType values categorize a logged event.
```
daniel-moreno-levy-gravwell, line 18: Some links to whatever documentation you used to make this structs / get this const values would be helpful.

ingesters/llm_ingester/ingest.go
daniel-moreno-levy-gravwell: Thinking on this a bit more, and after some research, there is nothing in the OpenAI completions spec that prevents several consecutive user messages. We cannot predict clients doing wonky stuff, even if it is best practice to not have consecutive user messages one after the other.

ingesters/llm_ingester/session.go
```diff
@@ -0,0 +1,206 @@
...
+// session (no matching prefix found).
+func (s *sessionStore) Resolve(client string, hashes []string) (string, bool) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.evictNolock()
+	if len(hashes) == 0 {
+		// nothing to match on; mint
+		return s.mintNolock(client, hashes), true
+	}
+	// We compare against hashes[:n-1] (everything except the just-added user
```
daniel-moreno-levy-gravwell: See https://github.com/gravwell/gravwell/pull/2518/changes#r3544276090 ... But I would caution against this assumption. Most of the time, this is true, but we cannot predict how clients use the completions API, we cannot predict how other providers will work and we cannot predict system errors. Two, three or more user messages could be stringed together in a convo. Clients could be sending us tool calls first or an old conversation that ended with an assistant message.

ingesters/llm_ingester/session.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+		if len(filtered) == 0 {
+			delete(s.byClient, k)
+		} else {
+			s.byClient[k] = filtered
+		}
+	}
+}
+
+// Flush writes the in-memory state to disk (if persistence is enabled).
```
daniel-moreno-levy-gravwell, line 230: Does this mean, in `ingesters/llm_ingester/gravwell_llm_ingester.conf`

```
	# Persistent state for the auto-derived session tracker. Stores only
	# message hashes (never prompt content) so it can be reloaded after a
	# restart without losing in-flight conversations.
	State-Store-Location = llm_ingester.state
```

Is optional? I think it should be commented or clarified as such.

ingesters/llm_ingester/session_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+package main
+
+import (
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+func TestSessionNewConversation(t *testing.T) {
```
daniel-moreno-levy-gravwell, line 17: I would love to see more tests here with some of the more wonky edge cases I mentioned before. Consecutive user messages, convos that start witht he assistant messages, clients mishaving, etc.

daniel-moreno-levy-gravwell (changes requested):
Biggest issue with the current implementation is that it assumes too much about the order of messages in a session. I think we need to take a step back and treat this as what it is, a poorly cobbled together semi-improvised standard.  Anything can happen. I think the only assumption should be that the messages are ordered by most recent to least. We should not make any other assumptions about when user messages go, tools, or assistants.

Additionally, I think it is worth exploring making e2e tests with an embedded claude / opencode / codex agent, just to have a guarantee that the happy paths for these will work and continue to work, using their cheapest models. This can be deferred to another ticket.

ingesters/llm_ingester/protocol/protocol.go
daniel-moreno-levy-gravwell: Ignore this comment... I saw the openai.go file afterwards, this is clearly our "domain" types.

ingesters/llm_ingester/config.go
david-fritz-gravwell: those are just typed fields / compound fields derived from the actual user-config options. For example sessionTTL is the same as Sesstion_TTL -- but it's a time.Duration instead of a string...

ingesters/llm_ingester/config.go
david-fritz-gravwell: done

ingesters/llm_ingester/ingest.go
david-fritz-gravwell: changed the approach

ingesters/llm_ingester/session.go
david-fritz-gravwell: changed the approach

ingesters/llm_ingester/session.go
david-fritz-gravwell: done

ingesters/llm_ingester/session_test.go
david-fritz-gravwell: done

ingesters/llm_ingester/config.go
```diff
@@ -0,0 +1,282 @@
...
+	// upstream Authorization header, replacing whatever the client sent. Empty
+	// passes the client's own Authorization header through unchanged.
+	Upstream_Authorization string
+	// Session_TTL is how long idle session prefix-match state is retained,
+	// expressed as a Go duration string (e.g. "30m"). Defaults to
+	// defaultSessionTTL when unset.
+	Session_TTL string
+	// Max_Body is the maximum inbound request body size in bytes. Defaults to
+	// defaultMaxBody when unset or non-positive.
+	Max_Body                          int
```
daniel-moreno-levy-gravwell, line 97: Not a change, just a suggestion to emphasize this property in our wiki docs, because some providers like google and openai have video / images as valid input, so 16 MiB could be very small for that. Plus I do wonder what a 2 mil token convo is in byte size!

daniel-moreno-levy-gravwell:
DO NOT MERGE THIS PR... it needs more testing, going to test this ingester against some models / providers. There are edge cases that I want to make sure this thing captures (encrypted thinking blocks, for example)

ingesters/llm_ingester/protocol/openai/chat.go
```diff
@@ -0,0 +1,249 @@
...
+// We hash role + content so that re-sending the same prior turns matches.
+func hashMessage(m chatMessage) string {
+	h := sha256.New()
+	h.Write([]byte(m.Role))
+	h.Write([]byte{0})
+	h.Write(contentToBytes(m.Content))
+	// include tool-call structure for assistant tool turns
+	if len(m.ToolCalls) > 0 {
+		for _, tc := range m.ToolCalls {
+			h.Write([]byte{0})
```
daniel-moreno-levy-gravwell, line 289: Why the null bytes?

ingesters/llm_ingester/protocol/openai/stream.go
```diff
@@ -0,0 +1,203 @@
+/*************************************************************************
```
daniel-moreno-levy-gravwell, line 1: Very nice code :)

ingesters/llm_ingester/protocol/openai/stream_test.go
```diff
@@ -0,0 +1,172 @@
+/*************************************************************************
```
daniel-moreno-levy-gravwell, line 1: Nitpick... But I would add some tests for malformed JSON or incoherent messages, which can happen. The code above does seem to deal with that, so not gonna make it a blocker.

ingesters/llm_ingester/proxy.go
```diff
@@ -0,0 +1,402 @@
+/*************************************************************************
```
daniel-moreno-levy-gravwell, line 1: I like how the `Protocol` interface is being used here. It worries me a bit that other providers will not have the quirks of openai and we will have to change the interface, but you can't plan for everything. This is a very good first stab at it.

ingesters/llm_ingester/proxy.go
```diff
@@ -0,0 +1,402 @@
...
+				log.KV("listener", h.name), log.KVErr(err))
+		}
+		h.forwardWithoutIngest(w, r, body)
+		return
+	}
+	ec.model = parsedReq.Model
+	ec.stream = parsedReq.Stream
+
+	// Sessions are partitioned by client IP
+	sessionID, newSession := h.sessions.Resolve(ec.clientIP.String(), parsedReq.MessageHashes)
```
daniel-moreno-levy-gravwell, line 131: It's worth mentioning that IP could change between sessions IE, clients continuing their convos from another location... I don't think it's worth a change atm, the edge case is pretty niche and at worst we will capture the same data again... but I am curious why you decided to partition by client ip. IMO, just looking at the messages hashes is as close as the source of truth as we can get.

daniel-moreno-levy-gravwell, line 131: Also, super nitpicky, but technically `newSession` can be new or not... `resolvedSession` or just `session` would be better.

ingesters/llm_ingester/session.go
```diff
@@ -0,0 +1,249 @@
...
+	bestLen := -1
+	for _, c := range candidates {
+		if c.Len == 0 || len(hashes) < c.Len {
+			continue
+		}
+		// The stored Tail covers absolute positions [c.Len-len(c.Tail), c.Len)
+		// of the prior request. Those positions must be identical in the
+		// incoming request for it to be a continuation.
+		start := c.Len - len(c.Tail)
+		if !slices.Equal(c.Tail, hashes[start:c.Len]) {
```
daniel-moreno-levy-gravwell, line 161: The problem with this is that, if clients modify the tail of the convos to keep context limits (on long enough sessions), this will fail miserably... I'll do some research into what "the common prior art (if there is any?)" is... 

When I originally suggested this, I was more thinking of talling up message hash hits for each session, and the one with the most hits wins, without doing all or nothing check for the tail length.

Again, this might be fine actually. It should probably be called out in the wiki for delta mode. If it actually becomes a problem, we can change it in the future to be less strict.

ingesters/llm_ingester/session.go
```diff
@@ -0,0 +1,249 @@
...
+	if len(bucket) > s.maxPerClient {
+		// drop the oldest
+		bucket = bucket[len(bucket)-s.maxPerClient:]
+	}
+	s.byClient[client] = bucket
+	return id
+}
+
+// evictNolock drops entries older than the TTL. Caller holds s.mu.
+func (s *sessionStore) evictNolock() {
```
daniel-moreno-levy-gravwell, line 207: https://github.com/gravwell/gravwell/pull/2518/changes#r3552028715 is even more relevant in this context, because we are actively eliminating old sessions. I for one continue old sessions all the time. This will make old long convos repeat entries.

To be fair, this probably doesn't have a good solution... Only thing I can think of, make `defaultSessionTTL` to 24 hours and call this out in the wiki docs.

daniel-moreno-levy-gravwell (changes requested):
NItpicks, praises and calling out trade-offs... This is likely very close to as good as it is going to get. (Cache invalidation and making up sessions are hard problems) Once these are answered, I'll give the approval.

daniel-moreno-levy-gravwell:
I still want to do some testing on this before merging, and create an issue for e2e tests using actual cli clients (they can be pointing to our self-hosted models to cut cost)

ingesters/llm_ingester/config.go
david-fritz-gravwell: This isn't for guarding against valid chat completions but rather to guard against invalid/malicious responses attempting to exhaust memory. There's no expectation that we'd see anything more than kilobytes in practice.

ingesters/llm_ingester/protocol/openai/stream_test.go
david-fritz-gravwell: I assume our use of `encoding/json` covers all malformed input by returning errors no?

ingesters/llm_ingester/proxy.go
david-fritz-gravwell: It was meant as some measure of future proofing, but in practice this ingester will likely live on as _just_ an openai protocol ingester. We could just remove the protocol registration outright.

ingesters/llm_ingester/proxy.go
david-fritz-gravwell: Right -- have to do _something_. My concern is attempting to scale this to 5k clients. Session tracking will start knocking over apple carts even with what we're doing here...

david-fritz-gravwell: newSession there is a bool and true if the session is new. From the Resolve() docstring:

```
// Resolve looks up or mints a session ID for the incoming request.
// hashes is the canonical per-message hash sequence (oldest message first,
// latest message last). It returns the session ID and a boolean indicating
// whether this was a new session (no matching conversation found).
```

ingesters/llm_ingester/session.go
david-fritz-gravwell: Well that's why we keep a configurable tail length -- if agents are doing conversation culling/compaction, then the "tail" should still be intact.

ingesters/llm_ingester/session.go
david-fritz-gravwell: Something something halting problem.

ingesters/llm_ingester/proxy.go
daniel-moreno-levy-gravwell: Damn, I missed that... Fair enough.

ingesters/llm_ingester/proxy.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+	host = strings.TrimSpace(host)
+	if host == "" {
+		host, _, _ = net.SplitHostPort(r.RemoteAddr)
+	}
+	if ip := net.ParseIP(host); ip != nil {
+		return ip
+	}
+	return net.ParseIP("127.0.0.1")
+}
```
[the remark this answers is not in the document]
daniel-moreno-levy-gravwell: @david-fritz-gravwell this is not a bad suggestion I think

ingesters/llm_ingester/protocol/openai/stream.go
```diff
@@ -0,0 +1,203 @@
...
+			Content   string `json:"content,omitempty"`
+			ToolCalls []struct {
+				Index    int    `json:"index"`
+				ID       string `json:"id,omitempty"`
+				Type     string `json:"type,omitempty"`
+				Function struct {
+					Name      string `json:"name,omitempty"`
+					Arguments string `json:"arguments,omitempty"`
+				} `json:"function,omitempty"`
+			} `json:"tool_calls,omitempty"`
```
daniel-moreno-levy-gravwell, line 70: You are missing a `reasoning` block here to capture when the agent is "reasoning"...

IE:
```json
{
   "id":"chatcmpl-373",
   "object":"chat.completion.chunk",
   "created":1783709903,
   "model":"qwen3.6:35b",
   "system_fingerprint":"fp_ollama",
   "choices":[
      {
         "index":0,
         "delta":{
            "role":"assistant",
            "content":"",
            "reasoning":"The" <--- THIS
         },
         "finish_reason":null
      }
   ]
}
```

ingesters/llm_ingester/ingest.go
```diff
@@ -0,0 +1,178 @@
...
+	logUsage     bool
+	clientIP     net.IP
+	sessionID    string
+	newSession   bool
+	upstreamCode int
+	stream       bool
+	requestID    string
+	model        string
+	durationMs   int64
+	lg           *log.Logger
```
daniel-moreno-levy-gravwell, line 39: should add `reasoning` here, and make sure to pass it as a intrinsic EV

ingesters/llm_ingester/protocol/openai/chat.go
```diff
@@ -0,0 +1,249 @@
...
+
+// The following types mirror the OpenAI Chat Completions request/response
+// schema; see https://developers.openai.com/api/docs for the authoritative
+// field definitions.
+
+// chatMessage covers the subset of the message shape we care about.
+// Content may be a string OR an array of content parts in the multimodal form.
+type chatMessage struct {
+	Role       string          `json:"role"`
+	Content    json.RawMessage `json:"content,omitempty"`
```
daniel-moreno-levy-gravwell: During my testing with opencode, I've notice that this can sometimes be a JSON array for user messages. For example, in plan mode:

Image

It looks like it appends a message, coming from the user, with the "reminder" to use plan mode. 

It's worth looking if `Content['type'] == 'text'` and try to capture `Context['text']` a user message, if the effort isn't too high.

daniel-moreno-levy-gravwell (changes requested):
Some addtional feedback based on my testing... Could you additionally add a README.md to `ingesters/llm_ingester/README.md` such as this for maintainability? There were some head scratchers trying to setup this ingester locally for testing:

```markdown
# LLM Ingester

An OpenAI-compatible HTTP proxy and Gravwell ingester. Point a client at it
instead of the upstream provider and it transparently forwards traffic while
ingesting prompts, responses, tool calls, and token usage as structured
events. See [`gravwell_llm_ingester.conf`](gravwell_llm_ingester.conf) and
[`config.go`](config.go) for all config options (log mode, auth, session
tracking, etc).

## Local testing

\`\`\`markdown
go build .
./llm_ingester -config-file gravwell_llm_ingester.conf
\`\`\`

Point `Upstream-URL` in the config at your real provider (and set
`Upstream-Authorization` if you want the proxy to inject the real key instead
of passing the client's through).

### opencode

Add a custom provider pointing at the listener (default `:4180`) to
`opencode.json` / `opencode.jsonc`:

`\`\`jsonc
{
  "$schema": "https://opencode.ai/config.json",
  "provider": {
    "llm-ingester": {
      "name": "LLM Ingester (local)",
      "options": {
        "baseURL": "http://localhost:4180/v1"
      },
      "models": {
        "<your-model>": {}
      }
    }
  }
}
\`\`\`

### crush

Add a custom provider to `.crush.json` / `crush.json`:

\`\`\`json
{
  "$schema": "https://charm.land/crush.json",
  "providers": {
    "llm-ingester": {
      "name": "LLM Ingester (local)",
      "type": "openai",
      "base_url": "http://localhost:4180/v1",
      "api_key": "$YOUR_API_KEY",
      "models": [
        { "id": "<your-model>", "name": "<your-model>" }
      ]
    }
  }
}
\`\`\`

### Codex / Claude Code

See their proxy config docs:

- [Codex](https://developers.openai.com/codex/config-advanced)
- [Claude Code](https://code.claude.com/docs/en/network-config)

### curl

\`\`\`shell
curl http://localhost:4180/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $YOUR_API_KEY" \
  -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"hello"}]}'
\`\`\`

Search the configured tag (`llm` by default) in Gravwell to see ingested
events.

Run tests with `go test ./...`.
```

ingesters/llm_ingester/protocol/openai/stream.go
david-fritz-gravwell: added

ingesters/llm_ingester/ingest.go
david-fritz-gravwell: added

ingesters/llm_ingester/protocol/openai/chat.go
david-fritz-gravwell: added but I'm struggling to get this to occur on my side -- can you test locally?

ingesters/llm_ingester/README.md
```diff
@@ -0,0 +1,83 @@
...
+An OpenAI-compatible HTTP proxy and Gravwell ingester. Point a client at it
+instead of the upstream provider and it transparently forwards traffic while
+ingesting prompts, responses, tool calls, and token usage as structured
+events. See [`gravwell_llm_ingester.conf`](gravwell_llm_ingester.conf) and
+[`config.go`](config.go) for all config options (log mode, auth, session
+tracking, etc).
+
+## Local testing
+
+\`\`\`markdown
```
daniel-moreno-levy-gravwell: Remove the `\` here and in other places below, don't escape the backticks

ingesters/llm_ingester/protocol/openai/chat.go
daniel-moreno-levy-gravwell: What you did worked! The plan mode message appears as it's own entry now!

image

daniel-moreno-levy-gravwell (changes requested):
Small readme nitpick, after that, we are good and will be LGTM

daniel-moreno-levy-gravwell (approved):

Issue #2462, opened 2026-06-08
gwcli: gate `admin` and `cbac` navs


rory-landau-gravwell:
# Enhancement

Before opening an enhancement request, please complete the following checklist:
- [x] Check to see if this is a duplicate

## What is the enhancement to be made?

The `admin` and `cbac` navs should be gated behind whether the user is an admin and whether the system has CBAC enabled.

## Why should we make this change? (Business justification? What problem is the feature trying to solve?)

This would better mimic the behavior of the old cli and reduce the frustration from unauthorized/unable users navigating all the way through an action only to be told they cannot complete it.

## Any other comments?

This *should* be relative easy, but it will require tweaking the traversal engine for both Cobra and Mother which always requires care.

rory-landau-gravwell:
Thinking out loud: it may make more sense to leverage Cobra's command annotation system to auto-hide specific actions if a user is not an admin. The current design follows the tree design of the origin (specifically, some navs have sister navs under `admin` that contain admin-only actions not appropriate for the standard nav). This, however, requires a user to recognize when an action is admin-only (and thus to check `admin X <action>` instead of `X <action>`.

The traversal and suggestion engines would need to read the annotations out of each command they walk and display the action iff:  it does *not* have the admin tag || user is an admin.

This check would need to be implemented in ppre() *and* Mother's traversal+suggestion engines.

rory-landau-gravwell:
Okay, our discussion in dev sync cleared up the path forward.

## Plan
- We are going to gate at action level, rather than nav level. I should be able to accomplish this by leveraging cobra's annotation system and a lookup table or two in Mother.
- The admin nav will be slimmed down severely as I redistribute actions to their formerly-non-admin counterpart navs. No longer will we split actions between `navX` and `admin navX`.
- Things the user does not have access to will be disabled and/or throw errors; they will not be hidden. I can trivially apply the "disabled" lipgloss style to actions and navs based on the user's admin status. I will attempt to catch and wrap these errors so they are clear.

## Notes:
- Greater granularity of enabling/disabling based on cbac may need to come later, as it will balloon the scope of this issue.
- All of this is functionally syntactic sugar; the TUI cannot grant any power the backend does not allocate the user/token.

PR #2558, opened 2026-07-15
feat(mono#2462): gate admin navs and actions

rory-landau-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/2462

## This PR proposes gating admin-only actions/navs

- Actions and navs that require admin permissions are disabled if the current user is not an admin.

An error is thrown if a user attempts to navigate to an admin-only nav or execute an admin-only action. Any user can still call -h on these nodes to learn about them.

image
image
image

- In interactive mode, admin-only nodes can still be suggested to non-admins, but are greyed out and inaccessible.

image
image
image

>[!NOTE]
>Admin status is cached at login-time. A user can force a refresh of their admin status by logging back in or calling `self myinfo -r`.

## This PR does NOT include CBAC gating

~~The client library does not have a way to tell if CBAC is enabled (it can only determine if the license allows CBAC). I am investigating alternative solutions.~~

## PR Tasks

## Reviewer Tasks


rory-landau-gravwell:
Realized my admin solution conflicts with my CBAC solution. Closing this while I rework the implementation.

PR #2579, opened 2026-07-21
Fix searchctrl list URL

john-floren-gravwell:
This is the correct URL to use for listing... the other works, but this follows the pattern better.


rory-landau-gravwell (changes requested):
This PR kills `/api/searchctrl`, but the [api-docs](https://github.com/gravwell/api-docs/pull/724/changes) and [backend](https://github.com/gravwell/backend/pull/4216/changes#diff-ff81bf13efa6b59da85bfc21f35cd56fcd5e90d8545e74c020ef11d527b9cc12) PRs retain it. 
Is this intentional? Am I missing something?

john-floren-gravwell:
> This PR kills `/api/searchctrl`, but the [api-docs](https://github.com/gravwell/api-docs/pull/724/changes) and [backend](https://github.com/gravwell/backend/pull/4216/changes#diff-ff81bf13efa6b59da85bfc21f35cd56fcd5e90d8545e74c020ef11d527b9cc12) PRs retain it. Is this intentional? Am I missing something?

It kills it because we never actually use it anywhere in the client library. On the backend, it's just the same code that handles /api/searchctrl/list, so it doesn't really cost anything to keep it around.

rory-landau-gravwell (approved):
Fair enough. Though, should we remove it from api-docs, then?

PR #2601, opened 2026-07-29
llm_ingester: fix user mode logging of assistant replies. 

david-fritz-gravwell:
Fixes gravwell/issues#2679

## This PR proposes...

- Fixing an issue where assistant replies were logged while in user mode
- Forcing tool calls to log when the log tool calls flag is set while in user mode


kris-watts-gravwell (approved):
LGTM

PR #2602, opened 2026-07-29
vector preprocessor: Don't create embeddings on entries without a DATA field

david-fritz-gravwell:
Fixes gravwell/issues#2681

## This PR proposes...

- Not creating embeddings on entries with no DATA field populated
- Slight refactor to make it work with batches
- Add tests


kris-watts-gravwell (approved):
LGTM

Issue #2645, opened 2026-08-14
Hosted Runner: config is not getting registered right and the UUID is not getting picked up


kris-watts-gravwell:
# Bug Report

Before opening a bug report, please complete the following checklist:
- [ ] Check the documentation for intended behavior, notes, warnings, and edge cases
- [ ] Check open AND closed issues to see if this is a duplicate
	- A closed duplicate issue may exist if this is a wontfix

## What are the steps to reproduce this issue?

1. Run hosted ingester

## What happens?

The main config block is partially populated and GUI does not pickup the UUID

## What were you expecting to happen?

<img width="1012" height="331" alt="Image" src="https://github.com/user-attachments/assets/1d8644ce-78a4-46f6-84d2-10230a4b8f28" />

GUI populates the UUID here

## Any logs, error output, etc?

See screenshot above

## Any other comments?

Looks like its just an artifact of the "configuration" not getting fully populated in the return blob and ingester registration

## What versions of software are you using?

5.10.0

nic-osullivan-gravwell:
lgtm 5.10.1 beta1

One config, before was missing the uuid after had it. 

<details>
<summary>Before</summary>

<img width="1920" height="1215" alt="Image" src="https://github.com/user-attachments/assets/72ae348b-950f-4e1d-94fc-325aec7acc72" />

</details>

<details>
<summary>After</summary>

<img width="1920" height="1215" alt="Image" src="https://github.com/user-attachments/assets/597efc3d-9c73-49a6-8764-5d9071e61581" />

</details>

PR #2646, opened 2026-08-14
Hosted runner config registration fix

kris-watts-gravwell:
This PR addresses #2645 

## This PR proposes...

- slightly tweak the config structure so it matches what the webserver and GUI expect

image

UUID is registered and the backend targets and log levels are registered too

## PR Tasks

## Reviewer Tasks


kris-watts-gravwell:
will likely fail until this is merged: https://github.com/gravwell/gravwell/pull/2644

hosted/runner/config.go
```diff
@@ -58,5 +80,30 @@ func (c cfgType) AttachConfig() attach.AttachConfig {
...
+		return
+	}
+	tagMp := make(map[string]bool, len(ptags))
+	for _, tag := range ptags {
+		if err = ingest.CheckTag(tag); err != nil {
+			err = fmt.Errorf("invalid tag %q: %w", tag, err)
+			return
+		} else if tagMp[tag] {
+			continue
+		}
```
nic-osullivan-gravwell: This is just a bit faster and a bit cleaner to me. Probably don't want to accept this directly, lots of spaces instead of tabs. 
```suggestion
		if tagMp[tag] {
			continue
		}
		if err = ingest.CheckTag(tag); err != nil {
			err = fmt.Errorf("invalid tag %q: %w", tag, err)
			return
             }
```

hosted/runner/config.go
```diff
@@ -58,5 +80,30 @@ func (c cfgType) AttachConfig() attach.AttachConfig {
...
+		tagMp[tag] = true
+		tags = append(tags, tag)
+	}
+	if len(tags) > 0 {
+		sort.Strings(tags)
```
nic-osullivan-gravwell, line 107: Is there a downstream benefit to having the tags sorted?

nic-osullivan-gravwell (changes requested):
one small nit and question, otherwise good.

hosted/runner/config.go
kris-watts-gravwell: they will be sorted in the GUI

nic-osullivan-gravwell: This seems odd to rely on the ingester to send these sorted if we want them displayed sorted. If we are going to be shipping these out of band, and in a dream world having people other than us write them, we should be handling it in the indexer or webserver when displaying them. 

However that's a future problem not a now problem.

nic-osullivan-gravwell (approved):
lgtm

Issue #2658, opened 2026-08-15
hosted ingesters: support dynamic reload


kris-watts-gravwell:
# Feature

Before opening a feature request, please complete the following checklist:
- [ ] Check to see if this is a duplicate

## What is the feature to be added?

hosted ingesters are taking on more and more load, they should support dynamic reload

## Why should we add this feature? (Business justification? What problem is the feature trying to solve?)

When we host 3-5 ingesters it sucks to have to restart the ingester to make small tweaks, a SIGHUP like the HttpIngester should reload the config, figure out what needs to be done and do it.

## How does this feature address the above problem?

Makes it so the hosted ingester can take on more and more load and be positioned to be a universal ingester

## Any other comments?

Also lays the groundwork for GUI configured ingesters

garrett-larsen-gravwell:
+1 and I will update our automation to take advantage of this capability

nic-osullivan-gravwell:
lgtm 5.10.1 beta1

Add/remove/modify for tester all look good. No extra ingesters running through the process.

PR #2659, opened 2026-08-15
Hosted ingester dynamic reload

kris-watts-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/2658

## This PR proposes...

- adding a SIGHUP to dynamically reload the config
- Tweak interfaces so that plugins can compare their configs and declare if a config changed
- Tweak the context on plugins so that they use their own context and can be shutdown individually
- Added logic to reload a configs and then restart ingesters that need to be restarted
  -  Or shutdown deleted ones
  - or start new ones
- added E2E tests for everything

## PR Tasks

## Reviewer Tasks

### DIFF
--- a/hosted/plugins/okta/okta.go
+++ b/hosted/plugins/okta/okta.go
@@ -156,7 +156,7 @@ func (o *OktaIngester) userLogRoutine(ctx context.Context, rt hosted.Runtime) {
 			}
 			//just look for changes over the last boundary
 			rt.Debug("requesting users", log.KV("start", start.Format(timeFormat)), log.KV("end", end.Format(timeFormat)))
-			if err := o.getUserLogs(start, end, rt); err != nil {
+			if err := o.getUserLogs(ctx, start, end, rt); err != nil {
 				rt.Error("failed to get users", log.KV("error", err))
 			} else {
 				//success, update last ts run
@@ -171,7 +171,7 @@ func (o *OktaIngester) systemLogRoutine(ctx context.Context, rt hosted.Runtime) 
 		log.KV("start-ts", o.latestTS.Format(time.RFC3339)),
 		log.KV("next-url", o.systemLogsNext != nil))
 	for {
-		if err := o.getSystemLogs(rt); err != nil {
+		if err := o.getSystemLogs(ctx, rt); err != nil {
 			rt.Error("system log error", log.KV("error", err))
 		}
 		if rt.Sleep(time.Minute) { // this sleep will quit if the context cancels
@@ -180,10 +180,10 @@ func (o *OktaIngester) systemLogRoutine(ctx context.Context, rt hosted.Runtime) 
 	}
 }
 
-func (o *OktaIngester) getSystemLogs(rt hosted.Runtime) error {
+func (o *OktaIngester) getSystemLogs(ctx context.Context, rt hosted.Runtime) error {
 	var req *http.Request
 
-	req, err := http.NewRequestWithContext(rt.Context(), http.MethodGet, fmt.Sprintf("https://%s", o.cfg.Domain), nil)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("https://%s", o.cfg.Domain), nil)
 	if err != nil {
 		return fmt.Errorf("failed to build request %w", err)
 	}
@@ -204,7 +204,7 @@ func (o *OktaIngester) getSystemLogs(rt hosted.Runtime) error {
 	req.Header.Set(`Authorization`, fmt.Sprintf(`SSWS %s`, o.cfg.Token))
 
 	var quit bool
-	for !quit && rt.Context().Err() == nil {
+	for !quit && ctx.Err() == nil {
 		if !rt.Alive() {
 			// okta can back off and wait if the runtime isn't healthy, just loop and wait again
 			quit = rt.Sleep(emptySleepDur)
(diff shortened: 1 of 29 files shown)


hosted/plugins/okta/okta.go
```diff
@@ -184,10 +180,10 @@ func (o *OktaIngester) systemLogRoutine(ctx context.Context, rt hosted.Runtime)
 	}
 }
 
-func (o *OktaIngester) getSystemLogs(rt hosted.Runtime) error {
+func (o *OktaIngester) getSystemLogs(ctx context.Context, rt hosted.Runtime) error {
 	var req *http.Request
 
-	req, err := http.NewRequestWithContext(rt.Context(), http.MethodGet, fmt.Sprintf("https://%s", o.cfg.Domain), nil)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("https://%s", o.cfg.Domain), nil)
```
nic-osullivan-gravwell, line 186: Is this (and similar changes) needed? Based on the `scopedRuntime` this _shouldn't_ be needed. I do prefer passing `ctx` down rather than passing an struct holding it, but undoing this would cut down on the changes in the PR.

hosted/runner/manager.go
```diff
@@ -142,3 +157,110 @@ func (rm *runtimeManager) startIngesters() (err error) {
...
+			if err = rm.createRunner(name, builder); err != nil {
+				rm.lgr.Error("failed to create new ingester on reload",
+					log.KV("kind", builder.Kind()),
+					log.KV("name", name),
+					log.KV("uuid", guid),
+					log.KVErr(err),
+				)
+				// just continue
+				continue
+			} else {
```
nic-osullivan-gravwell: nit: drop the else unnecessary

hosted/runner/manager.go
```diff
@@ -142,3 +157,110 @@ func (rm *runtimeManager) startIngesters() (err error) {
...
+			id, name := wr.ID(), wr.Name()
+			if lerr := wr.Close(); lerr != nil {
+				rm.lgr.Error("failed to shutdown ingester on reload",
+					log.KV("id", id),
+					log.KV("name", name),
+					log.KV("uuid", guid),
+					log.KVErr(lerr),
+				)
+				// not much to do here other than complain about it... :/
+			} else {
```
nic-osullivan-gravwell: add a continue above and drop the else for consistency across the handling.

hosted/runner/manager.go
```diff
@@ -142,3 +157,110 @@ func (rm *runtimeManager) startIngesters() (err error) {
...
+				rm.lgr.Info("shutdown ingester on reload",
+					log.KV("id", id),
+					log.KV("name", name),
+					log.KV("uuid", guid),
+				)
+			}
+			delete(rm.mp, guid)
```
nic-osullivan-gravwell: This will delete even when `Close` fails losing a handle on the runner. The above continue/else should resolve that.

hosted/runner/manager.go
```diff
@@ -142,3 +157,110 @@ func (rm *runtimeManager) startIngesters() (err error) {
 	}
 	return
 }
+
+// TODO FIXME
```
nic-osullivan-gravwell: ?

hosted/runner/manager.go
```diff
@@ -75,31 +77,44 @@ func (rm *runtimeManager) createRunners(c *cfgType, ib base.IngesterBase) (err e
...
-		}
-		if existing, exists := rm.mp[builder.UUID()]; exists {
-			ib.Logger.Error("hosted runner UUID collision",
-				log.KV("existing-uuid", existing.UUID()),
-				log.KV("colliding-type", existing.ID()),
-				log.KV("colliding-name", name),
-				log.KV("colliding-uuid", builder.UUID()))
-			continue // just skip it
+		if err = rm.createRunner(name, builder); err != nil {
+			if err == errExists {
```
nic-osullivan-gravwell: use `errors.Is`

nic-osullivan-gravwell (changes requested):
The core bits looks good, mostly small changes.

hosted/runner/manager.go
kris-watts-gravwell: if we fail to close an ingester it is kind of a zombie and we will hit it every time we reload.
Is that a probem?  Is there a scenario where a close throws an error but succeeds later?  Do we WANT to keep an ingester around that we tried to remove and the close failed?

hosted/plugins/okta/okta.go
kris-watts-gravwell: I put it in to make restarts easier where we don't have to cancel the main runtime context (which is tied to the process shutdown).

I can be convinced otherwise, but I would leave it in for the following:
1. its the stdlib pattern
2. I have some visions of other runtimes that need to be signaled independently as a why to abruptly halt a misbehaving plugin.  So the one we pass in is a "please" and the one in the runtime is more violent.

hosted/runner/manager.go
nic-osullivan-gravwell: I think keeping a handle on it is better than not. Otherwise we have 0 recourse to try and shut it down. Losing the handle prevents us from sending another signal to try and close again. 

We probably don't need to do anything in code other than keep a handle on it for now. This gets trickier when we ever get to HA support, but without keeping a handle our options shrink.

hosted/plugins/okta/okta.go
nic-osullivan-gravwell: I agree on the stdlib pattern and am fine to keep it as that's generally how I've been writing hosted ingesters, but wasn't sure if it's worth changing most the ingesters for it.

hosted/runner/manager.go
kris-watts-gravwell: LGTM, current code does this.

nic-osullivan-gravwell (approved):
lgtm

PR #2670, opened 2026-08-18
feat(search): use access endpoint

tyler-elton-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/2708

## This PR proposes...

- updating the cli/client to use the search access endpoint for updating readers

## PR Tasks

## Reviewer Tasks

### DIFF
--- a/client/searchCtrl.go
+++ b/client/searchCtrl.go
@@ -69,11 +69,7 @@ func (c *Client) BackgroundSearch(sid string) error {
 // SetAccess sets the Readers/Writers ACLs and optionally reassigns ownership (for admins only)
 // for an existing search. Pass a nil ownerID to leave ownership unchanged.
 func (c *Client) SetAccess(sid string, ownerID *int32, readers, writers types.ACL) error {
-	request := struct {
-		OwnerID *int32
-		Readers types.ACL
-		Writers types.ACL
-	}{
+	request := types.SearchCtrlSetAccessRequest{
 		OwnerID: ownerID,
 		Readers: readers,
 		Writers: writers,
--- a/client/searchCtrl.go
+++ b/client/searchCtrl.go
@@ -66,17 +66,19 @@ func (c *Client) BackgroundSearch(sid string) error {
 	return c.patchStaticURL(searchCtrlBackgroundUrl(sid), nil)
 }
 
-// SetGroups sets the list of groups that can read the search
-func (c *Client) SetGroups(sid string, gids []int32) error {
-	request := struct{ GIDs []int32 }{gids}
-	return c.putStaticURL(searchCtrlGroupsUrl(sid), request)
-}
-
-// SetGlobal is an admin-only function to toggle sharing of results
-// with the entire system.
-func (c *Client) SetGlobal(sid string, global bool) error {
-	request := struct{ Global bool }{global}
-	return c.putStaticURL(searchCtrlGlobalUrl(sid), request)
+// SetAccess sets the Readers/Writers ACLs and optionally reassigns ownership (for admins only)
+// for an existing search. Pass a nil ownerID to leave ownership unchanged.
+func (c *Client) SetAccess(sid string, ownerID *int32, readers, writers types.ACL) error {
+	request := struct {
+		OwnerID *int32
+		Readers types.ACL
+		Writers types.ACL
+	}{
+		OwnerID: ownerID,
+		Readers: readers,
+		Writers: writers,
+	}
+	return c.putStaticURL(searchCtrlAccessUrl(sid), request)
 }
 
 // ListSearches returns a list of all searches the current user has access to
--- a/gwcli/tree/queries/queries.go
+++ b/gwcli/tree/queries/queries.go
@@ -377,7 +377,7 @@ func setGroup() action.Pair {
 		func(IDs []string, _ *pflag.FlagSet) (results []scaffold.Result, _ error) {
 			results = make([]scaffold.Result, len(IDs))
 			for i, ID := range IDs {
-				if err := connection.Client.SetGroups(ID, GIDs); err != nil {
+				if err := connection.Client.SetAccess(ID, nil, types.ACL{GIDs: GIDs}, types.ACL{}); err != nil {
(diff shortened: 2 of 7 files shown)


client/searchCtrl.go
```diff
@@ -66,17 +66,19 @@ func (c *Client) BackgroundSearch(sid string) error {
...
-}
-
-// SetGlobal is an admin-only function to toggle sharing of results
-// with the entire system.
-func (c *Client) SetGlobal(sid string, global bool) error {
-	request := struct{ Global bool }{global}
-	return c.putStaticURL(searchCtrlGlobalUrl(sid), request)
+// SetAccess sets the Readers/Writers ACLs and optionally reassigns ownership (for admins only)
+// for an existing search. Pass a nil ownerID to leave ownership unchanged.
+func (c *Client) SetAccess(sid string, ownerID *int32, readers, writers types.ACL) error {
```
rory-landau-gravwell, line 71: I don't generally mind anon structs, but we should solidify this one; these are inherently coupled.

rory-landau-gravwell (changes requested):
One required change (see below) and a qualm:

Registry treats asset updates as struct-wide clobbers; it has no concept of "update field A but not field B".  Skimming the backend sister PR, it appears this follows that same philosophy, which I like. My concern is that combining these endpoints means that updating readers can now also clobber writers (and vice-versa). This endpoint will require the caller to know the pre-existing permissions and send back items it doesn't want to update. Is that Evan's intention?

tyler-elton-gravwell:
> One required change (see below) and a qualm:
> 
> Registry treats asset updates as struct-wide clobbers; it has no concept of "update field A but not field B". Skimming the backend sister PR, it appears this follows that same philosophy, which I like. My concern is that combining these endpoints means that updating readers can now also clobber writers (and vice-versa). This endpoint will require the caller to know the pre-existing permissions and send back items it doesn't want to update. Is that Evan's intention?

@evan-gravwell please see Rory's comment above

evan-gravwell:
> This endpoint will require the caller to know the pre-existing permissions and send back items it doesn't want to update. Is that Evan's intention?

This sounds good to me. One worry I have is, what if...
1. adminA grants userA write access via groupA membership to the searchinfo.
2. adminA gants groupB write access to the searchinfo.
3. userA doesn't have the ability to see groupB.

Will this situation behave the same as other asset implementations? EG userA GETs this searchinfo. the writers includes groupB. UserA can update the writers while including groupB.

Edit: and also, UserA can send back the same OwnerID=1 (AdminA)?

client/searchCtrl.go
tyler-elton-gravwell: addressed in d4e4bcd15da41af37649dcc724b6e22d308c47e5

rory-landau-gravwell:
>UserA can update the writers while including groupB.

I don't actually know. I'm not well versed in our permissions; Saint Nic is king there.
That being said, if that is how other assets (should) behave, then we should mirror it, keep cognitive load down.

Tyler, can you write a test for SearchInfo to validate that this behavior is currently represented and that it doesn't regress later?

tyler-elton-gravwell:
> > This endpoint will require the caller to know the pre-existing permissions and send back items it doesn't want to update. Is that Evan's intention?
> 
> This sounds good to me. One worry I have is, what if...
> 
> 1. adminA grants userA write access via groupA membership to the searchinfo.
> 2. adminA gants groupB write access to the searchinfo.
> 3. userA doesn't have the ability to see groupB.
> 
> Will this situation behave the same as other asset implementations? EG userA GETs this searchinfo. the writers includes groupB. UserA can update the writers while including groupB.
> 
> Edit: and also, UserA can send back the same OwnerID=1 (AdminA)?

Membership checks are diff-based. It will only validate _newly added_ GIDs against the caller's membership. This means that unchanged GIDs that's already on the ACL (even ones that the caller cannot see) are passed through untouched because that field is unchanged. 
This is preexisting behavior in registry, identical to every other asset type, not something specific to `SetAccess`. 

`SetAccess` will reject any non-nil owner ID from a non-admin outright, regardless of whether the value matches the current owner. Non-admins _have_ to pass `nil`, not echo the current value.

tyler-elton-gravwell:
> > UserA can update the writers while including groupB.
> 
> I don't actually know. I'm not well versed in our permissions; Saint Nic is king there. That being said, if that is how other assets (should) behave, then we should mirror it, keep cognitive load down.
> 
> Tyler, can you write a test for SearchInfo to validate that this behavior is currently represented and that it doesn't regress later?

There is already test coverage for this on the backend.

gwcli/tree/queries/queries.go
```diff
@@ -377,7 +377,12 @@ func setGroup() action.Pair {
 		func(IDs []string, _ *pflag.FlagSet) (results []scaffold.Result, _ error) {
 			results = make([]scaffold.Result, len(IDs))
 			for i, ID := range IDs {
-				if err := connection.Client.SetAccess(ID, nil, types.ACL{GIDs: GIDs}, types.ACL{}); err != nil {
+				si, err := connection.Client.GetSearch(ID)
+				if err != nil {
```
rory-landau-gravwell, line 381: Ty for updating gwcli!

rory-landau-gravwell (dismissed):
Concerns addressed, Evan confirmed shape.

LGTM!

ingest/processors/cisco_ise.go
```diff
@@ -73,9 +73,8 @@ func CiscoISELoadConfig(vc *config.VariableConfig) (c CiscoISEConfig, err error)
 func (c *CiscoISEConfig) validate() (err error) {
 	if c.Max_Multipart_Latency != `` {
 		if c.maxLatency, err = time.ParseDuration(c.Max_Multipart_Latency); err != nil {
-			if err = fmt.Errorf("Invalid Max-Multipart-Latency %q: %v", c.Max_Multipart_Latency, err); err != nil {
-				return
-			}
+			err = fmt.Errorf("Invalid Max-Multipart-Latency %q: %v", c.Max_Multipart_Latency, err)
```
rory-landau-gravwell, line 76: This is beyond scope so I don't want to touch it now, but we may need to come back to that `%v` later.

rory-landau-gravwell (approved):
LGTM. Don't know what got staticcheck so upset at code Tyler didn't touch???

Issue #2678, opened 2026-08-20
ingester: add anthropic api support to llm ingester


david-fritz-gravwell:
# Feature

Before opening a feature request, please complete the following checklist:
- [x] Check to see if this is a duplicate

## What is the feature to be added?

Add anthropic API support to the llm ingester to support using claude code.

## Why should we add this feature? (Business justification? What problem is the feature trying to solve?)

Greater coverage.

daniel-moreno-levy-gravwell:
Messed up and approved the [PR](https://github.com/gravwell/gravwell/pull/2679) when I meant to gave it more feedback. Reverted [here](https://github.com/gravwell/gravwell/pull/2687).

kris-watts-gravwell:
got it working end to end, seems fine.  LGTM 5.10.1-beta1

PR #2679, opened 2026-08-20
llm ingester: add anthropic API support

david-fritz-gravwell:
Fixes https://github.com/gravwell/gravwell/issues/2678


ingesters/llm_ingester/README.md
```diff
@@ -18,6 +28,134 @@ Point `Upstream-URL` in the config at your real provider (and set
...
+	Log-Mode = "delta"
+	Log-Tool-Calls = true
+	Log-Usage = true
+```
+
+Start the ingester as above, and it is listening on `:4181`.
+
+The antropic key can be held by claude code or the proxy:
+
+| | Listener | Claude Code's `ANTHROPIC_API_KEY` |
```
daniel-moreno-levy-gravwell: (FOR ME) Test these arrangements.

ingesters/llm_ingester/README.md
```diff
@@ -75,9 +210,10 @@ curl http://localhost:4180/v1/chat/completions \
...
-Search the configured tag (`llm` by default) in Gravwell to see ingested
-events.
 
-Run tests with `go test ./...`.
+curl http://localhost:4181/v1/messages \
+  -H "Content-Type: application/json" \
+  -H "anthropic-version: 2023-06-01" \
+  -H "x-api-key: $YOUR_ANTHROPIC_API_KEY" \
+  -d '{"model":"claude-opus-5","max_tokens":64,"messages":[{"role":"user","content":"hello"}]}'
+```
```
daniel-moreno-levy-gravwell, line 219: Nitpick. To make this obvious, so the reader's eyes land quickly to what the want.

```shell
# OpenAI
curl http://localhost:4180/v1/chat/completions \
-H "Content-Type: application/json" \
-H "Authorization: Bearer $YOUR_API_KEY" \
-d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"hello"}]}'

# Anthropic
curl http://localhost:4181/v1/messages \
  -H "Content-Type: application/json" \
  -H "anthropic-version: 2023-06-01" \
  -H "x-api-key: $YOUR_ANTHROPIC_API_KEY" \
  -d '{"model":"claude-opus-5","max_tokens":64,"messages":[{"role":"user","content":"hello"}]}'
```

ingest/processors/cisco_ise.go
```diff
@@ -73,9 +73,8 @@ func CiscoISELoadConfig(vc *config.VariableConfig) (c CiscoISEConfig, err error)
 func (c *CiscoISEConfig) validate() (err error) {
 	if c.Max_Multipart_Latency != `` {
 		if c.maxLatency, err = time.ParseDuration(c.Max_Multipart_Latency); err != nil {
-			if err = fmt.Errorf("Invalid Max-Multipart-Latency %q: %v", c.Max_Multipart_Latency, err); err != nil {
```
daniel-moreno-levy-gravwell: I almost want to ask you to leave it as is. Such beautiful artisanal tapestry, stamped out by the cold machine!

ingesters/llm_ingester/proxy.go
```diff
@@ -161,13 +153,80 @@ func (h *proxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	return body, true
+}
+
+// resolveSession decides which session a request belongs to. When the listener
+// names a session header and the request carries a usable value, the client is
+// telling us which conversation this is and we take its word for it; otherwise
+// we derive the session by matching the message prefix.
+func (h *proxyHandler) resolveSession(r *http.Request, client string, hashes []string) (string, bool) {
+	if hdr := h.cfg.Session_ID_Header; hdr != "" {
+		if id := sanitizeSessionID(r.Header.Get(hdr)); id != "" {
```
daniel-moreno-levy-gravwell, line 223: You should get the value r.Header.Get(hdr) before passing it to sanitizeSessionID. Then check if it is empty. If it is, you should at least log it with a Warn or an Error, because either the user's client is misbehaving or the llm_proxy is misconfigured, and the user should be notified. It's better than failing silently.

ingesters/llm_ingester/proxy.go
```diff
@@ -161,13 +153,80 @@ func (h *proxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// forwardUpstream asks for an identity encoding; an upstream that
+	// compresses anyway hands us bytes the protocol parser cannot read. Relay
+	// them untouched instead of feeding it garbage.
+	if enc := resp.Header.Get("Content-Encoding"); enc != "" && !strings.EqualFold(enc, "identity") {
+		if h.lg != nil {
```
daniel-moreno-levy-gravwell, line 160: When is logger ever not defined? I'm chasing it down, and it isn't ever the case, other than in tests.. I would rather throw an error in `newProxyHandler` if `lg` is nil, than have ifs covering scenarios that cannot happen or, if they do happen, we want to fail epically at startup so we can catch and fix it.

For tests use `log.NewDiscardLogger` instead, make `lg` use the https://github.com/gravwell/gravwell/blob/dd69fc792327fde879f7b5e1ccd179e828f54dbb/ingest/api.go#L53 Logger interface.

ingesters/llm_ingester/proxy.go
```diff
@@ -180,10 +239,20 @@ func (h *proxyHandler) forwardUpstream(r *http.Request, body []byte) (*http.Resp
...
+	// negotiation does not carry upstream (see the transport setup in
+	// newProxyHandler). The response we relay is simply uncompressed.
+	out.Header.Del("Accept-Encoding")
 	setForwardedFor(out, r)
+	// The Anthropic Messages API requires an anthropic-version header. When the
+	// proxy injects the upstream key the client never sends one, so supply a
+	// configured default if the request lacks it.
+	if v := h.cfg.Anthropic_Version; v != "" && out.Header.Get("anthropic-version") == "" {
+		out.Header.Set("anthropic-version", v)
+	}
```
daniel-moreno-levy-gravwell, line 255: This path is shared with OpenAI's, guard against it explicitely

```suggestion
	if h.proto.Name() == "anthropic-messages" && h.cfg.Anthropic_Version != "" && out.Header.Get("anthropic-version") == "" {
		out.Header.Set("anthropic-version", h.cfg.Anthropic_Version)
	}
```

ingesters/llm_ingester/session.go
```diff
@@ -249,5 +325,35 @@ func (s *sessionStore) Flush() error {
...
+// The value is client-controlled and lands in an enumerated value on every
+// entry, so anything oversized, empty, or non-printable is rejected (the caller
+// then falls back to prefix matching) rather than trusted.
+func sanitizeSessionID(v string) string {
+	v = strings.TrimSpace(v)
+	if v == "" || len(v) > maxSessionIDLen {
+		return ""
+	}
+	for i := 0; i < len(v); i++ {
+		if v[i] < 0x21 || v[i] > 0x7e {
```
daniel-moreno-levy-gravwell, line 354: Can you leave a comment about what this is about? What are these values? Why do they have to return empty string?

ingesters/llm_ingester/gravwell_llm_ingester.conf
```diff
@@ -88,6 +88,55 @@
...
+	# and the client was never configured to talk to Anthropic directly.
+	Anthropic-Version = "2023-06-01"
+
+	# Adopt the client's own conversation ID (from this header) as the
+	# session ID instead of deriving one by matching message prefixes.
+	# Claude Code stamps every request with x-claude-code-session-id, which
+	# survives the history rewrites (compaction, retries, edited context)
+	# that prefix matching cannot follow, and lets an ingested session be
+	# lined up with the client's own transcript. Unset to always derive.
+	Session-ID-Header = "x-claude-code-session-id"
```
daniel-moreno-levy-gravwell, line 127: I find this confusing. Are we leaving this open so our users can customize it for their clients, in case their providers do also set Session-ID-Headers? Do we know if any other provider does this?

There iss two reasons I ask:

1. If only anthropic does this, and we are not thinking of anyone else, then it shouldn't be an option to customize.
2. If we are thinking of other use cases, then I would explicitly say so in this doc string or the README because it reads as if we are only thinking of anthropic.

ingesters/llm_ingester/proxy.go
daniel-moreno-levy-gravwell: Additionally, this path is shared with `openAI`, but `Session_ID_Header` seems only relevant for anthropic. See my comment [here](https://github.com/gravwell/gravwell/pull/2679/changes#r3824566827).

ingesters/llm_ingester/session.go
```diff
@@ -41,6 +42,11 @@ type sessionEntry struct {
 type sessionStore struct {
 	mu       sync.Mutex
 	byClient map[string][]*sessionEntry
+	// explicit tracks sessions named by the client itself (see
+	// ResolveExplicit), keyed by client then by the client's identifier. These
+	// need no hash matching, only a last-seen stamp so the TTL and the
+	// new-session flag still work.
+	explicit map[string]map[string]time.Time
```
daniel-moreno-levy-gravwell, line 49: See https://github.com/gravwell/gravwell/pull/2679/changes#r3824566827

daniel-moreno-levy-gravwell (changes requested):
This feedback is mine, not from the robot. Claude found some extra stuff, but it is taking me time to express it clearly and remove the noise. Expect it tomorrow morning.

ingesters/llm_ingester/proxy.go
```diff
@@ -161,13 +153,80 @@ func (h *proxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	}
 }
 
+// serveUnparsedPath handles a request on a path the protocol module does not
+// parse. A provider's API is wider than the one endpoint we understand — the
+// Messages API has /v1/messages/count_tokens next to /v1/messages, and Claude
+// Code calls it — so by default those requests are proxied untouched and simply
+// not ingested. Reject_Unknown_Paths narrows the listener to the parsed paths.
+func (h *proxyHandler) serveUnparsedPath(w http.ResponseWriter, r *http.Request) {
+	if h.cfg.Reject_Unknown_Paths {
```
daniel-moreno-levy-gravwell, line 183: `Reject_Unknown_Paths` defaults to false. This makes the factory-configuration less secure. A user could stand a proxy of their own and re-route the llm_proxy to call an unrecognized path, to capture the `Upstream-Authorization` header from it.

For this reason, I would default to `Reject_Unknown_Paths` = true and make an explicit warning for this case in the documentation AND the logging, that it is inherently insecure. If there are paths that claude code is using that make the default too strict, add the to the allow list.

ingesters/llm_ingester/protocol/anthropic/messages.go
```diff
@@ -0,0 +1,374 @@
...
+			return []protocol.Event{{
+				Type:    protocol.EventUserMessage,
+				Role:    roleUser,
+				Content: []byte(str),
+			}}
+		}
+		var evs []protocol.Event
+		var text strings.Builder
+		for _, b := range blocks {
+			switch b.Type {
```
daniel-moreno-levy-gravwell, line 244: There is no default case in this switch. Unrecognized types (images, documents) are silently dropped. This contradicts the precedent from OpenAI in https://github.com/gravwell/gravwell/blob/dd69fc792327fde879f7b5e1ccd179e828f54dbb/ingesters/llm_ingester/protocol/openai/chat.go#L151 and https://github.com/gravwell/gravwell/blob/dd69fc792327fde879f7b5e1ccd179e828f54dbb/ingesters/llm_ingester/protocol/openai/chat.go#L245-L247 , that sends the raw JSON "to not drop anything silently". Something similar should happen here.

ingesters/llm_ingester/protocol/anthropic/messages.go
```diff
@@ -0,0 +1,374 @@
...
+			h.Write([]byte(b.ID))
+			h.Write([]byte{0})
+			h.Write([]byte(b.Name))
+			h.Write([]byte{0})
+			h.Write(b.Input)
+		case blockToolResult:
+			h.Write([]byte{0})
+			h.Write([]byte(b.ToolUseID))
+			h.Write([]byte{0})
+			h.Write(flattenText(b.Content))
```
daniel-moreno-levy-gravwell, line 358: Similar observation as https://github.com/gravwell/gravwell/pull/2679/changes#r3830691998. Since there is not a default branch in this switch, unrecognized blocks with the same type will have the same hash (from the above):

```golang
		h.Write([]byte{0})
		h.Write([]byte(b.Type))
```

I propose decoding the `content` for this case.

ingesters/llm_ingester/proxy.go
```diff
@@ -339,15 +410,15 @@ func copyRequestHeaders(dst, src *http.Request, upstreamAuth string) {
 		if _, drop := connDrop[ck]; drop {
 			continue
 		}
-		if upstreamAuth != "" && ck == "Authorization" {
+		if upstreamAuth != "" && ck == authKey {
 			continue // replaced below with the configured upstream credential
 		}
 		for _, v := range vs {
```
daniel-moreno-levy-gravwell, line 416: A client can sneak in their own credentials here. Anthropic accepts both `x-api-key` and `Authorization: Bearer ...`. A client could have both defined, and if upstreamAuth is set, it will only replace one of those. I would guard here against both `Authorization` and `x-api-key` for the case where `upstreamAuth` is defined. See this gist Claude generated, it illustrates the point: https://gist.github.com/daniel-moreno-levy-gravwell/dffcc5d2f5398f4026aeacd3d0fb1c3a

ingesters/llm_ingester/config.go
daniel-moreno-levy-gravwell, line 197: As this stands, we accept `Anthropic_Version` for OpenAI listeners. We should use the `Protocol` to throw errors on options that only make sense for the provider they are meant to.

ingesters/llm_ingester/protocol/anthropic/stream.go
```diff
@@ -0,0 +1,277 @@
...
+				r.id = e.Message.ID
+			}
+			if r.model == "" {
+				r.model = e.Message.Model
+			}
+			if u := e.Message.Usage; u != nil {
+				r.inputTokens = u.InputTokens
+				r.cacheRead = u.CacheReadInput
+				r.cacheCreation = u.CacheCreationInput
+				if u.OutputTokens > 0 {
```
daniel-moreno-levy-gravwell, line 156: I went down a rabbit hole about `OutputToken`. The Claude API sets this to 1 always, so this if is dead, it will always pass. We can't control Claude lying, but we can correct for it. Instead, make this guard against `> 1` (or don't ever read it at all in `message_start`), so output tokens stays 0 for the initial message.

daniel-moreno-levy-gravwell (approved):
This round of comments are from the Claude Code review session I ran in parallel to mine. I've verified them and expressed them in my own words to cut down on the noise and verbosity.

Here are some it found OUTSIDE the diff:

---

1. In `ingest.go`: 
https://github.com/gravwell/gravwell/blob/dd69fc792327fde879f7b5e1ccd179e828f54dbb/ingesters/llm_ingester/ingest.go#L87-L102

A message with no assistant message right after a system message will duplicate the system prompt (`tailStart` returns zero). IE:

```
request.system_message  data="PRETEND THIS IS A 25KB CLAUDE CODE SYSTEM PROMPT"
request.user_message    data="hello"
```

Will produce:

```
entry: request.system_message  data="PRETEND THIS IS A 25KB CLAUDE CODE SYSTEM PROMPT"
entry: request.system_message  data="PRETEND THIS IS A 25KB CLAUDE CODE SYSTEM PROMPT"
entry: request.user_message    data="hello"
```

I asked claude to create a test to replicate the issue: https://gist.github.com/daniel-moreno-levy-gravwell/914e4da7858003081df4c03fb1b78fe9

The fix is to skip the `EventSystemMessage` in the tailStart for loop, newSession already captures it.

ingesters/llm_ingester/README.md
david-fritz-gravwell: done

PR #2688, opened 2026-08-21
task(issues#2733): force public types to marshal to zero array/map instead of null

rory-landau-gravwell:
This PR addresses closes https://github.com/gravwell/issues/issues/2735.

## This PR adds custom marshalers to most public types

This is a rather goofy, temporary solution to hold us over until JSON v2 stabilizes and 6.0 is out the door. I fully intend to throw it all away when given an opportunity (https://github.com/gravwell/issues/issues/2765).

You can read more about the issues with fixing this in v1 in the above-linked PR. Long story short, none of the possible solutions are good and even this one has a lot of footguns.

I drew up a couple draft implementations and then *had a bot replicate the implementations on public types with slices/maps*.

This should allow frontend to stop dealing with `null` arrays/maps (but there may be anon structs or structs that live only in backend still floating around).

## PR Tasks

## Reviewer Tasks


client/types/commonfields.go
```diff
@@ -93,6 +94,19 @@ type CommonFields struct {
 	Can Actions
 }
 
+// MakeNilSlices returns a copy of cf with nil slice field (currently just Labels)
+// replaced by an empty, non-nil slice s.t. it marshals as "[]".
+func (cf CommonFields) MakeNilSlices() CommonFields {
```
kris-watts-gravwell, line 99: implement this in a marshaller rather than copying back.

Doing it this way means everyone that ever uses this type has to manually do this where doing it in a marshaller means everyone gets it for free.

client/types/cbac.go
```diff
@@ -1044,6 +1087,22 @@ type TokenFull struct {
 	Value string
 }
 
+// MarshalJSON here marshals into an anon struct instead of the usual dummyTokenFull type.
+// Long story short, a naive dummy alias would silently call Token's MarshalJSON and drop the Value field entirely.
+// Take my word for it.
+func (t TokenFull) MarshalJSON() ([]byte, error) {
+	type dummyToken Token
+	t.CommonFields = t.CommonFields.MakeNilSlices()
```
kris-watts-gravwell, line 1095: do this in the marshaller for CommonFields, not like this.

client/types/actionables.go
```diff
@@ -27,6 +29,13 @@ type Actionable struct {
 	Disabled bool
 }
 
+// MarshalJSON ensures slices and maps marshal as "[]"/"{}" instead of "null".
+func (a Actionable) MarshalJSON() ([]byte, error) {
+	type dummyActionable Actionable
+	a.CommonFields = a.CommonFields.MakeNilSlices()
```
kris-watts-gravwell, line 35: implement a marshaller for common fields instead of trying to remember to do a re-assignment on every thing.

client/types/actionables.go
```diff
@@ -91,3 +100,11 @@ type ActionableListResponse struct {
 	BaseListResponse
 	Results []Actionable
 }
+
+// MarshalJSON ensures slices and maps marshal as "[]"/"{}" instead of "null".
+func (a ActionableListResponse) MarshalJSON() ([]byte, error) {
+	type dummyActionableListResponse ActionableListResponse
+	a.Results = nonNilSlice(a.Results)
+	a.BaseListResponse = a.BaseListResponse.MakeNilSlices()
```
kris-watts-gravwell, line 108: same for base list response, do this in the marshaller, not in a reassignment every where you use it.

client/types/autoextract.go
```diff
@@ -36,6 +37,14 @@ type AX struct {
 	Tags   []string `toml:"tags"`
 }
 
+// MarshalJSON ensures slices and maps marshal as "[]"/"{}" instead of "null".
+func (dc AX) MarshalJSON() ([]byte, error) {
+	type dummyAX AX
+	dc.CommonFields = dc.CommonFields.MakeNilSlices()
```
kris-watts-gravwell, line 43: same

kris-watts-gravwell (changes requested):
when you have complete types implement a marshaller instead of the function and assign back

client/types/commonfields.go
rory-landau-gravwell: I'm a little shaky on (un)marshaler precedence, but I was under the impression we want to explicitly avoid putting a marshaler on our embedded type. Doing so means it will be promoted to the parent type if the parent type has no marshaler. Any parent types without its own MarshalJSON will call CF's, which will drop the parent types fields. ex: https://go.dev/play/p/Upes2jb5oz-

It also disallows us from doing the dummyX trick I am using to avoid infinite recursion. The new `dummy` type doesn't inherit the methods of its base type, but *would* inherit the methods of its embedded fields (CF). Every type would then have to use the TokenFull method: create an anonymous copy of its own type AND include a dummy wrapper for the embedded type to ensure parent fields are actually included when its embedded type (Token) is marshaled.

rory-landau-gravwell: Updated example: https://go.dev/play/p/4BMnd8qIH4c

kris-watts-gravwell: isn't that what you want?  if the goal is to make sure types (including embedded) do not ship "null" then I would argue you WANT that to happen everywhere, including the dummy wrapper types with marshallers.

rory-landau-gravwell: Implementing Marshal on CF would do so everywhere, but it would come with significant consequences:
it would dropping fields, which is untenable, or require we use anonymous structs for each type, like TokenFull must.

Assume CF.MarshalJSON(). The File type has no slices that need coercion, so it doesn't need to overload .MarshalJSON(). Because File.MarshalJSON doesn't exist, the embedded CF.MarshalJSON will take priority and drop all of its parent's fields.

The only way around this afaik would be to implement:
```go
func (f File) MarshalJSON() ([]byte, error){
    return json.Marshal(struct{
        ....
}{
        CommonFields: json.Marshal(f.CF.Marshal),
        Hash: f.Hash,
        Size: f.Size,
        ...
})
}

```

rory-landau-gravwell: I guess I have to add MarshalJSON to all types to call MakeNilSlices() anyways so that doesn't matter so much.

Ex: files will still require:
```go
func (f File) MarshalJSON() ([]byte, error){
    if f.Labels == nil {
        f.Labels = []string{}
    }
    return json.Marshal(f)
}
```

So I suppose the discussion is more around if we want to do anonymous structs and guarantee CF won't be forgotten or use a helper func + the raw struct and potentially forget to call MakeNilSlice()

rory-landau-gravwell: I'm making a bot do it either way until we adopt JSONv2 so I guess it doesn't really matter to me whether we do:

```go
func (f File) MarshalJSON() ([]byte, error){
    type dummy File
    f.CF = f.CF.MakeNilSlices()
    return json.Marshal(dummy(f))
}
```

or anonymous wrappers

```go
func (f File) MarshalJSON() ([]byte, error){
    return json.Marshal(struct{
        ....
}{
        CommonFields: json.Marshal(f.CF.Marshal),
        Hash: f.Hash,
        Size: f.Size,
        ...
})
}
```

kris-watts-gravwell: ok, this is a giant PIA.....   Lets leave as is until post 6.0

if the v2 marshaller has less blast radius than this then maybe we push on it.

kris-watts-gravwell:
Additional comments:

 items 1 nad 2 i think have to be fixed.

Item 3 is tough to swallow, that is a real slowdown on a hot path that will hurt us because of dumb browser JSON parsers that don't handle the spec...

Item 5 confuses me a bit, that should break everything.

## Review notes: null-safe marshalers

Verified against the PR head by diffing the marshaled JSON of every exported type in
`client/types` (360 types) against the merge base. CI is green and `go test ./client/...`
passes, so none of these are caught by the existing suite.

---

### 1. `IdxStats.IndexStats` silently vanishes from responses

`client/types/stats.go` deletes `func (m *IdxStats) MarshalJSON()` and no replacement is
added anywhere. The other three marshalers removed from that file (`StatSet`,
`IndexManagerStats`, `SearchModuleStatsUpdate`) all got value-receiver replacements —
`IdxStats` was missed.

Because the field is tagged `json:",omitempty"`, the plain encoder now drops it instead of
emitting `[]`:

```
old: {"UUID":"00000000-0000-0000-0000-000000000000","IndexStats":[]}
new: {"UUID":"00000000-0000-0000-0000-000000000000"}
```

This is the opposite of the PR's goal, and it is on a live API path —
`Client.GetIndexStats()` returns `map[string]types.IdxStats` (`client/staticActions.go:472`).

The `omitempty` tag also means `TestMarshalingNilCreatesEmpty` cannot catch it, since the
test skips `omitempty` fields by design.

---

### 2. `ListAllResponse.MarshalJSON` mutates the caller's data and races

`client/types/commonfields.go` — `l` is a value copy, but `l.Results` shares the caller's
backing array, so this writes through to the caller:

```go
l.Results = nonNilSlice(l.Results)
for i := range l.Results {
	l.Results[i] = l.Results[i].MakeNilSlices()   // <-- mutates caller's slice
}
```

Observed:

```
before marshal: Results[0].Labels == nil ? true
after  marshal: Results[0].Labels == nil ? false
```

Two goroutines marshaling the same response is a data race; `go test -race` reports it
immediately. Every other marshaler in this PR is side-effect-free — this is the only loop
that writes into an element. Needs a copied slice:

```go
res := make([]CommonFields, len(l.Results))
for i := range l.Results {
	res[i] = l.Results[i].MakeNilSlices()
}
l.Results = res
```

---

### 3. Adding `MarshalJSON` to `SearchEntry` costs ~4x on the hottest JSON path

Marshaling 1000 entries with representative data (73-byte `Data`, 2 enumerated pairs):

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| base | 399,573 | 383,122 | 2,003 |
| PR | 1,531,969 | 1,048,284 | 8,010 |

3.8x slower, 4x the allocations. Every raw/text/hex render, search download, and alert
payload pays this per entry, and it buys `"Enumerated":[]` instead of `null`.

Same mechanism applies to the other per-entry types that newly gained a marshaler:
`StringTagEntry`, `Chartable`, `Element` (per table cell), `RawEnumeratedValue`,
`ResultsGraphNode`, `ResultsGraphLink`.

Worth noting the reverse also happened: `TableRow` got *faster* (512µs → 395µs), because
the dummy-conversion replaced an existing anon-struct marshaler. The regression comes
specifically from adding a marshaler where none existed, so the hot per-entry types may be
worth exempting.

---

### 4. `gofmt` fails

`client/types/users.go:448` — trailing whitespace after the `UserWithCBAC` marshaler's
closing brace.

```
$ gofmt -l client/types/
client/types/users.go
```

`go build` and `go vet` are clean, and CI does not appear to check formatting.

---

### 5. 13 `[]byte` fields flipped `null` → `""`

`[]byte` marshals as a base64 **string**, not an array, so this is not the null-array
problem being solved:

| Type | Field |
|---|---|
| `CapabilitySet` | `Grants` |
| `LicenseInfo` | `Hash`, `Metadata` |
| `MFATOTPSetupResponse` | `QRCode` |
| `TokenSigningKey` | `Key` |
| `SearchEntry` | `Data` |
| `StringTagEntry` | `Data` |
| `RawEnumeratedValue` | `Data` |
| `Thing` | `Contents` |
| `UserMailAttachment` | `Content` |
| `FlowResults` | `DebugOutput` |
| `ScheduledScriptResults` | `DebugOutput` |

The new test explicitly skips these (`// []byte should be "", not "[]"`), so the change is
both out of scope for the issue and unverified. Either drop the `nonNilSlice` calls on
`[]byte` fields, or assert the `""` behavior in the test if it is intended.

rory-landau-gravwell:
1, 2, and 4 are oversights; I will fix those.

3 sucks. And, even worse, it is kind of a tangential result; the main targets are the assets. The crux of the issue is that I am `make`ing nil slices, which costs allocation, when all we really need is to print `[]` instead of `null`; I have an idea on a faster method, which may be worth the tradeoff for items in in the hotpath (like `SearchEntry`). I'm loathe to spend too much time on it when v2 does what we want natively, but let me poke around a bit more.

5 is a consequence of considering []byte equivalent to strings in JSON. P sure it isn't breaking because we currently treat zeros and nulls as the same. I can flip it back to `null`. I just wasn't sure that `[]` makes sense in this case.

rory-landau-gravwell:
Alright, we are switching to JSON/v2 and throwing all of this out (thank goodness). I'm killing this PR and reopen one when v2 is ready

PR #2692, opened 2026-08-21
fix: user put using base user instead of update user

nic-osullivan-gravwell:
This PR addresses no issue, but done as part of https://github.com/gravwell/issues/issues/2647 as I was using these types for the e2e tests. 

## This PR proposes...

When updating a user we were sending the base User, instead of UpdateUser. This is mostly an issue when working with the DefaultSearchGroups as they use different structure (an array of Groups vs a array of gids)

## PR Tasks

## Reviewer Tasks

### DIFF
--- a/client/users.go
+++ b/client/users.go
@@ -75,7 +75,7 @@ func (c *Client) CreateUser(m types.AddUser) (result types.User, err error) {
 
 // UpdateUser (admin-only) modifies an existing user.
 func (c *Client) UpdateUser(m types.User) error {
-	return c.putStaticURL(usersInfoUrl(m.ID), m)
+	return c.putStaticURL(usersInfoUrl(m.ID), m.ForUpdate())
 }
 
 // UpdateUserInfo changes basic information about the specified user.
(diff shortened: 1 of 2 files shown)


client/users.go
```diff
@@ -75,7 +75,7 @@ func (c *Client) CreateUser(m types.AddUser) (result types.User, err error) {
 
 // UpdateUser (admin-only) modifies an existing user.
 func (c *Client) UpdateUser(m types.User) error {
-	return c.putStaticURL(usersInfoUrl(m.ID), m)
+	return c.putStaticURL(usersInfoUrl(m.ID), m.ForUpdate())
```
nic-osullivan-gravwell, line 78: I left the func to use `types.User` as we do actually need the ID for the method and this was probably what anyone was working with anyways.

rory-landau-gravwell (approved):
LGTM

client/users.go
rory-landau-gravwell: Good call

PR #2702, opened 2026-08-25
refactor(2768): Duration is now int64, not string

tyler-elton-gravwell:
This PR **PARTIALLY** addresses https://github.com/gravwell/issues/issues/2768

## This PR proposes...

- removing the custom `MarshalJSON`/`UnmarshalJSON` methods on `SearchInfo` to allow `Duration` to just be an `int64`, not `string`
- clean up linter warnings in this file
- update test

## PR Tasks

## Reviewer Tasks

### DIFF
--- a/client/types/search.go
+++ b/client/types/search.go
@@ -711,38 +711,6 @@ func (l LaunchResponse) MarshalJSON() ([]byte, error) {
 	})
 }
 
-func (si SearchInfo) MarshalJSON() ([]byte, error) {
-	type alias SearchInfo
-	return json.Marshal(struct {
-		alias
-		Duration string `json:",omitempty"`
-	}{
-		alias:    alias(si),
-		Duration: si.Duration.String(),
-	})
-}
-
-func (si *SearchInfo) UnmarshalJSON(data []byte) error {
-	type aalias SearchInfo
-	type alias struct {
-		aalias
-		Duration string `json:",omitempty"`
-	}
-	var v alias
-	if err := json.Unmarshal(data, &v); err != nil {
-		return err
-	}
-	*si = SearchInfo(v.aalias)
-	if len(v.Duration) > 0 {
-		dur, err := time.ParseDuration(v.Duration)
-		if err != nil {
-			return err
-		}
-		si.Duration = dur
-	}
-	return nil
-}
-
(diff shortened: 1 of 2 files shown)


client/types/search.go
```diff
@@ -811,7 +778,7 @@ func (p SaveSearchPatch) MergeLaunchInfo(li *SearchLaunchInfo) (changed bool) {
 	// this marks the search as saved and sets an expiration.  A user gets the alert, clicks the search,
 	// sees that the results should be kept and clicks save; this action means the user wants to kill
 	// the timer, so we wipe the expiration.
-	changed = !li.Expires.Equal(p.Expires)
```
rory-landau-gravwell, line 814: "we ALWAYS take the expiration" is no longer true if we OR it, right? Was this change intentional?

rory-landau-gravwell (commented):
Changes look good, but I want to make sure this was intentional.

client/types/search.go
rory-landau-gravwell: Tyler pointed out that I missed the above assignments as ineffectual. All good.

rory-landau-gravwell (approved):
LGTM

client/types/search.go
tyler-elton-gravwell: pasting what we talked about in chat for documentation:
I was trying to fix some linting warnings, and that one had a "ineffectual assignment to changed" warning. 
```go
    if p.Method != `` {
        li.Method = p.Method
        changed = true
    }
    if p.Reference != `` {
        li.Reference = p.Reference
        changed = true
    }
    // the expiration is special, we ALWAYS take the experation, even if its the zero value.
    // We do this so that a human clicking save on the GUI ALWAYS blows the search expiration.
    // The use case is that a scheduled search gets results and fires an alert, the alert logic
    // this marks the search as saved and sets an expiration.  A user gets the alert, clicks the search,
    // sees that the results should be kept and clicks save; this action means the user wants to kill
    // the timer, so we wipe the expiration.
    changed = !li.Expires.Equal(p.Expires)
    li.Expires = p.Expires
```
to me, that comment really only justifies the li.Expires = p.Expires line, not the unconditional changed override

PR #2710, opened 2026-08-27
pin staticcheck version

ashnwade:
This PR addresses no issue

## This PR proposes...

- pinning the staticcheck version to match backend

## PR Tasks

## Reviewer Tasks

### DIFF
--- a/.github/workflows/run_local_build_checks.sh
+++ b/.github/workflows/run_local_build_checks.sh
@@ -8,7 +8,7 @@ go mod tidy
 go mod download
 go mod verify
 go install golang.org/x/vuln/cmd/govulncheck@latest
-go install honnef.co/go/tools/cmd/staticcheck@ff63afafc529279f454e02f1d060210bd4263951
+go install honnef.co/go/tools/cmd/staticcheck@2026.2.1
 
 echo "Running go vet"
 	go vet ./generators/ipgen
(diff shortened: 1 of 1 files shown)


rory-landau-gravwell (changes requested):
It looks like we recently updated the pin again to point to 2026.2.1: https://github.com/gravwell/backend/commit/0e33270059f49dd18acd1ed15e1370fbdd0ccaff

rory-landau-gravwell:
Wait, that is next-major, this is for next-minor. Do we want to have next-major and next-minor differ?

EDIT: if this is intentional, lmk and I'll approve

kris-watts-gravwell:
the release version 2026.2.1 appears to cover everything, is there a reason to pin against a specific commit?

ashnwade:
We discussed on Tuesday at the Dev Sync that we should pin the staticcheck version like we do many of our other dependencies to avoid situations like we had last week where we had false positives and were not able to merge any PRs. We pinned to this specific version on backend. This PR targets next-patch and you can see that the same version is already pinned in next-patch in the internal repo (check next-patch specifically).

ashnwade:
We updated major to 2026.2.1 because it adds support for go 1.27 (which is only on the next-major branch currently) and it resolves the false positive that previously gave problems that forced us to pin the commit seen here.

ashnwade:
I do think it is good for the two repos to match

ashnwade:
if we want to update to 2026.2.1 on next-patch, I can do that, but we need to make the same change on backend then

rory-landau-gravwell (approved):
LGTM.

So should I submit a PR for backend#next-patch?

ashnwade:
I already submitted one

PR #2718, opened 2026-08-28
TOY: Optional[T] with Patch types

rory-landau-gravwell:
This PR prototypes https://github.com/gravwell/issues/issues/1467.

>[!IMPORTANT]
>This PR is NOT intended to be merged; it breaks nearly all update functionality! It is to serve as a toy other backenders can experiment with before we proceed with a full rollout. Assuming it is satisfactory, I will kill this PR and create a new one when the rollout is complete.

Sister PR: https://github.com/gravwell/backend/pull/4455

I did not create a sister api-docs prototype.

## This PR implements PATCH semantics for Macros

**The code, especially the backend code, is messy. Evaluate the syntax and ergonomics; I already have plans to clean up, optimize, and consolidate the hell out of these changes/files if we move forward.**

- Introduces Optional[T] type. This is the crux of this implementation; Patch/updates fields are wrapped in `Optional[T]` such that they inherit the properties required to patch their parent type's fields.
  - You can see it in action in `gravwell/client/types/marshalers_v2_test.go` and end-to-end at `backend/test/env/single_node/tests/patch.go`
- Introduces MacroPatch and CFPatch, each of which only contains fields a user is allowed to update.
- Updates Client.UpdateMacro() to send as PATCH.
  - Introduces a new func `Client.PATCH()` that would serve as the driver function for PATCH calls.

This PR is built on JSON/v2 and the toy assumes v2 across the board.

### Patching private types

The toy patches the pubtypes.*Patch directly onto the private type. It felt silly to convert the existing private type to a public type, patch the public type, and then convert back.

## Open Question: slice and map patching

Over all, the design we landed on feels great. Clean, easy to grok, easy to maintain, relatively ergonomic, and closer to API docs than what we were doing before.

However, there is still one open question: how do we handle maps/slices? 

- Do patches still clobber them (if Set)?
- Do we need to introduce Add/Clobber fields to Patch types (ex: instead of `CFP.Labels`, we have `CFP.LabelsAppend` and `CFP.LabelsReplace`)? 
- Do they need their own version of `Optional[T]` that includes an enum for how to handle it (ex: `OptionalBox[T]` has `.value`, `.set`, and `.behavior` where behavior is an enumerated `append`, `remove`, and `replace`)?

This tangentially relates to nested structs too. How deep do we place `Optional[T]`? `Optional[pubtypes.ACL]`? `pubtypes.ACL{GIDs Optional[[]int32], Global Optional[bool]}`?

## PR Tasks

  - I have included some unit tests to ensure proper behavior, but am not planning to make this toy pass tests.

## Reviewer Tasks

### DIFF
--- a/client/macros.go
+++ b/client/macros.go
@@ -94,7 +94,7 @@ func (c *Client) PATCH[T types.PatchType, K any](url string, data T) (patched K,
 	body, err := json.Marshal(data, json.OmitZeroStructFields(true))
 	if err != nil {
 		return patched, err
-	} else if string(body) == "{}" { // if this marshaled to no data, throw away the request
+	} else if body == nil || string(body) == "{}" { // if this marshaled to no data, throw away the request
 		return patched, ErrEmptyPatch
 	}
 
--- a/client/macros.go
+++ b/client/macros.go
@@ -8,7 +8,15 @@
 
 package client
 
-import "github.com/gravwell/gravwell/v4/client/types"
+import (
+	"bytes"
+	"encoding/json/v2"
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/gravwell/gravwell/v4/client/types"
+)
 
 // ListMacros returns all macros accessible to the current user.
 func (c *Client) ListMacros(opts *types.QueryOptions) (ret types.MacroListResponse, err error) {
@@ -64,9 +72,67 @@ func (c *Client) CreateMacro(m types.Macro) (result types.Macro, err error) {
 	return
 }
 
-// UpdateMacro modifies an existing macro.
-func (c *Client) UpdateMacro(m types.Macro) error {
-	return c.putStaticURL(macroUrl(m.ID), m)
+var (
+	ErrEmptyPatch  = errors.New("empty PATCHs are ineffectual")
+	ErrNilResponse = errors.New("nil response")
+	ErrNilID       = errors.New("ID must not be nil")
+)
+
+// UpdateMacro modifies an existing macro and returns complete, updated struct.
+func (c *Client) UpdateMacro(ID string, p types.MacroPatch) (updated types.Macro, _ error) {
+	if ID == "" {
+		return types.Macro{}, ErrNilID
+	}
+	return c.PATCH[types.MacroPatch, types.Macro](macroUrl(ID), p)
+}
+
+// PATCH marshals data to JSON and submits a PATCH request against the target URL.
+//
+// TODO we can probably consolidate a lot of the methodStaticPushURLs by taking a method param in
+// wrapper calls and converting this to a driver func
+func (c *Client) PATCH[T types.PatchType, K any](url string, data T) (patched K, _ error) {
+	body, err := json.Marshal(data, json.OmitZeroStructFields(true))
+	if err != nil {
+		return patched, err
+	} else if string(body) == "{}" { // if this marshaled to no data, throw away the request
+		return patched, ErrEmptyPatch
+	}
+
+	uri := fmt.Sprintf("%s://%s%s", c.httpScheme, c.server, url)
+	req, err := http.NewRequest(http.MethodPatch, uri, bytes.NewBuffer(body))
+	if err != nil {
+		return patched, err
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	c.hm.populateRequest(req.Header) // add in the headers
+
+	// add in any queries like ?admin=true
+	if req.URL.RawQuery, err = c.qm.appendEncode(req.URL.RawQuery); err != nil {
+		return patched, err
+	}
+
+	c.objLog.Log("WEB REQ "+http.MethodPatch, url, data)
+	resp, err := c.clnt.Do(req)
+	if err != nil {
+		c.objLog.Log("WEB "+http.MethodPatch+" Error "+err.Error(), url, nil)
+		return patched, err
+	}
+	if resp == nil {
+		return patched, ErrNilResponse
+	}
+	if resp.StatusCode != http.StatusOK {
+		c.objLog.Log("WEB "+http.MethodPatch, url+" "+resp.Status, nil)
+		return patched, aliasResponseError(c, resp)
+	}
+	defer drainResponse(resp)
+
+	if err := json.UnmarshalRead(resp.Body, &patched); err != nil {
+		return patched, err
+	}
+
+	c.objLog.Log("WEB RECV", url, patched)
+	return patched, nil
 }
 
 // CleanupMacros (admin-only) purges all deleted macros for all users.
--- a/client/types/optional.go
+++ b/client/types/optional.go
@@ -1,6 +1,12 @@
 package types
 
-import "encoding/json"
+import (
+	"encoding/json"
+)
+
+type PatchType interface {
+	MacroPatch
+}
 
 // An Optional type represents a field which may be unset during an update, preserving its prior value.
 // If !Optional.IsSet(), this field will be omitted from a JSON marshal of the type.
@@ -15,7 +21,7 @@ func NewOptional[T any](v T) Optional[T] {
 }
 
 // Set installs the given value and marks it as valid to include when marshaling.
-func (o Optional[T]) Set(v T) {
+func (o *Optional[T]) Set(v T) {
 	o.value = v
 	o.set = true
 }
@@ -37,13 +43,22 @@ func (o Optional[T]) IsSet() bool {
 }
 
 // Unset marks this field s.t. it will be skipped when marshaling.
-func (o Optional[T]) Unset() {
+func (o *Optional[T]) Unset() {
 	o.set = false
 }
 
+func (o Optional[T]) IsZero() bool {
+	return !o.IsSet()
+}
+
+// MarshalJSON causes optional to print the set value or the zero value.
+//
+// NOTE(rlandau): as of go1.27.0, a field cannot skip itself with MarshalJSON.
+// Instead, we use IsZero and the OmitZeroStructFields option.
 func (o Optional[T]) MarshalJSON() ([]byte, error) {
 	if !o.IsSet() {
-		return nil, nil
+		var zero T
+		return json.Marshal(zero)
 	}
 	return json.Marshal(o.value)
 }
--- a/client/types/optional.go
+++ b/client/types/optional.go
@@ -47,11 +47,16 @@ func (o *Optional[T]) Unset() {
 	o.set = false
 }
 
+// IsZero returns true iff !o.Set.
+//
+// This allows marshalers to omit unset values while still including zero values
+// (assuming json.OmitZeroStructFields(true) or `json:",omitzero"` is set).
 func (o Optional[T]) IsZero() bool {
 	return !o.IsSet()
 }
 
-// MarshalJSON causes optional to print the set value or the zero value.
+// MarshalJSON causes optional to always marshal to a safe value.
+// If !o.IsSet(), T zero will be used.
 //
 // NOTE(rlandau): as of go1.27.0, a field cannot skip itself with MarshalJSON.
 // Instead, we use IsZero and the OmitZeroStructFields option.
--- a/client/types/optional.go
+++ b/client/types/optional.go
@@ -1,7 +1,7 @@
 package types
 
 import (
-	"encoding/json"
+	"encoding/json/v2"
 )
 
 type PatchType interface {
--- a/client/types/optional.go
+++ b/client/types/optional.go
@@ -67,3 +67,12 @@ func (o Optional[T]) MarshalJSON() ([]byte, error) {
 	}
 	return json.Marshal(o.value)
 }
+
+// UnmarshalJSON decodes the given data into o's value and marks it as set.
+func (o *Optional[T]) UnmarshalJSON(data []byte) error {
+	if err := json.Unmarshal(data, &o.value); err != nil {
+		return err
+	}
+	o.set = true
+	return nil
+}
--- a/client/types/optional.go
+++ b/client/types/optional.go
@@ -1,0 +1,49 @@
+package types
+
+import "encoding/json"
+
+// An Optional type represents a field which may be unset during an update, preserving its prior value.
+// If !Optional.IsSet(), this field will be omitted from a JSON marshal of the type.
+type Optional[T any] struct {
+	value T
+	set   bool
+}
+
+// NewOptional returns a set field with value v.
+func NewOptional[T any](v T) Optional[T] {
+	return Optional[T]{value: v, set: true}
+}
+
+// Set installs the given value and marks it as valid to include when marshaling.
+func (o Optional[T]) Set(v T) {
+	o.value = v
+	o.set = true
+}
+
+// Value fetches the value contained within the Optional.
+// It is intended to prevent nil-derefs and is always safe to call.
+// If !o.IsSet(), T's zero value will be returned
+func (o Optional[T]) Value() T {
+	if !o.IsSet() {
+		var zero T
+		return zero
+	}
+	return o.value
+}
+
+// IsSet states if this field has a value and thus will be included in the JSON representation.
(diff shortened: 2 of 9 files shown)


client/types/optional.go
```diff
@@ -0,0 +1,78 @@
...
+	o.set = true
+}
+
+// Value fetches the value contained within the Optional.
+// It is intended to prevent nil-derefs and is always safe to call.
+// If !o.IsSet(), T's zero value will be returned
+func (o Optional[T]) Value() T {
+	if !o.IsSet() {
+		var zero T
+		return zero
```
nic-osullivan-gravwell, line 35: nit: `new(T)` _I think_. That might be a pointer, but I don't think it is.

client/types/optional.go
```diff
@@ -0,0 +1,78 @@
...
+func (o Optional[T]) IsZero() bool {
+	return !o.IsSet()
+}
+
+// MarshalJSON causes optional to always marshal to a safe value.
+// If !o.IsSet(), T zero will be used.
+//
+// NOTE(rlandau): as of go1.27.0, a field cannot skip itself with MarshalJSON.
+// Instead, we use IsZero and the OmitZeroStructFields option.
+func (o Optional[T]) MarshalJSON() ([]byte, error) {
```
nic-osullivan-gravwell, line 63: Shouldn't this be a pointer receiver? 

Is is standard to have Marshal on the struct and unmarshal on the pointer?

client/macros.go
```diff
@@ -64,9 +72,67 @@ func (c *Client) CreateMacro(m types.Macro) (result types.Macro, err error) {
...
+	}
+	return c.PATCH[types.MacroPatch, types.Macro](macroUrl(ID), p)
+}
+
+// PATCH marshals data to JSON and submits a PATCH request against the target URL.
+//
+// TODO we can probably consolidate a lot of the methodStaticPushURLs by taking a method param in
+// wrapper calls and converting this to a driver func
+func (c *Client) PATCH[T types.PatchType, K any](url string, data T) (patched K, _ error) {
+	body, err := json.Marshal(data, json.OmitZeroStructFields(true))
```
nic-osullivan-gravwell, line 94: Do we have to specify `json.OmitZeroStructFields(true)` at all? This should be on a per type basis (via `json:",omitzero"`). Otherwise this hard prevents that from being an option. This is also the only thing "forcing" us to json v2 explicitly, as v1 support the `omitzero` tag.

client/macros.go
```diff
@@ -64,9 +72,67 @@ func (c *Client) CreateMacro(m types.Macro) (result types.Macro, err error) {
...
+		return types.Macro{}, ErrNilID
+	}
+	return c.PATCH[types.MacroPatch, types.Macro](macroUrl(ID), p)
+}
+
+// PATCH marshals data to JSON and submits a PATCH request against the target URL.
+//
+// TODO we can probably consolidate a lot of the methodStaticPushURLs by taking a method param in
+// wrapper calls and converting this to a driver func
+func (c *Client) PATCH[T types.PatchType, K any](url string, data T) (patched K, _ error) {
```
nic-osullivan-gravwell, line 93: nit: `Patch` or `patch`, don't need the code to yell lol

client/macros.go
```diff
@@ -64,9 +72,67 @@ func (c *Client) CreateMacro(m types.Macro) (result types.Macro, err error) {
...
+	ErrNilResponse = errors.New("nil response")
+	ErrNilID       = errors.New("ID must not be nil")
+)
+
+// UpdateMacro modifies an existing macro and returns complete, updated struct.
+func (c *Client) UpdateMacro(ID string, p types.MacroPatch) (updated types.Macro, _ error) {
+	if ID == "" {
+		return types.Macro{}, ErrNilID
+	}
+	return c.PATCH[types.MacroPatch, types.Macro](macroUrl(ID), p)
```
nic-osullivan-gravwell, line 86: nit, maybe: I'm not sure you have to specify types here, but the return might make that funky. Oh and maybe the new stuff is less intuitive on the types right.

client/types/optional.go
```diff
@@ -0,0 +1,78 @@
+package types
+
+import (
+	"encoding/json/v2"
+)
+
+type PatchType interface {
+	MacroPatch
```
nic-osullivan-gravwell, line 8: nit: flip these? `MacroPatch` should embed `PatchType` and this should embed `CommonFieldsPatch`

client/types/search.go
```diff
@@ -679,6 +679,13 @@ func (si SearchInfo) StorageSize() int64 {
...
 const AllowedMacroChars = "ABCDCEFGHIJKLMNOPQRSTUVWXYZ1234567890_-"
 
+// MacroPatch is the type used to request an update to an existing Macro.
+type MacroPatch struct {
+	CommonFieldsPatch
+	Expansion Optional[string]
+}
+
+// Macro is the type used to Get and Create Macros.
 type Macro struct {
```
nic-osullivan-gravwell, line 689: Add a `ForUpdate()` method that returns a `MacroPatch` and I think I'm in.

nic-osullivan-gravwell (commented):
Honestly I'm pretty good with this. 

Only [this comment](https://github.com/gravwell/gravwell/pull/2718/changes#r3883644645) is about the approach abstractly, everything else are nits or implementation details. 

I like the new generic method (other than it shouting), seems simpler than the methodDoSomethingWeirdName

client/types/optional.go
rory-landau-gravwell: `new(T)` will return *T, whereas the return value expects T. Ex: T of []string would return *[]string

image

client/types/optional.go
rory-landau-gravwell: That's a good question and a good catch. If anything, I'd want to move Unmarshal to a value method as it makes no changes (and thus doesn't require a pointer).

nic-osullivan-gravwell: Unmarshal _has_ to be a pointer as it does modify in place. Otherwise nothing would ever be marshaled I don't think.

client/types/optional.go
nic-osullivan-gravwell: boooooooooo

client/macros.go
rory-landau-gravwell: Erm, we definitely can do it on a field basis, but I can't think of a circumstance where we don't want the option applied to PATCHes and rather than do it per field we can do it as a funnel.

Even if we apply the option to json.Marshal() calls outside of PATCH, it has no bearing on unmarshaling and I don't believe it would alter anything. Hence why I wanted to guarantee it is invoked.

v1 *does* support the omitzero tag, but I'm under the (possibly mistaken) understanding that it behaves marginally differently. v2 and v1 can already interoperate, but assuming v2 for mono keeps it clean and strict.

I'm open to changing it, but I can't think of a reason it would cause problems for us

client/types/optional.go
rory-landau-gravwell: Oh,duh, right, because we are unmarshaling into the receiver. I can poke around about Marshal. I just default to value receivers for safety, I guess

client/macros.go
rory-landau-gravwell: hah, fair enough. It is out of scope, but I would love to replace some of the obtuse `methodStatic*URL` with more dedicated endpoints.

client/macros.go
rory-landau-gravwell: I used generics here to guarantee the types coming in and out. Inbound could definitely be an `any`, though, given we just marshal it.

client/macros.go
nic-osullivan-gravwell: It'll just be harder later to go back and add the tags to all the types then as we go (if we need to change it). We'll also have the info we need when generating to add the tag so if we get there we can reflect the entire spec in the type. Basically we don't want the info on what makes it to the request in two different places. Having it in the tag on the patch types makes it clearer faster. 

My understanding is that it behaves the same, just v2 gives you the ability to have omitZero without the tag (as we are doing now).

nic-osullivan-gravwell: also without the tags the create and patch types _seem_ similar when looking at the types alone. That should be avoided.

client/types/optional.go
rory-landau-gravwell: Er, I'm using PatchType as a constraint, not a real type. Wouldn't you usually UNION types in a constraint? https://go.dev/doc/tutorial/generics#declare_type_constraint & https://pkg.go.dev/golang.org/x/exp/constraints (now, these are constraints on primitives so maybe it is different)

rory-landau-gravwell: Same as PublicType on the backend: 
image

client/macros.go
nic-osullivan-gravwell: if we keep it `patch` for now it's not part of the public api so we can do what we want when we want. I don't like those method either, very obtuse.

client/macros.go
nic-osullivan-gravwell: I meant more that I thought go could infer them base on the surrounding types and the vars passed in, rather than you specifying them explicitly in the `[]`.

client/types/optional.go
nic-osullivan-gravwell: Oh, I had no idea you could do a union type like that, good to know.

client/macros.go
rory-landau-gravwell: That is fair enough.

I got one more qualm: 
you don't think Optional[T] overriding IsZero is more confusing with a tag on it? We are handling omitting Optionals under the hood via their IsZero method. `Optional[string]` is omitted, but `Optional[string]("")` is not. `string json:",omitzero"` would be. I find that more confusing, but again, could be a me issue

client/macros.go
rory-landau-gravwell: AH. Understood.

`K` cannot be inferred because there is no type-intrinsic link between Macro and MacroPatch. And you must specify all type params or infer all type params; you can't do it halfway.

image

Which might be a good case for throwing away `K` and using `any`, but I like the extra type safety.

client/types/search.go
rory-landau-gravwell: As Macro doesn't have `Optional` fields, would you expect `.ForUpdate()` to `Set` fields with non-zero values or set all fields?

client/macros.go
nic-osullivan-gravwell: Yeah that was my maybe, the return was a little wacky.

I bet something like:
```go

var err error
var m types.Macro
m, err = c.PatchRequest(macroUrl(ID), p)
```
would be fine. since `m` has a type.

rory-landau-gravwell: Interesting.... It looks like it cannot infer the type even when given a concrete type:

image

image

(`ResponseT` being a renamed `K`)

I'd still push for genericizing the types to guarantee I/O types, especially if we are making this private. But I could be swayed. We can also reeval later as more types get installed

client/macros.go
nic-osullivan-gravwell: `IsZero` is only called when you have the explicit `omitzero` either via tag or options. Not just by its existence. 

I think `Optional[string] json:",omitzero"` is just more explicit and requires less context to understand. I will always prefer needing less context.

nic-osullivan-gravwell: To expand, I don't really need to know that Optional has a custom IsZero, or remember that. I have the info I need. I only need to understand go json encoding.

however `Optional[string]` does nothing to tell me how that will end up in the json. Does optional mean it's encoded as null or the zero value? 

With the current setup we'd need to understand all of Go, Optional (while small), and how we handle `PATCH` specifically in our client. It also makes it harder to encode these types on their own outside of the client.

rory-landau-gravwell: Fair enough, point taken

client/types/optional.go
```diff
@@ -0,0 +1,78 @@
...
+
+// MarshalJSON causes optional to always marshal to a safe value.
+// If !o.IsSet(), T zero will be used.
+//
+// NOTE(rlandau): as of go1.27.0, a field cannot skip itself with MarshalJSON.
+// Instead, we use IsZero and the OmitZeroStructFields option.
+func (o Optional[T]) MarshalJSON() ([]byte, error) {
+	if !o.IsSet() {
+		var zero T
+		return json.Marshal(zero)
```
daniel-moreno-levy-gravwell, line 66: Shouldn't this be `return []byte("null"), nil`? If the value is not set, that indicates the absence of a value, not the "zero" value. In JSON, the absence of a value is `null`.

If this marshalls to the same, then ignore this comment. As a counterargument to my suggestion, we would want `omitempty` to be respected, by not emitting anything at all for that case. My suggestion wouldn't do that, so it would have to look different.

daniel-moreno-levy-gravwell: I'm touching on the same themes as this conversation: https://github.com/gravwell/gravwell/pull/2718/changes#r3883644645

I would want an unset optional to marshall to `null`, unless `omitempty` or `omitzero` is set, which would skip it all together. I'm unsure on how to achieve that.

daniel-moreno-levy-gravwell (changes requested):

client/types/optional.go
tyler-elton-gravwell: I _believe_ `MarshalJSON` can be either a value or pointer receiver, but it changes how it can be called.

`func (t T) MarshalJSON` can be called via `json.Marshal(val)` or `json.Marshal(&val)` (supports both)
This one creates a copy of your struct which, depending on the size of the struct, could be inefficient.  

`func (t *T) MarshalJSON` can **ONLY** be called via `json.Marshal(&val)` (only supports an addressable pointer)
This one has a weird gotcha though.  If you pass the struct by value, place it inside a slice/map, or embed it in a non-pointer parent struct, `json.Marshal` will silently ignore your `MarshalJSON` method and fall back to default field encoding.

tyler-elton-gravwell (commented):
Left a comment in the discussion around `MarshalJSON`. Overall, this looks solid, great work!

client/types/optional.go
tyler-elton-gravwell: tl;dr:
Use value receivers unless your struct is large or you need to explicitly mutate/access internal pointer states during serialization (ex: `UnmarshalJSON`)

Don't use pointer receivers for types that might be embedded as fields in other structs, since that might lead to bugs where it would skip the custom marshal/unmarshal logic.

client/types/optional.go
rory-landau-gravwell: >IMO, I would want an unset optional to marshall to null, unless omitempty or omitzero is set, which would skip it all together. I'm unsure on how to achieve that.

JSON can theoretically support 3 states and I would very much prefer we only support 2. I do NOT want to differentiate between null and omitted, even though they are technically 2 different states. Specifically, JSON considers them differents states, but our PATCH semantics do not.

And if we are thinking about it from a UX perspective, someone sending a request via curl is probably going to omit the key as it is shorter.

So I'd much prefer we treat null and omitted as the same and then leverage `omitempty` to shrink our payloads, which is counter to printing null.

daniel-moreno-levy-gravwell: That is fair enough.  I think it should be called out in the go doc of MarshallJSON so people don't attempt "to fix it" in the future.

I agree with this rationale.

rory-landau-gravwell: I see your point about marshaling Optionals to nulls instead of zero values to prevent accidental clobbers, though.

https://github.com/gravwell/gravwell/pull/2718/changes#r3883644645 is actually about using tags over inline encoder options, not really about sending versus omitting a null.

I would argue that, in our case (especially because we are using `omitzero` tags), the
```go
if !o.IsSet() {
		var zero T
		return json.Marshal(zero)
```
case should never crop up.

client/types/search.go
nic-osullivan-gravwell: Oh right that would just create our problem again... Not having one is fine, maybe we should drop the ones we have ugh.

rory-landau-gravwell:
This is never intended to merge, so I am closing it. I will incorporate the suggestions into a full rollout once JSON/v2 is in place.

PR #2722, opened 2026-08-31
task(issues#2765): upgrade to JSONv2

rory-landau-gravwell:
This PR, in conjunction with its sister PR (pending), closes https://github.com/gravwell/issues/issues/2765.

## This PR upgrades mono to JSON/v2

This PR upgrades monorepo to JSONv2. In many places, this just means changing the import and passing in the options shim.

It also throws out `json.New{Encoder,Decoder}` as those are no longer necessary in v2.

### Options Shim

To ease the transition, this PR introduces a shim (`jsoncompat`) to keep our handling flexible while also taking advantage of the boons v2 brings.

This shim is a set of options:

- FormatDurationAsNano: mandatory. Our api contract obligates us to send time as an int64 of nanoseconds and JSONv2 errors otherwise.
- AllowInvalidUTF8: probably mandatory. We can't drop data so better we stick with the v1 default of mangling invalid UTF8.
- MatchCaseInsensitiveNames: advisable. While it adds ambiguity, we have data from other versions in production. Omitting this 

#### OmitEmptyWithLegacySemantics

The safest option is to include `OmitEmptyWithLegacySemantics(true)` and remove it later. However, I have opted to attempt to switch over to v2 tag semantics across the board so this isn't necessary. Riskier, but I don't like the idea of the same tag having differing semantics based on the options. We would have to deal with it later anyways, may as well benefit from it now.

As such, the shim does not include the option. This PR instead updates field tags s.t. our current behavior is retained.

Consider:

```go
type Entry struct {
	TS   Timestamp
	SRC  net.IP
	Tag  EntryTag
	Data []byte
	EVB  EVBlock `json:",omitempty"`
}
```

The struct tag is ineffectual in v1 as structs were *always* included. In v2, however, EVB's fields being zero would actually cause it to be omitted. While this was probably the intention, I opted to retain current behavior so the tag has been removed.

---

Bool fields are another case of changing-to-retain-behavior. Under v1, omitempty would omit a false bool. Under v2, omitempty will include a bool in all cases. Therefore, we actually want to use omitzero to omit a false bool.

Ex:
```go
var st = struct {
	B bool `json:",omitempty"`
}{}
one, _ := v1.MarshalIndent(st, "", "  ")
println("false bool under v1: ", string(one))
// Indent output for readability.
twoB, _ := v2.Marshal(st)
two := jsontext.Value(twoB)
two.Indent()
println("false bool under v2: ", string(two))
// in v2, you actually want omitzero
var stZero = struct {
	B bool `json:",omitzero"`
}{}
twoB, _ = v2.Marshal(stZero)
two = jsontext.Value(twoB)
two.Indent()
println("false bool under v2 omitzero: ", string(two))
```

>[!IMPORTANT]
>v2 considers the output JSON, not the input struct. Hence the nuance in omitempty v1 versus omitempty & omitzero v2.
>Further, the guidance for v2 is to explicitly prefer omitzero and satisfy the IsZero interface in many cases. I am leveraging this trick in the upcoming PATCH upgrade.

## Testing

This is a mammoth change across mono and backend. It comes with some functional changes as we embrace the new defaults. As such, I am relying heavily on pre-existing tests and getting them passing again.

## PR Tasks

## Reviewer Tasks


rory-landau-gravwell:
I found a substantial implementation difference in here to the point where I no longer think a bot is capable of safely working this task. As such, I'm throwing out this PR to ensure some of the bad changes don't sneak through.

PR #2738, opened 2026-09-04
fix(ingesters/kinesis): fail fast on cancellation and honor shutdowns everywhere

tyler-elton-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/2739

### Cause of the Bug

The Kinesis ingester spins up one goroutine _per Kinesis shard_ that all share a single `context.Context`. When the ingester shuts down via `SIGINT|SIGTERM|SIGQUIT` we call `cancel()` on that context. Any shard goroutine that is in the middle of a retry when this cancellation happens will get a `context.Canceled` error back from the `GetRecords` API call. After we log this error, we sleep via `utils.QuittableSleep` but **never checked if that sleep returned because of a context cancellation** and never checked `ctx.Err()` before retrying. This causes the goroutine to immediately retry the `GetRecords` API call again, causing the same error, log, repeat. `wg.Done()` never fired, so `main` never exited. 

There are some other high priority fixes in here revolving around race conditions as well (see below).

## This PR proposes...

- failing fast on context cancellation instead of spinning forever
- fixing a race condition around `metricsTracker` by guarding with a mutex
- fixing a race condition around `running` that's not synced by removing it in favor of `ctx.Err() == nil`
- bounding `wg.Wait()` with `utils.SyncExitTimeout`
- fixing a possible divide by zero error on stats counting

## PR Tasks

  - The Kinesis ingester is not readily testable via e2e or unit tests. We should follow up on this and make it testable (not in this pr to avoid scope creep) 

## Reviewer Tasks

### DIFF
--- a/ingesters/KinesisIngester/main.go
+++ b/ingesters/KinesisIngester/main.go
@@ -273,13 +273,17 @@ func main() {
 					output, err := svc.GetShardIterator(ctx, gsii)
 					if err != nil {
 						_ = lg.Error("error on shard", log.KV("number", shardid), log.KV("stream", stream.Stream_Name), log.KV("shard", *shard.ShardId), log.KVErr(err))
-						utils.QuitableSleep(ctx, 5*time.Second)
+						if utils.QuitableSleep(ctx, 5*time.Second) {
+							break reconnectLoop
+						}
 						continue
 					}
 					if output.ShardIterator == nil {
 						// this is weird, we are going to bail out
 						_ = lg.Error("got nil initial shard iterator, sleeping and retrying")
-						utils.QuitableSleep(ctx, 5*time.Second)
+						if utils.QuitableSleep(ctx, 5*time.Second) {
+							break reconnectLoop
+						}
 						continue
 					}
 					iter := *output.ShardIterator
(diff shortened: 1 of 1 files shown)


tyler-elton-gravwell:
Steps to test the fix:

1. Bring up Kinesis locally

```bash
cd ingesters/examples/aws
docker compose up -d
```

environment:
```yaml
  - GRAVWELL_INGEST_AUTH=IngestSecrets
  - GRAVWELL_INGEST_SECRET=IngestSecrets
```

Set up AWS CLI profiles per [ingesters/examples/aws/readme.md](https://github.com/gravwell/gravwell/blob/next-patch/ingesters/examples/aws/readme.md), then:

```bash
aws --profile kinesis-local kinesis create-stream --stream-name test --shard-count 1 --region us-east-1
```

2. Build and run the ingester

```bash
go build -o /tmp/kinesis-ingester ./ingesters/KinesisIngester
/tmp/kinesis-ingester -config-file ingesters/examples/aws/kinesis.conf -v &
echo "PID=$!"
```

Wait for `/tmp/kinesis.log` to show Ingester gone hot.

3. Force the error condition

```bash
docker stop kinesis-local
```

Within 5-10s, `/tmp/kinesis.log` should show "answer error" with context canceled/connection refused. That matches the crash report's error class.

4. Send shutdown signal, check for hang

```bash
kill -TERM <PID>
for i in $(seq 1 20); do kill -0 <PID> 2>/dev/null || { echo "exited"; break; }; sleep 0.5; done
kill -0 <PID> 2>/dev/null && echo "STILL RUNNING, bug present"
```

Fixed: exits within 1-2s, log shows a few more "answer error" lines then exiting.

Bug present: never exits, log floods thousands of lines/sec, needs `kill -9`.

5. Cleanup

```bash
docker start kinesis-local
cd ingesters/examples/aws && docker compose down
```

ingesters/KinesisIngester/main.go
```diff
@@ -370,7 +415,6 @@ func main() {
 	utils.WaitForQuit()
 	ib.AnnounceShutdown()
```
kris-watts-gravwell, line 420: this might be super racy now, we don't signal through that boolean anymore, we signal through the context which is on the other side of the sleep, this feels weird

ingesters/KinesisIngester/main.go
kris-watts-gravwell, line 425: now that the context is what shuts everything down is this an issue?  we sleep THEN cancel.

Why?

kris-watts-gravwell: this one is still ignored, is that on purpose?

ingesters/KinesisIngester/main.go
```diff
@@ -245,10 +261,19 @@ func main() {
...
 				// make the shardMetrics and add it to the array
 				tracker := shardMetrics{}
+
+				metricsTrackersMu.Lock()
 				metricsTrackers = append(metricsTrackers, &tracker)
+				metricsTrackersMu.Unlock()
+
 				if stream.Metrics_Interval == 0 {
 					// disable it
 					tracker.Disabled = true
```
kris-watts-gravwell, line 278: does this need to be guarded and happen before the locking above?

kris-watts-gravwell (changes requested):
check on comments and advise.

ingesters/KinesisIngester/main.go
tyler-elton-gravwell: nope, missed this one. addressed in a6576498eeb08ba0be6f53947c8f1dbfb29acfb1

ingesters/KinesisIngester/main.go
tyler-elton-gravwell: yes, addressed in 0e34344734a0868ae3f6f1409e943b983a594758

ingesters/KinesisIngester/main.go
tyler-elton-gravwell: the sleep mirrors our s3 and sqs ingesters by giving them time to process any in-flight `ProcessContext` calls, not about interrupting the aws polling calls. i added the comment around this that s3/sqs already have in 1851651b7ce2557cb0365c45c035915a3841dadd

ingesters/KinesisIngester/main.go
tyler-elton-gravwell: `running` would have only ever short-circuted in a narrow window between blocking aws calls. i believe my changes in here makes the best-case shutdown happen around 1s slower in favor of deleting an unsynchronized shared variable.

david-fritz-gravwell:
@claude review

david-fritz-gravwell:
@claude review

kris-watts-gravwell (approved):
LGTM

PR #2753, opened 2026-09-08
chore(#1325): upgrade gopsutil from v2 to v4

daniel-moreno-levy-gravwell:
This PR addresses https://github.com/gravwell/issues/issues/1325

## This PR proposes...

- Upgrade `github.com/shirou/gopsutil` from `v2.20.9+incompatible` (2020) to `v4.26.6` in the three importers: `ingest/log/utils.go`, `ingesters/base/base.go`, `ingesters/massFile/main.go`. Every call site (`host.PlatformInformation`, `host.KernelVersion`, `mem.VirtualMemory`) has the same signature in v4, so this is import paths only.
- Update `go.mod`/`go.sum` in the root module and in `tools/mock/mimecast/`, which `replace`s the parent. `e2e/` is unchanged: it depends on the published module rather than a local `replace`, so it picks this up at the next module release.
- Why: v2 uses deprecated APIs that warn on every macOS build, it is five years and two majors old, and the monorepo already had both majors in its graph. v2 did not break the monorepo on macOS (it imports `host` and `mem`, not the `disk` package whose darwin build fails), so this is hygiene rather than unblocking.
- `gopsutil v2.20.9 // indirect` remains in the root `go.mod` because `gravwell/v4/qa` imports `gravwell/v3/ingest/log`, which pins it. Out of scope here.
- Companion backend PR: gravwell/backend#4488. The two are independent: `client/types` already defines its own `AvgStat`, so neither PR needs the other to build. Targets `next-major` per team decision.

## PR Tasks

- [ ] e2e and/or unit tests included. If not, please provide an explanation. (No new tests: this is an import-path change with identical call-site signatures, and the existing tests exercise these packages.)
- [ ] **Bug fixes only:** minimal repro steps included on the issue (for PR QA + Release QA). (Not a bug fix.)
- [x] `go build` / `go vet` of the three packages clean for both `linux/amd64` and `darwin/arm64` with `CGO_ENABLED=0`.
- [x] `go mod tidy` is idempotent in all three modules.
- [x] No `gopsutil` v2 import path remains (`git grep`).

## Reviewer Tasks


kris-watts-gravwell (approved):
LGTM

PR #2756, opened 2026-09-09
fix(ingest): deadline and chunk writes to avoid stalling

tyler-elton-gravwell:
This PR **PARTIALLY** addresses https://github.com/gravwell/issues/issues/2820

## This PR proposes...

- **Root cause**: `bufio`'s internal auto-flush bypassed the only write deadline, so a stalled indexer could block writes forever
- Fix chunks writes with a per-block deadline, so slow transfers survive but truly stalled ones don't
- New config: `EntryReaderWriterConfig.FlushTimeout`/`WriteBlockSize`, `EntryWriter.OverrideFlushTimeout()`
- Hardened against zero-value misconfiguration (would otherwise spin or fail instantly)
- Tests (paused indexer, confirms no data loss)

## PR Tasks

## Reviewer Tasks

### DIFF
--- a/ingest/entryWriter.go
+++ b/ingest/entryWriter.go
@@ -91,7 +91,6 @@ type EntryWriter struct {
 	flshr         flusher
 	bIO           *bufio.Writer
 	bAckReader    *bufio.Reader
-	errCount      uint32
 	mtx           *sync.Mutex
 	ecb           entryConfBuffer
 	hot           bool
@@ -193,10 +192,14 @@ func (ew *EntryWriter) OverrideFlushTimeout(ft time.Duration) error {
 	}
 	ew.flushTimeout = ft
 
+	// ew.flushTimeout alone only governs flush()'s own deadline.
+	// The conn's Write() loop resets the deadline from its own copy
+	// before every block, so that copy has to be updated too or the
+	// override is a no-op.
+	ew.conn.SetFlushTimeout(ft)
+
 	return nil
 }
-
-type connWrapper func(conn) conn
 
 // wrapConn passes in a function that can wrap a reader/writer
 // when called we reset the write buffer, caller should make sure there isn't anything buffered
@@ -254,14 +257,6 @@ func (ew *EntryWriter) forceAckCtx(ctx context.Context) error {
 	ew.mtx.Lock()
 	defer ew.mtx.Unlock()
 	return ew.forceAckNoLock(ctx)
-}
-
-// outstandingEntries gives you a list of entries that have not been confirmed yet
-// the list IS NOT CLEARED, if you call it over and over you will get them all over and over
-func (ew *EntryWriter) outstandingEntries() []*entry.Entry {
-	ew.mtx.Lock()
-	defer ew.mtx.Unlock()
-	return ew.ecb.outstandingEntries()
 }
 
 // ejectOutstandingEntries is almost identical to outstandingEntries, but it also resets the confirmation buffer
@@ -571,16 +566,22 @@ func (ew *EntryWriter) flush() (err error) {
 		return
 	}
 
+	// Clear it unconditionally on the way out, not just on success.
+	// It sticks with the connection otherwise, even past the point
+	// this flush is done or has failed.
+	defer ew.conn.ClearWriteTimeout()
+
 	//issue the flush with timeout
-	if err = ew.bIO.Flush(); err == nil {
-		//check if we can cast to a flusher and flush
-		if ew.flshr != nil {
-			if err = ew.flshr.Flush(); err != nil {
-				return
-			}
-		}
-		err = ew.conn.ClearWriteTimeout()
+	if err = ew.bIO.Flush(); err != nil {
+		return
 	}
+	//check if we can cast to a flusher and flush
+	if ew.flshr != nil {
+		if err = ew.flshr.Flush(); err != nil {
+			return
+		}
+	}
+
 	return
 }
 
--- a/ingest/entryWriter.go
+++ b/ingest/entryWriter.go
@@ -1,5 +1,5 @@
 /*************************************************************************
- * Copyright 2017 Gravwell, Inc. All rights reserved.
+ * Copyright 2026 Gravwell, Inc. All rights reserved.
  * Contact: <legal@gravwell.io>
  *
  * This software may be modified and distributed under the terms of the
@@ -46,7 +46,7 @@ const (
 
 	maxThrottleDur time.Duration = 5 * time.Second
 
-	flushTimeout        time.Duration = 10 * time.Second
+	defaultFlushTimeout time.Duration = 10 * time.Second
 	negotiateTagTimeout time.Duration = 10 * time.Second
 )
 
@@ -74,6 +74,10 @@ const (
 	CONFIRM_DITTO_BLOCK_MAGIC    IngestCommand = 0x55667788
 )
 
+var (
+	ErrInvalidDuration = errors.New("invalid duration")
+)
+
 type IngestCommand uint32
 type entrySendID uint64
 
@@ -94,6 +98,7 @@ type EntryWriter struct {
 	buff          []byte
 	id            entrySendID
 	ackTimeout    time.Duration
+	flushTimeout  time.Duration
 	serverVersion uint16
 	ctx           context.Context
 }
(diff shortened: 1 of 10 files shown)


ingest/throttle.go
```diff
@@ -123,6 +162,50 @@ func (w *throttleConn) Write(b []byte) (n int, err error) {
...
+		blockSize = defaultWriteBlockSize
+	}
+
+	writeTimeout := fs.writeTimeout
+	if writeTimeout <= 0 {
+		writeTimeout = defaultFlushTimeout
+	}
+
+	for n < len(b) {
+		if err = fs.SetWriteDeadline(time.Now().Add(writeTimeout)); err != nil {
```
kris-watts-gravwell, line 193: you need to make sure to remove the write deadline on exit

ingest/entryWriter.go
```diff
@@ -535,7 +567,7 @@ func (ew *EntryWriter) WriteDittoBlock(ents []entry.Entry) error {
 // if the timeout expires we attempt to service acks and go back to attempting to flush
 func (ew *EntryWriter) flush() (err error) {
 	//set the write timeout
-	if err = ew.conn.SetWriteTimeout(flushTimeout); err != nil {
+	if err = ew.conn.SetWriteTimeout(ew.flushTimeout); err != nil {
```
kris-watts-gravwell, line 570: make sure to clear this deadline on the way out of this function, it sticks with the connection even after you are done with the flush

kris-watts-gravwell (changes requested):
few issues:

- [x] ingest/entryWriter.go:187 — OverrideFlushTimeout is a no-op. fullSpeed.Write re-sets the deadline from the construction-time fs.writeTimeout before every block, overwriting what flush() just installed from ew.flushTimeout. Reproduced: FlushTimeout: 5s + OverrideFlushTimeout(200ms) → write still failed at i/o timeout after 5.004506053s. The new test only asserts on the struct field.
- [x] ingest/throttle.go:152 — throttleConn.Write doesn't add r to n on the error paths, unlike the fullSpeed.Write sibling fixed in the same diff. Reproduced with a counting wrapper conn: reported n=0 while 65482 bytes actually reached the peer. With rate limiting on, syncAndCloseConnection → syncTimeout → flush() then re-sends those bytes on the same connection. Line 155 (lm.WaitN) has the same defect.
- [x] e2e/ingesters/HttpIngester/stall_test.go:76 — no defer e2e.UnpauseInstance(t); the t.Fatalf at line 120 skips the unpause at line 131 and leaves the shared container frozen.

Plus the deferred clears on the deadlines, a deadline does not go away after its hit or a write is done.

ingest/entryWriter.go
tyler-elton-gravwell: addressed in ca8672c3ec6ce5018a88df541de963fcebf82a06

ingest/throttle.go
tyler-elton-gravwell: addressed in ca8672c3ec6ce5018a88df541de963fcebf82a06

ingest/throttle.go
```diff
@@ -25,6 +42,12 @@ type throttleConn struct {
 	to    time.Duration
 	ctx   context.Context
 	cncl  func()
+	// writeTimeout/blockSize configure the per-block deadline behavior in Write
+	// below. Set to sane defaults by newThrottleConn. Tests construct a throttleCon
```
kris-watts-gravwell:
```suggestion
	// below. Set to sane defaults by newThrottleConn. Tests construct a throttleConn
```

ingest/throttle.go
```diff
@@ -105,15 +126,48 @@ func (w *throttleConn) Write(b []byte) (n int, err error) {
 		ctx, cancel = context.WithTimeout(w.ctx, w.to)
 		defer cancel()
 	}
+
+	// A caller that constructs a throttleConn literal directly (tests do
+	// exactly this, see throttle_test.go) can leave these values zeroed.
+	// Without this fallback, blockSize == 0 makes every chunk a zero-length
+	// write that "succeeds" without advancing n. An infite spin, not a clean
```
kris-watts-gravwell:
```suggestion
	// write that "succeeds" without advancing n. An infinite spin, not a clean
```

kris-watts-gravwell (changes requested):
few more minor nits


I am also working through some other stuff with deadlines on shutdown

kris-watts-gravwell:
- [x] While the deferred clear deadline calls on flush are correct it looks like we can now have an infinite deadline on Close which means shutdown can stall forever.
- [x] The deadlines on flush and what not are probably too tight, we had a 60s timeout before for really slow disks.  we might want to keep that?
- [x] newThrottleConn doesn't keep the configured values.  It silently rewrites flush timeouts and what not.   
- [x] throttleConn.Write will spin forever if burst is <= 0

ingest/throttle.go
tyler-elton-gravwell: addressed in f7a7e91abc5cd18b8cd0c1424f61e50303c7320f

ingest/throttle.go
tyler-elton-gravwell: addressed in f7a7e91abc5cd18b8cd0c1424f61e50303c7320f

tyler-elton-gravwell:
> * [ ]  While the deferred clear deadline calls on flush are correct it looks like we can now have an infinite deadline on Close which means shutdown can stall forever.

addressed in f7a7e91abc5cd18b8cd0c1424f61e50303c7320f

> * [ ]  The deadlines on flush and what not are probably too tight, we had a 60s timeout before for really slow disks.  we might want to keep that?

i've adjusted both the timeout to 30s (middle ground) and also lowered the chunk size so a stalled peer can get caught sooner without needing to have high/low timeouts in 84d3afeb4e4aeaef3ed571b7c397d7c53f7e7d91

> * [ ]  newThrottleConn doesn't keep the configured values.  It silently rewrites flush timeouts and what not.

addressed in f7a7e91abc5cd18b8cd0c1424f61e50303c7320f

> * [ ]  throttleConn.Write will spin forever if burst is <= 0

addressed in f7a7e91abc5cd18b8cd0c1424f61e50303c7320f

ingest/entryWriter.go
```diff
@@ -203,6 +248,17 @@ func (ew *EntryWriter) closeTimeout(to time.Duration) (err error) {
...
+	// forceAckNoLock above flushes at least once (throwAckSync), and
+	// flush()/the conn's own Write() always clear the write deadline on
+	// their way out, success or not (see flush() below, and
+	// fullSpeed.Write/throttleConn.Write in throttle.go). That erases the
+	// closeTimeout deadline set at the top of this function. flshr.Close()
+	// (ex: a compressor writing a trailer) and conn.Close() (ex: TLS writing
+	// a close_notify) can still hit the wire, so reinstate the deadline right
+	// before them or a peer that has stopped reading can stall shutdown
+	// forever instead of respecting closeTimeout.
+	ew.conn.SetDeadline(time.Now().Add(to))
```
kris-watts-gravwell, line 261: forceAckNoLock calls a flush which overrides the deadline set above.

The context above is passed into forceAckNoLock, so maybe we don't care, but the above deadline is still a no-op.

The above ops can also take time, so resetting the deadline to the original timeout means that the context will expire before the timeout/deadline does.

kris-watts-gravwell (changes requested):
if we have rate limiting total transfers are still bounded by the context
```
throttleConn.Write derives a context from w.to (set by flush()'s SetWriteTimeout) and passes it to lm.WaitN for every block, so the rate-limited path still enforces a whole-transfer deadline -- the exact behavior the per-block deadline was added to remove.
```

the burst test case still fails:

```
The `burst = blockSize` fallback claims to handle "a misconfigured rate limit with bps <= 0", but in that case w.lm was built with burst 0 too, so lm.WaitN(ctx, blockSize) rejects the request outright after blockSize bytes have already gone out on the wire.
```


there are also two issues with the e2e tests

kris-watts-gravwell:
Here is a patch that should help fix and express some of the issues called out:
[pr2756-review-fixes.patch](https://github.com/user-attachments/files/32215043/pr2756-review-fixes.patch)


Use as you see fit

tyler-elton-gravwell:
Here's what's been addressed in b75fd70fa86d7d9d7166cfcf4fb7e0625948459b:

- rate-limited path still enforces a whole transfer deadline
- `burst = blockSize` fallback half fix
- e2e test issues
- `forceAckNoLock`'s flush overrides the close deadline
- nested conn `SetFlushTimeout` doesn't propagate
- `hardDeadline` machinery

A few things from your patch/suggestions that I didn't add:

- `hardDeadline` field and accompanying functions
  - `clampFlushTimeout` gets the same result in less lines
- `flush()` change
  - not needed with the clamp approach
- `bufferWithoutFlushing` + 512KB payloads in tests
  - 512KB doesn't stall a loopback peer, so both tests would pass emptily
- wall-clock version of `TestOverrideFlushTimeoutReachesNestedConn`
  - w/ 16MB payload and 1KB blocks the write makes real progress before stalling, so `elapsed > 2s` is time-dependent noise. we should be asserting the enforced `writeTimeout` fields tests
- Missing `defer e2e.UnpauseInstance`
  - Already handled with `pause` flag + guarded `defer`
- `closeTimeout` absolute vs fresh deadline
  - I recommend absolute as it honors the stated budget. While fresh guarantees the close leg always gets a window for a compressor trailer / TLS `close_notify`, its at the cost of a 2x`to` at the worst case.


I also added `-race` to the ci checks for the `./ingest` package, as I found a race condition unrelated to my changes in there while testing locally.

kris-watts-gravwell (changes requested):
looking

kris-watts-gravwell (approved):
LGTM

Issue #2531, opened 2026-07-09
client: clean up comments & unify function nomenclature


rory-landau-gravwell:
# Task

Before opening a task, please complete the following checklist:
- [x] Check to see if this is a duplicate
  - I am 100% certain there is a duplicate of this issue floating around somewhere but cannot seem to find it. One day it was in my queue and the next day it was gone. If someone finds it, we can close this as a dup.

## What needs to be done?

The client library needs a bit of a refactor, especially now that the overhaul for registry has more or less settled.

There are three, primary tasks:

- [ ] Use consistent function names. This has been mostly taken care of as registry was installed, but we should go through and make sure there are no lingering GetAll versus ListAll calls, that ID paths have ID in the name for clarity, etc. 
- [ ] Update function comments. Make sure every exported function and important var/const is documented and that documentation is up to date with how it function sin registry.
- [ ] Remove dead code.

## Why is this task needed? 

6.0 is our chance to make changes to the client library and throw off some of the cruft that has built up over time.

## Any other comments?

This could take as much or as little effort as would like to be assigned. There are rougher edges that could/should be sanded off, but the client library is perfectly functional as is.

rory-landau-gravwell:
A substantial amount of this was completed in https://github.com/gravwell/gravwell/pull/2757. I am going to leave it open for a bit longer to give the client library one more pass to clean up verbiage.

PR #2757, opened 2026-09-09
feat(#2735): Encode collections as []/{} via JSONv2

rory-landau-gravwell:
This PR closes https://github.com/gravwell/issues/issues/2765 (per https://github.com/gravwell/issues/issues/2765#issuecomment-5546132099) and https://github.com/gravwell/issues/issues/2735 (while other instances may pop up, this should address the vast majority).

Sister PR: https://github.com/gravwell/backend/pull/4495

## This PR adopts JSON/v2 in many encoders and types

Importing JSON/v2 makes a substantial number of our marshalers superfluous.

For example: We have a lot of marshalers like:

```go
func (m *IdxStats) MarshalJSON() ([]byte, error) {
	type alias IdxStats
	return json.Marshal(&struct {
		alias
		IndexStats is
	}{
		alias:      alias(*m),
		IndexStats: is(m.IndexStats),
	})
}

type is []IndexManagerStats

func (i is) MarshalJSON() ([]byte, error) {
	if len(i) == 0 {
		return emptyList, nil
	}
	return json.Marshal([]IndexManagerStats(i))
}
```

All this does is ensure an empty array marshals to `[]` instead of `null`. And we have so, *so* many nearly identical cases. This is all because the v1 solution for sending frontend `[]`/`{}` instead of null is painful, arduous, and honestly silly.

JSON/v2 makes these entirely redundant as this behavior is now the default. I killed as many of these now-superfluous MarshalJSON as I could find.

**Please note! I am NOT expecting this to catch every slice/map null we send frontend. I am expecting it to catch *most* and set the stage for trivial follow up PRs to catch errant nulls.**

### DeadCustomMarshalers

In an effort to ensure types still marshal to the same (effective) result, I installed `client/types/marshallers_test.go:TestDeadCustomMarshalers()`, which matches behavior against what we are currently sending.

It does not test what we *expected* to send; I specifically built around what was *actually* being sent to ensure upgrading to v2 doesn't change the output. We had quite a number of structs that were marked with ineffectual tags (ex:  `SearchStatsResponse.Set` has an omitempty tag, but the alias in `SearchStatsResponse.MarshalJSON()` dropped this tag so the field was never being dropped).

>[!NOTE]
>DeadCustomMarshalers guarantees that the actual values will match. If we were using custom marshalers to send `[]` or `{}` before, we are now doing so natively. However, it does *not* guarantee byte-for-byte equity, only that the pre and post values are *functionally* equivalent.
>ex: `{"Stats":[],"TS":"0001-01-01T00:00:00Z"}` vs `{"TS":"0001-01-01T00:00:00Z","Stats":[]}`

## This PR implements JSON/v2 static actions

Rather than force everything to adopt JSON/v2, this PR implements a parallel set of static actions (`client/http_calls.go`) that *do* leverage v2. These new methods are functionally equivalent; they just have a nicer surface and rely on JSON/v2.

As we convert more calls to v2, we can slowly phase out older operators.

## This PR fits the client library asset functions to their settled shape

Addresses https://github.com/gravwell/gravwell/issues/2531.

Now that registry's assets have more or less settled, we can fit the client library to its actual shape. Rather than taking broad shapes where only a subset is respected (ex: taking QueryOptions as a param but only checking IncludeDeleted), we now use local structs that contain only what each call *can* accept.

These new local structs will still allow us to follow semver as we can transparently add fields to them.

## PR Tasks

  - The only reason this PR is a positive delta is because it adds 400+ lines of test code

## Reviewer Tasks


client/types/api.go
```diff
@@ -32,11 +32,7 @@ const (
 
 // Helpers for the marshaling functions
 var (
-	emptyList   = []byte(`[]`)
-	emptyObj    = []byte(`{}`)
-	emptyRawObj = RawObject(`{}`)
-	emptyString = []byte(`""`)
-	jsonNull    = []byte(`null`)
+	emptyList = []byte(`[]`)
```
rory-landau-gravwell, line 35: This remains for the few types where upgrading to v2 would have been more disruptive than I wanted for this already-disruptive PR.

client/types/marshallers_test.go
```diff
@@ -13,15 +13,18 @@ package types_test
...
+	v2 "encoding/json/v2"
+	"math"
 	"net"
 	"reflect"
 	"strings"
 	"testing"
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/gravwell/gravwell/v3/ingest"
```
kris-watts-gravwell, line 27: wrong import

client/http_calls.go
```diff
@@ -33,17 +60,61 @@ func (c *Client) patch[PatchT types.PatchType, ResponseT any](url string, data P
...
+	}
+	return resp.Body, nil
+}
+
 // reqDriver powers outbound requests and serves as a funnel to keep them consistent.
 // If err == nil, the caller is responsible for draining the response.
-func (c *Client) reqDriver(method string, url string, body []byte) (*http.Response, error) {
+//
+// okCodes is the set of non-200 codes that will be swallowed (instead of erroring).
+func (c *Client) reqDriver(method string, url string, body []byte, okCodes []int, params ...urlParam) (*http.Response, error) {
```
kris-watts-gravwell, line 117: this removes the STATE_AUTHED check, so we do a full round trip to figure out we are not authenticated rather than failing fast.  That intended?

client/types/search.go
```diff
@@ -558,16 +557,24 @@ type RowSelection struct {
 // recursion doom loop.
 type aliasRowSelection RowSelection
 
-func (rs RowSelection) MarshalJSON() ([]byte, error) {
+// MarshalJSONTo causes a Marshal to fail if the RowSelection is invalid.
+//
+// NOTE(rlandau): implemented as MarshalJSONTo instead of MarshalJSON in order to propagate encoder
```
kris-watts-gravwell, line 562: this means that the JSON is now case-sensitive, intended?

client/types/render.go
```diff
@@ -413,18 +413,18 @@ type IngesterStatsResponse struct {
...
-	RangeEnd    entry.Timestamp
-	Current     entry.Timestamp
-	Set         []StatSet         `json:",omitempty"`
+	Addendum    jsontext.Value  `json:",omitempty"`
+	RangeStart  entry.Timestamp `json:",omitzero"`
+	RangeEnd    entry.Timestamp `json:",omitzero"`
+	Current     entry.Timestamp `json:",omitzero"`
+	Set         []StatSet
 	OverviewSet []OverviewStatSet `json:",omitempty"`
 	Size        int               `json:",omitempty"`
```
kris-watts-gravwell, line 429: The old encoder always dropped Addendum and OverviewSet, this can now emit them.  Intended?

client/types/host.go
```diff
@@ -20,15 +19,15 @@ var (
 
 // SysInfo as displayed in the System Overview in Gravwell.
 type SysInfo struct {
```
kris-watts-gravwell, line 21: SysInfo.Empty() deliberately ignores VirtSystem, VirtRole and Error, so the deleted MarshalJSON collapsed an error-carrying SysInfo to `{}`. With the marshaler gone and per-field omitzero tags, those fields now serialize.
An indexer that fails to read /proc/cpuinfo returns `SysInfo{Error: "failed to read cpu info"}`. Empty() is true (no CPUCount/TotalMemoryMB/CPUModel/CPUMhz/CPUCache/SystemVersion), so GetSystemDescriptions previously delivered `{}` for that host; it now delivers `{"Error":"failed to read cpu info"}`. Same for a guest with only VirtSystem/VirtRole populated. The new SysInfo cases in TestDeadCustomMarshalers cover only all-zero and all-populated, so this divergence from the PR's stated output-equivalence goal is untested — surfacing the error is probably the better behavior, but it is an unflagged wire change.

kris-watts-gravwell (commented):
several questions and items

client/types/search.go
rory-landau-gravwell: Sort of. MarshalJSONTo/From passes the encoder from the catalyzing call, allowing options to be included in this custom (un)marshaller. If the catalyzing call allows case insensitivity (which our jsoncompat.Opts does), this should also be insensitize. I can add a test to double check that.

client/http_calls.go
rory-landau-gravwell: Good catch. No it was not; I must have missed it while mirroring the logic. Will fix.

client/types/render.go
rory-landau-gravwell: Ah, point taken. The anonymous struct before was missing fields. I will probably reinstall the old encoder then as a MarshalTo

client/types/host.go
rory-landau-gravwell: Do you want me to revert this then? You are right; this is a wire change more than just `null` -> `[]`/`{}`, but if it is preferred before I can make it explicit in the test code.

rory-landau-gravwell: While I like the idea of carrying `Error` through, the api-spec currently does not include it:
```json
 "properties": {
          "CPUCache": {
            "type": "string"
          },
          "CPUCount": {
            "type": "integer"
          },
          "CPUMhz": {
            "type": "string"
          },
          "CPUModel": {
            "type": "string"
          },
          "SystemVersion": {
            "type": "string"
          },
          "TotalMemoryMB": {
            "format": "int64",
            "type": "integer"
          },
          "VirtRole": {
            "description": "\"host\" or \"guest\"",
            "type": "string"
          },
          "VirtSystem": {
            "description": "e.g. \"kvm\" or \"xen\"",
            "type": "string"
          }
        },
```

client/http_calls.go
rory-landau-gravwell: https://github.com/gravwell/gravwell/pull/2757/commits/4f6d77758d46ccaccbc7702b6666211b66691686

client/types/search.go
rory-landau-gravwell: https://github.com/gravwell/gravwell/pull/2757/commits/46227eb7d71db14c9bd3698ea9935bd1d7767fe8

client/types/marshallers_test.go
rory-landau-gravwell: https://github.com/gravwell/gravwell/pull/2757/commits/106fe63d4cd8c1b37cd51914dd5354c1c5b4d460

client/types/render.go
rory-landau-gravwell: https://github.com/gravwell/gravwell/pull/2757/commits/2f7d524a79613f7590b8a7f60eaa6231152bb650

client/types/host.go
kris-watts-gravwell: no, an error getting this is so catastrophic that just not shipping data is fine i guess

kris-watts-gravwell (approved):
LGTM

PR #2758, opened 2026-09-10
Plugin: Add separate tag per section feature

keith-smiley-gravwell:
Patching jamf plugin to make each section have it's own tag because all sections bundled together creates huge JSON objects that

[Addresses this issue](https://github.com/gravwell/gravwell/issues/2776)

## This PR proposes...

- Each section gets it's own tag
- Adding the config parameter "Tag-Prefix"

## PR Tasks

## Reviewer Tasks

### DIFF
--- a/hosted/plugins/jamf/config.go
+++ b/hosted/plugins/jamf/config.go
@@ -18,11 +18,13 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"slices"
 
 	"github.com/gravwell/gravwell/v3/hosted"
 )
 
 const (
+	defaultIngesterUUIDStr string = "7e468cc4-ab10-4b33-b066-90eb4905980b"
 	defaultTagPrefix         = `jamf`
 	defaultPageSize          = 100
 	defaultLookback          = 1   // hours
@@ -60,7 +62,27 @@ type Config struct {
 	Insecure_Skip_TLS_Verify bool
 }
 
+// Equal implements hosted.Config so the runner can decide whether a config reload
+// actually changed anything for this ingester.
+func (c *Config) Equal(ncp any) bool {
+	nc, ok := hosted.EqualTarget[Config](ncp)
+	if c == nil || !ok {
+		return false
+	}
+	return c.BaseConfig == nc.BaseConfig &&
+		c.Tag_Prefix == c.Tag_Prefix &&
+		c.PollingConfig == nc.PollingConfig &&
+		c.Host == nc.Host &&
+		c.Client_Id == nc.Client_Id &&
+		c.Client_Secret == nc.Client_Secret &&
+		c.Page_Size == nc.Page_Size &&
+		c.Insecure_Skip_TLS_Verify == nc.Insecure_Skip_TLS_Verify &&
+		slices.Equal(c.Sections, nc.Sections)
+}
+
 func (c *Config) Verify() error {
+	c.ApplyDefaultIngesterUUID(defaultIngesterUUIDStr)
+
 	if c.Host == "" {
 		return errors.New("Host not specified")
 	}
--- a/hosted/plugins/jamf/config.go
+++ b/hosted/plugins/jamf/config.go
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	defaultIngesterUUIDStr string = "7e468cc4-ab10-4b33-b066-90eb4905980b"
+	defaultIngesterUUIDStr   = "7e468cc4-ab10-4b33-b066-90eb4905980b"
 	defaultTagPrefix         = `jamf`
 	defaultPageSize          = 100
 	defaultLookback          = 1   // hours
@@ -70,7 +70,7 @@ func (c *Config) Equal(ncp any) bool {
 		return false
 	}
 	return c.BaseConfig == nc.BaseConfig &&
-		c.Tag_Prefix == c.Tag_Prefix &&
+		c.Tag_Prefix == nc.Tag_Prefix &&
 		c.PollingConfig == nc.PollingConfig &&
 		c.Host == nc.Host &&
 		c.Client_Id == nc.Client_Id &&
--- a/hosted/plugins/jamf/config.go
+++ b/hosted/plugins/jamf/config.go
@@ -62,6 +62,8 @@ type Config struct {
 	Insecure_Skip_TLS_Verify bool
 }
 
+var _ hosted.Config = (*Config)(nil) // compile time interface check
+
 // Equal implements hosted.Config so the runner can decide whether a config reload
 // actually changed anything for this ingester.
 func (c *Config) Equal(ncp any) bool {
@@ -71,6 +73,7 @@ func (c *Config) Equal(ncp any) bool {
 	}
 	return c.BaseConfig == nc.BaseConfig &&
 		c.Tag_Prefix == nc.Tag_Prefix &&
+		c.MultiTagConfig == nc.MultiTagConfig &&
 		c.PollingConfig == nc.PollingConfig &&
 		c.Host == nc.Host &&
 		c.Client_Id == nc.Client_Id &&
--- a/hosted/plugins/jamf/config.go
+++ b/hosted/plugins/jamf/config.go
@@ -8,43 +8,47 @@
 
 // Package jamf is a hosted ingester plugin that polls the Jamf Pro API's
 // computers-inventory endpoint for records that were modified in a given
-// time window, using general.reportDate as the tracking cursor.
+// time window, using general.reportDate as the tracking cursor. Each
+// configured section is emitted to its own tag, with the GENERAL section's
+// data folded into every entry for context.
 package jamf
 
 import (
+	"cmp"
 	"errors"
 	"fmt"
-	"slices"
 	"strings"
 
 	"github.com/gravwell/gravwell/v3/hosted"
 )
 
 const (
-	defaultIngesterUUIDStr string = "7e468cc4-ab10-4b33-b066-90eb4905980b"
-
-	defaultTag               = `jamf`
+	defaultTagPrefix         = `jamf`
 	defaultPageSize          = 100
 	defaultLookback          = 1   // hours
 	defaultRequestsPerMinute = 60  // token bucket for the inventory + oauth calls
 	defaultInterval          = 600 // seconds (10 minutes)
 
-	// PollBufferSeconds keeps the window from reaching all the way up to
+	// pollBufferSeconds keeps the window from reaching all the way up to
 	// "now", since Jamf can take a few seconds to finish writing a record
 	// after a device checks in. Without this, records near the trailing
 	// edge of a window could be missed entirely.
 	pollBufferSeconds = 60
 )
 
-const sectionGeneral string = "GENERAL"
+// generalSectionName is the Jamf computers-inventory section that's always
+// requested (for reportDate/id) and folded into every other section's
+// entry. It only needs a place in Sections if you also want a dedicated
+// general-only stream.
+const generalSectionName = "GENERAL"
 
-// Sections lists the computers-inventory sections we'll request if the
-// config doesn't specify any explicitly.
-var defaultSections = []string{sectionGeneral}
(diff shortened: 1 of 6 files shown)


hosted/plugins/jamf/config.go
```diff
@@ -8,43 +8,47 @@
...
 	"errors"
 	"fmt"
-	"slices"
 	"strings"
 
 	"github.com/gravwell/gravwell/v3/hosted"
 )
 
 const (
-	defaultIngesterUUIDStr string = "7e468cc4-ab10-4b33-b066-90eb4905980b"
```
nic-osullivan-gravwell, line 24: Should we be dropping this default?

hosted/plugins/jamf/config.go
```diff
@@ -56,29 +60,7 @@ type Config struct {
 	Insecure_Skip_TLS_Verify bool
 }
 
-var _ hosted.Config = (*Config)(nil) // compile time interface check
-
-// Equal implements hosted.Config so the runner can decide whether a config reload
-// actually changed anything for this ingester.
-func (c *Config) Equal(ncp any) bool {
```
nic-osullivan-gravwell, line 63: We don't want to delete these. Removing this prevent dynamic reload from working.

hosted/plugins/jamf/example.conf
```diff
@@ -17,18 +17,32 @@
 	Path="/opt/gravwell/etc/hosted_runner.state"
 	Sync=false
 
-[Jamf "yourserver"]
+[Jamf "utad"]
 	Ingester-UUID="99100000-0000-0000-0000-000000000000" #example UUID, remember to set
-	Host=https://yourserver.jamfcloud.com
+	Host=https://jamf00.utad.utoledo.edu:8443
```
nic-osullivan-gravwell: I assume we don't want this here...

hosted/plugins/jamf/config.go
```diff
@@ -56,29 +60,7 @@ type Config struct {
...
-		c.Host == nc.Host &&
-		c.Client_Id == nc.Client_Id &&
-		c.Client_Secret == nc.Client_Secret &&
-		c.Page_Size == nc.Page_Size &&
-		c.Insecure_Skip_TLS_Verify == nc.Insecure_Skip_TLS_Verify &&
-		slices.Equal(c.Sections, nc.Sections)
-}
-
 func (c *Config) Verify() error {
-	c.ApplyDefaultIngesterUUID(defaultIngesterUUIDStr)
```
nic-osullivan-gravwell, line 80: What's the reason for deleting this?

hosted/plugins/jamf/config_test.go
```diff
@@ -75,90 +111,101 @@ func TestConfig_Verify_Defaults(t *testing.T) {
...
+			want:     []string{"jamf_applications", "jamf_storage"},
 		},
 		{
-			name:         "empty sections get GENERAL defaults",
-			sections:     []string{},
-			wantSections: defaultSections,
+			name:     "custom prefix",
+			sections: []string{"APPLICATIONS"},
+			prefix:   "utad",
+			want:     []string{"utad_applications"},
```
nic-osullivan-gravwell:
```suggestion
			prefix:   "custom",
			want:     []string{"custom_applications"},
```

hosted/plugins/jamf/example.conf
```diff
@@ -17,18 +17,32 @@
...
 	#Page-Size=100
 	#Lookback=1              # hours; how far back to look on first run
 	#Requests-Per-Minute=60
 	#Request-Interval=600    # seconds between poll cycles
-	#Sections=GENERAL
-	#Sections=DISK_ENCRYPTION
-	#Sections=STORAGE
+
+	# Tag-Prefix overrides the "jamf" prefix used to build per-section tags
+	# (e.g. Tag-Prefix=utad -> utad_disk_encryption, utad_storage, ...).
```
nic-osullivan-gravwell:
```suggestion
	# (e.g. Tag-Prefix=custom -> custom_disk_encryption, custom_storage, ...).
```

hosted/plugins/jamf/example.conf
```diff
@@ -17,18 +17,32 @@
 	Path="/opt/gravwell/etc/hosted_runner.state"
 	Sync=false
 
-[Jamf "yourserver"]
+[Jamf "utad"]
```
nic-osullivan-gravwell:
```suggestion
[Jamf "yourserver"]
```

hosted/plugins/jamf/example.conf
nic-osullivan-gravwell:
```suggestion
	Host=https://yourserver.jamfcloud.com
```

nic-osullivan-gravwell (changes requested):
Looks like some unrelated changes leaked in, but the core looks ok.

hosted/plugins/jamf/config.go
keith-smiley-gravwell: Fixed in [1464adb](https://github.com/gravwell/gravwell/pull/2758/commits/1464adbaf834fac06a3f5a99c54db8de1f40314a)

hosted/plugins/jamf/config.go
keith-smiley-gravwell: Fixed in [1464adb](https://github.com/gravwell/gravwell/pull/2758/commits/1464adbaf834fac06a3f5a99c54db8de1f40314a)

hosted/plugins/jamf/config.go
keith-smiley-gravwell: Fixed in [1464adb](https://github.com/gravwell/gravwell/pull/2758/commits/1464adbaf834fac06a3f5a99c54db8de1f40314a)

hosted/plugins/jamf/config.go
```diff
@@ -56,8 +62,6 @@ type Config struct {
 	Insecure_Skip_TLS_Verify bool
 }
 
-var _ hosted.Config = (*Config)(nil) // compile time interface check
```
nic-osullivan-gravwell, line 59: We need to keep this for the check.

hosted/plugins/jamf/config.go
```diff
@@ -66,7 +70,7 @@ func (c *Config) Equal(ncp any) bool {
 		return false
 	}
 	return c.BaseConfig == nc.BaseConfig &&
-		c.SingleTagConfig == nc.SingleTagConfig &&
+		c.Tag_Prefix == nc.Tag_Prefix &&
```
nic-osullivan-gravwell, line 75: I think this should be:
```suggestion
		c.MultiTagConfig == nc.MultiTagConfig &&
```
as this should include changes to TagName as well.

nic-osullivan-gravwell (changes requested):
Few small issues.

hosted/plugins/jamf/config.go
keith-smiley-gravwell: Fixed in [aa9db03](https://github.com/gravwell/gravwell/pull/2758/commits/aa9db039556f9631a9409dd462e71d9284749940)

hosted/plugins/jamf/config.go
keith-smiley-gravwell: Fixed in [aa9db03](https://github.com/gravwell/gravwell/pull/2758/commits/aa9db039556f9631a9409dd462e71d9284749940)

keith-smiley-gravwell:
Commit [6f8c198](https://github.com/gravwell/gravwell/pull/2758/commits/6f8c1982628f98dfadaede84a037436c76b471a9) includes the following:

Review complete — 10 findings reported, ranked most-severe first. Highlights:

- **Two real correctness bugs in `splitRecord`** ([jamf.go:183](hosted/plugins/jamf/jamf.go:183), [jamf.go:192](hosted/plugins/jamf/jamf.go:192)): top-level record fields outside `general`/the matching section are now silently dropped (evidenced by the PR's own test fixtures being rewritten around the loss), and a device with no data for any configured section produces zero entries — it vanishes from ingestion for that cycle with no log line.
- **A default-behavior break**: the new default `Sections` splits what used to be a single `jamf` tag into `jamf_disk_encryption`/`jamf_storage`, with no migration note for existing deployments.
- **Lost test coverage**: `TestConfigEqual` was deleted with no replacement, so `Config.Equal` (which drives dynamic-reload restart decisions) is now untested — notable since this PR is exactly the kind of change that test existed to catch.
- A **latent-bug trap**: `tagForSection` reimplements `hosted.MultiTagConfig.ResolveTag` with a different separator; a future "simplification" to use the inherited method would silently rename every tag.
- Smaller items: a reverted context-handling convention (currently harmless but fragile), repeated fragile `QuitableSleep`-check boilerplate in the Kinesis fix, a per-record recomputation that could be cached, a gofmt violation, and a silently-dead defensive branch.

hosted/plugins/jamf/config.go
```diff
@@ -97,18 +100,73 @@ func (c *Config) Verify() error {
...
+
+// tagForSection returns the tag that a configured section's entries should
+// be written under: Tag-Name verbatim if set (only valid for a single
+// configured section), otherwise "<Tag-Prefix or 'jamf'>_<section, lowercased>".
+func (c *Config) tagForSection(section string) string {
+	if c.Tag_Name != "" {
+		return c.Tag_Name
+	}
+	prefix := cmp.Or(c.Tag_Prefix, defaultTagPrefix)
+	return prefix + "_" + strings.ToLower(section)
```
nic-osullivan-gravwell: This was a good callout from Claude we generally do `-` separators is there a reason we need to use `_`? 

If we already have data in `_` tags then it's probably best to keep those users on `Tag-Name` and have `-` be the standard going forward, rather than a single plugin using a different standard. 

I assume we have a way to reingest data to a different tag, then we could let the old tag ageout.

nic-osullivan-gravwell (changes requested):
@keith-smiley-gravwell were those findings fully addressed or are you still working through those? Most of those seem important to address.

keith-smiley-gravwell:
> @keith-smiley-gravwell were those findings fully addressed or are you still working through those? Most of those seem important to address.

Sorry I guess I didn't go through the code-review for jamf yet...I'll get that in asap

hosted/plugins/jamf/config.go
keith-smiley-gravwell: "_" is just my preference, but "-" is usually standard
I'll have to add a bit to remove the "_" from the section name e.g. DISK_ENCRYPTION so we don't went up with a funky tag name e.g. jamf-disk_encryption

keith-smiley-gravwell:
Commit xxx includes the following:

All 10 findings applied and verified (build, vet, gofmt, and the full `hosted/...` test suite all pass). Summary of changes:

**jamf.go**
- `splitRecord` now preserves top-level record fields (e.g. `id`/`udid`) that aren't tied to any requested section, instead of dropping them.
- A device with no data for any configured section now emits a single GENERAL-only fallback entry instead of vanishing for that poll cycle (Handle negotiates a GENERAL tag up front for this).
- `initClient` uses the `ctx` parameter again instead of `rt.Context()`, restoring the established convention shared with mimecast/wiz.
- Section→JSON-key conversion is precomputed once per `drain()` call (via new `sectionKeyIndex`) instead of recomputed per record.
- The "tag not found" branch in `splitRecord` now panics with a clear message instead of silently dropping data.
- Added two tests: `TestHandle_PreservesTopLevelRecordFields` and `TestHandle_FallsBackToGeneralWhenNoConfiguredSectionMatches`.

**config.go**
- Fixed import ordering (gofmt-clean now).
- Added a doc comment on `defaultSections` warning about the tag-naming breaking change for upgraders relying on defaults.
- Added a doc comment on `tagForSection` explaining why it doesn't call `hosted.MultiTagConfig.ResolveTag` (different separator/casing — calling it would silently rename existing tags).

**config_test.go**
- Restored `TestConfigEqual` using `configtest.CheckEqual`, adapted to the new `MultiTagConfig`/`Sections` shape.

Also changed line 160 in ```config.go``` to:

``` return prefix + "-" + strings.Replace(strings.ToLower(section),"_","",-1)```

PR #2759, opened 2026-09-10
Preprocessor: normalize heavily escaped json objects

keith-smiley-gravwell:
[Addresses this issue](https://github.com/gravwell/gravwell/issues/2777)

## This PR proposes...

- A new preprocessor to normalize heavily escaped json objects

## PR Tasks

## Reviewer Tasks


kris-watts-gravwell (changes requested):
### Blocking — silent data corruption on the happy path

- **`ingest/processors/jsonnormalize.go:190`** — New `entry.Entry` copies only
  Tag/SRC/TS/Data, dropping the `EVB` enumerated-value block. Any EVs attached
  by an upstream processor (e.g. `attach`) are lost. Convention is
  `CopyEnumeratedBlock` (`jsonsplit.go:170`), or set `ent.Data` in place like
  `gzip.go`.

- **`ingest/processors/jsonnormalize.go:173`** — `json.Unmarshal` into
  `interface{}` coerces all numbers to `float64`, losing precision above 2^53:
  `{"id":1234567890123456789}` → `{"id":1234567890123456800}`. Breaks 19-digit
  IDs (Snowflake, EventRecordID) and epoch-nanosecond timestamps. Use
  `json.Decoder` with `UseNumber()`.

- **`ingest/processors/jsonnormalize.go:183`** — `json.Marshal` HTML-escapes by
  default: `?a=1&b=2` → `?a=1\u0026b=2`, `<b>` → `\u003cb\u003e`. Breaks
  downstream regex/grok extraction. Use `json.Encoder` with
  `SetEscapeHTML(false)` (trim the trailing newline `Encode` adds).

### Should fix

- **`ingest/processors/jsonnormalize.go:283`** — `normalizeValue` decrements
  `depth` per *structural* level, not per escaping layer, so the budget is spent
  walking plain objects. With the default `Max_Depth` of 8, a document nested 8
  objects deep comes back byte-identical — the escaped payload is never inlined.
  CloudTrail, k8s audit and Windows event JSON routinely exceed this, so the
  processor no-ops on exactly the sources it targets, with no error or log line.

- **`ingest/processors/jsonnormalize.go:183`** — Invalid UTF-8 in string values
  is replaced with U+FFFD (`0xff` → `ef bf bd`, confirmed at byte level).
  `json.Valid` accepts such input, so this hits the success path and no
  `Passthrough_Non_JSON` setting avoids it. Affects latin-1/cp1252 payloads.

### Cleanup

- **`ingest/processors/jsonnormalize.go:215`** — The `unescapeOnce` "strict path"
  is unreachable: it is only called under `!json.Valid(data)`, but
  `json.Unmarshal` into a string succeeds only when the input *is* valid JSON.
  All repair falls through to the byte heuristic, and the documented
  `\n`/`\t`/`\uXXXX` handling never runs. Drop the block, or restructure so
  quoted literals are unquoted strictly before the heuristic.

keith-smiley-gravwell:
Commit [34c425f](https://github.com/gravwell/gravwell/pull/2759/commits/34c425f7bdc1cea007ef7ac60475e20281dae61a) includes the following:

Both fixes applied to jsonnormalize.go:

1. **Line ~340** — removed `'"'` from the re-parse trigger switch, so a string is only treated as further-escaped JSON when it looks like an object/array (`{`/`[`). Literal quoted text (e.g. `"quoted phrase"`) now passes through untouched. Updated the two adjacent doc comments that referenced the removed behavior.
2. **Line ~175, 209** — `depth` is now declared outside Step 1's repair loop and subtracted from the budget passed into Step 2 (`jn.maxDepth-depth`), so `Max_Depth` bounds escaping layers across the whole record instead of granting each step its own full budget.

Verified both against a standalone harness reproducing the two bug scenarios (now fixed) plus the pre-existing test scenarios from jsonnormalize.go (whole-entry quoting, nested field, and the Max_Depth=1 cases) — all still pass unchanged.

PR #2760, opened 2026-09-10
Plugin: Converting thinkst plugin from experimental to current

keith-smiley-gravwell:
[Addresses this issue](https://github.com/gravwell/gravwell/issues/2775)

## This PR proposes...

- Updating the experimental thinkst plugin to the current hosted framework

## PR Tasks

## Reviewer Tasks


hosted/plugins/thinkst/thinkst_test.go
```diff
@@ -0,0 +1,375 @@
...
+
+// fakeRuntime is a minimal, self-contained implementation of hosted.Runtime
+// for exercising Handle() without a real ingest muxer or BoltDB. The real
+// hosted package has its own equivalent (testRuntime), but it is unexported
+// and lives in package hosted, so plugin packages need their own; this
+// mirrors the pattern the real hosted/plugins/jamf and hosted/plugins/msgraph
+// tests use, embedding hosted.StatusTracker for the SetError/ClearError/
+// SetWarn/ClearWarn methods hosted.Runtime requires rather than hand-rolling
+// them.
+type fakeRuntime struct {
```
nic-osullivan-gravwell: I believe we have a mock runtime in the hosted package now for testing so we don't need to reimplement.

hosted/plugins/thinkst/config.go
```diff
@@ -0,0 +1,69 @@
...
+// IsValidApi reports whether a is a supported Api value.
+func IsValidApi(a Api) bool {
+	switch a {
+	case IncidentApi, AuditApi:
+		return true
+	}
+	return false
+}
+
+type Config struct {
```
nic-osullivan-gravwell, line 41: This might make more sense to match closer to Mimecast with a `hosted.MultiTagConfig` and allow specifying api multiple times. 

Are there only 2 thinkst apis, or only 2 we support?

nic-osullivan-gravwell (changes requested):
tests seem to have good coverage, but some duplicated bits. 

Only main things is the multiple APIs like mimecast over distinct stanzas.

hosted/plugins/thinkst/config.go
keith-smiley-gravwell: Seems to be several hundred endpoints across 31 categories, but audit logs and incidents are the 2 main ones

hosted/plugins/thinkst/thinkst_test.go
keith-smiley-gravwell: Fixed in [a860f4c](https://github.com/gravwell/gravwell/pull/2760/commits/a860f4ca8eb9ba0a46f73316a78018847de323f2)

keith-smiley-gravwell: Address in [a860f4c](https://github.com/gravwell/gravwell/pull/2760/commits/a860f4ca8eb9ba0a46f73316a78018847de323f2)

hosted/plugins/thinkst/config.go
nic-osullivan-gravwell: Ah then yeah I think we want this to be MultiTag to avoid needing a config Stanza per, or force everything into one tag in the future. Easier to do now.

hosted/plugins/thinkst/config.go
```diff
@@ -0,0 +1,69 @@
...
+	hosted.BaseConfig
+	hosted.SingleTagConfig
+	hosted.PollingConfig
+	Domain string
+	Token  string `json:"-"` // DO NOT send this when marshalling
+	Api    Api
```
nic-osullivan-gravwell, line 47: Oh and this probably needs to be a slice 

```suggestion
	Api    []Api
```

With the down stream changes to Handle to support multiple.

nic-osullivan-gravwell (changes requested):
Bumping back to update to MultiTag like Mimecast as there are a lot of potential apis we might support in the future, don't want to get boxed in.

keith-smiley-gravwell:
Commit [efe360b](https://github.com/gravwell/gravwell/pull/2760/commits/efe360bdd07979ed1f5efc1e5c8284f23414308b) includes the following:

All nine findings scoped to `hosted/plugins/thinkst/` are fixed; the gofmt finding in `hosted/plugins/builder.go`/`config.go` was skipped per your instruction to leave other files alone.

## What changed

**config.go**
- Added `Config.Equal` (+ `var _ hosted.Config`) so config reloads no longer force a restart every reconcile cycle.
- `Verify()` now rejects an empty `Api` list, calls `MultiTagConfig.ValidateTags()`, calls `BaseConfig.Verify()` and `ApplyDefaultIngesterUUID`, and only requires `Tag-Name` when a single API is configured — `Tag-Prefix` now actually works.
- Added `Api.Tag(tag, prefix)` (mirrors `mimecast`'s pattern) and updated `Tags()` to resolve one tag per configured API.

**thinkst.go**
- `Handle()` now iterates over *every* configured `Api` (not just the first), negotiating a per-API tag.
- Split storage keys per API (`incident-since-id`/`incident-cursor` vs `audit-cursor`/`audit-timestamp`/`audit-timestamp-hashes`) — testing the multi-API fix surfaced a real collision (both handlers shared one `cursor` key, so running incidents+audit together would clobber each other's pagination state).
- `handleIncidents` now skips (rather than ingests) a record whose metadata fails to parse, matching `handleAudit` and preventing that record from being re-ingested on every poll forever.
- Added a `stringify` helper so a cursor gated as "present" but JSON-`null` persists as `""`, not the literal `"<nil>"`.
- `handleAudit`'s dedup now tracks content hashes of records at the current watermark second, so two distinct audit events sharing the same second no longer silently lose one of them.

**client.go**
- `doJSON` decodes with `UseNumber()` so a numeric pagination cursor round-trips exactly instead of risking float precision loss.

**Tests**: updated existing tests for the new key names/semantics and added coverage for each fix (multi-API dispatch, malformed-record skip, null-cursor guard, same-second audit dedup, `Equal`/`Tags`/`Api.Tag`, UUID defaulting). `go build`, `go vet`, and `go test -race ./hosted/plugins/thinkst/...` all pass; `gofmt -l` is clean on the touched files.

One tradeoff worth flagging: the audit dedup fix is state-format-breaking for any already-running Thinkst audit instance — on the first poll after upgrade, records at the previously-stored watermark second get one-time re-ingested (since their hashes aren't yet known), after which it behaves correctly going forward.

hosted/plugins/thinkst/config.go
keith-smiley-gravwell: Added in [96d9ec3](https://github.com/gravwell/gravwell/pull/2760/commits/96d9ec36814549f438adefb34fb118d4d8390faa)

hosted/plugins/thinkst/config.go
keith-smiley-gravwell: Fixed in [96d9ec3](https://github.com/gravwell/gravwell/pull/2760/commits/96d9ec36814549f438adefb34fb118d4d8390faa)

PR #2761, opened 2026-09-11
feat(2820): update Azure Event Hubs ingester to use new SDK

tyler-elton-gravwell:
This PR **PARTIALLY** addresses https://github.com/gravwell/issues/issues/2820 alongside https://github.com/gravwell/gravwell/pull/2756

## This PR proposes...

- updating the Azure Event Hubs ingester to use the new, actually unarchived and maintained package that Microsoft recommends
  - as-is this would also break existing customers, but there's logic in here to migrate existing checkpoints to the new format
- backoff retry logic to avoid hammering
- Go modernization for this ingester
- tests
- add offline support + example

image

## PR Tasks

## Reviewer Tasks


ingesters/AzureEventHubs/config.go
```diff
@@ -14,29 +14,32 @@ import (
...
 	config.IngestConfig
+	// State_Store_Location is a local directory where per-partition
+	// checkpoint files are written. It's created on startup if it
+	// doesn't already exist.
 	State_Store_Location string
 }
 
 type eventHubConf struct {
 	Event_Hubs_Namespace  string
+	Event_Hubs_Endpoint   string // optional; when set, connects to a local Event Hubs emulator at this host instead of Event_Hubs_Namespace on Azure
```
nic-osullivan-gravwell, line 42: 👏  yay for ease of local testing

nic-osullivan-gravwell (changes requested):
We should add an e2e test for this now we're on an SDK version that supports it. Should be easy enough to use the compose file we have directly in the test (maybe sans the gravwell instance) 

Doesn't need to be exhaustive, but at least a "yep we get data on the happy path" with maybe some follow ups for more testing later.

e2e/ingesters/AzureEventHubs/azureeventhubs_test.go
```diff
@@ -0,0 +1,159 @@
...
+		for _, e := range ents {
+			if strings.Contains(e.String(), msg) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			e2e.Fatalf(t, "did not find expected event %q in results: %+v", msg, ents)
+		}
+	}
```
nic-osullivan-gravwell, line 158: This doesn't assert event ordering is that the expected behavior of the event hubs or should things be in the right order?

nic-osullivan-gravwell (changes requested):
overall looks good one Q before approval though.

e2e/ingesters/AzureEventHubs/azureeventhubs_test.go
tyler-elton-gravwell: this does not test the underlying logic of how event hubs orders things. its more testing the happy path of "did both events get ingested with the right content"

nic-osullivan-gravwell (approved):
lgtm

PR #2770, opened 2026-09-15
task(issues#2531): standardize client library

rory-landau-gravwell:
This PR closes https://github.com/gravwell/gravwell/issues/2531.

Sister backend PR: https://github.com/gravwell/backend/pull/4519

## This PR normalizes the methods and diction used by the client library

- Kills the depreciated AddUser/AddGroup methods
- Renames quite a number of methods s.t. they follow common patterns established by the rest of the client library:
  - AddExtraction --> CreateExtraction
  - TestAddExtraction --> ValidateExtraction
  - AddSelfTargetedNotification --> CreateNotification
  - DeleteKitEx --> DeleteKitVerbose
  - MailConfig --> GetMailConfig
  - ConfigureMail --> SetMailConfig
  - WellData --> GetWellData
  - SearchQueue --> GetSearchQueue
  - PutGuiPreferences --> UpdateGuiPreferences
  - DeleteGuiPreferences --> ClearGuiPreferences
  - GetFullSecret --> GetSecretFull
  - CapabilityList --> ListCapabilities
  - CapabilityTemplateList --> ListCapabilityTemplates
  - TOTPClear --> ClearTOTP
  - ClearAllMFA --> ClearMFA
  - AdminClearUserMFA --> AdminClearMFA
- Adds missing functions so all assets share the same set (List, ListAll, Get, GetEx, Delete, Purge, Create, Update, Cleanup)
- Swapped param order in DownloadRequestWithContext(path, ctx) to comply with Go idiom

## This PR fixes ValidateExtraction

`client.ValidateExtraction()` was straight up broken and backend's `testAddAX` was out of compliance with the api-spec.

The backend sent either an Error (as `{"Error":"x"}`) or `struct {TagExists bool; Error string}`. Only the latter is expected in the api-spec.

`client.ValidateExtraction()` could only unpack to ClientError. Any other response was tossed. However, it purported to return []WarnResp, but this was obviously never populated.

This PR (and its backend sister) fix this by making `struct {TagExists bool; Error string}` a concrete type and forcing backend to always return it.

>[!IMPORTANT]
>I have some qualms about the shape of extractor operations in the spec, but will kick out a follow up issue if deemed necessary. I don't want it holding up this PR.

## This PR moves extractor client operation into their own file

Rather than `admin.go`.

## This PR implements some v2 drivers I missed

Extractors were still sending with the old static actions. This PR fixes that, making them send using the reqDriver-derived JSON/v2 helpers (ex: `c.post()`).

## PR Tasks

  - The only logical changes are in the Extraction operations that got cleaned up mildly.

## Reviewer Tasks


client/tokens.go
```diff
@@ -35,6 +35,17 @@ func (c *Client) ListTokens(opts *types.QueryOptions) (ts types.TokenListRespons
...
 }
 
+// ListAllTokens (admin-only) gets a list of all tokens on the system. If
+// non-nil, the QueryOptions will be applied for pagination,
+// filtering, etc.
+func (c *Client) ListAllTokens(opts *types.QueryOptions) (ts types.TokenListResponse, err error) {
+	if opts == nil {
+		opts = &types.QueryOptions{}
+	}
+	opts.AdminMode = true
```
kris-watts-gravwell: you need to clear this after the call, opts is a pointer which means you mutate the main opts struct in the client and this will carry forward.


It can't be a blind clear either, store what was there, set it, do the call, then restore what was there

kris-watts-gravwell (dismissed):
Admin parameter changes opts through pointer and is never restored in tokens call

client/tokens.go
rory-landau-gravwell: That's.... a fair point, though it is the pattern we use across the board for all types; tokens just happen to have been missing that call before. I'll fix it across the board

rory-landau-gravwell: https://github.com/gravwell/gravwell/pull/2770/commits/1771a5466401f3e10599552e15bab6ec4d00e34d

rory-landau-gravwell:
Pulling this and its sister https://github.com/gravwell/backend/pull/4519 back while I update signatures to pass QO by value.

rory-landau-gravwell:
Found some inconsistencies in how AXs are handled by the client library. Fixing those before sending this back out

client/alerts.go
```diff
@@ -18,20 +18,14 @@ func (c *Client) CreateAlert(def types.Alert) (result types.Alert, err error) {
 }
 
 // ListAlerts returns a list of alerts the user has access to.
-func (c *Client) ListAlerts(opts *types.QueryOptions) (result types.AlertListResponse, err error) {
-	if opts == nil {
-		opts = &types.QueryOptions{}
-	}
-	return c.post[types.QueryOptions, types.AlertListResponse](ALERTS_LIST_URL, opts)
+func (c *Client) ListAlerts(opts types.QueryOptions) (result types.AlertListResponse, err error) {
+	return c.post[types.QueryOptions, types.AlertListResponse](ALERTS_LIST_URL, &opts)
```
kris-watts-gravwell, line 22: don't pass by reference on these query option functions

kris-watts-gravwell (commented):
some package level common query options that you can just re-use might be useful:

```
var (
   NoOpts = types.QueryOptions{}
   AdminOpts = types.QueryOptions{Admin: true}
)
```

client/alerts.go
rory-landau-gravwell: JSON/v2 suggests passing all data to be marshaled by reference as it will address it anyways so doing so in advance saves a shallow copy. That is why post/get/delete/etc take *T:

encoding/json/v2/arshal.go:

image

kris-watts-gravwell (approved):
LGTM

client/alerts.go
kris-watts-gravwell: LGTM

Issue #2603, opened 2026-08-03
netflow v9 generator


ashnwade:
# Feature

Before opening a feature request, please complete the following checklist:
- [ ] Check to see if this is a duplicate

## What is the feature to be added?
netflow v9 generator

## Why should we add this feature? (Business justification? What problem is the feature trying to solve?)
We have netflow v5 and v10, but we should also add v9 

## Any other comments?
Related to https://github.com/gravwell/gravwell/issues/387#issuecomment-5123313547

PR #2778, opened 2026-09-16
Implement issue #2603 and add a pure netflowv9 generator

kris-watts-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/2603

## This PR proposes...

- add a pure NetflowV9 generator to gravwellGenerator

image

## PR Tasks

## Reviewer Tasks


tyler-elton-gravwell (commented):
More clanker notes:

## Important

**1. `netflowv9gen.go:38-39, 93-96` — the uint32 millisecond uptime wrap the comment claims to prevent is still reachable, in roughly half of long-duration runs.**

```go
// keep uptime well below the uint32 millisecond wrap (~49.7 days)
nfv9MaxBootBehind = 40 * 24 * time.Hour
...
uptime := uint32(ts.Sub(nfv9Boot).Milliseconds())
```

The bound is applied only to the *initial* boot offset, never to `offset + elapsed generated span`. `nfv9MaxBootBehind` = 3,456,000,000 ms against a uint32 ceiling of 4,294,967,295 ms leaves just 9.71 days of headroom for the timestamp span.

Concrete scenario: `gravwellGenerator -type netflowv9 -duration 4w` (the `w`/`d` suffixes are supported by `getDuration` in base.go:304-330). `newSequentialTSGenerator` starts at `now-28d` and walks forward, so the first call picks `nfv9Boot = ts0 - U(0, 40d)`. Whenever that draw exceeds 21.7 days — about 46% of runs — `ts.Sub(nfv9Boot)` crosses 49.7 days partway through the run and the `uint32` conversion truncates. `msg.Header.SysUptime` jumps backwards to a small value mid-stream (the exporter appears to spontaneously reboot with no sequence reset), and for the records right after the wrap `uptime < 1000` drives both underflow guards at lines 194-200, pinning `FIRST_SWITCHED` and `LAST_SWITCHED` to 0. Downstream flow-time reconstruction (which derives absolute flow times from `ExportTime - (SysUptime - LAST_SWITCHED)`) then produces flow times ~49 days in the past. The generated corpus silently contains garbage for the tail of the run.

Fix direction: bound the *total* uptime rather than just the boot offset — compute `d := ts.Sub(nfv9Boot)` and, when `d.Milliseconds()` approaches `math.MaxUint32`, deliberately re-base `nfv9Boot` forward to `ts` (an explicit simulated reboot) instead of letting the conversion truncate. Better still, derive the boot offset from `cfg.Duration` so the whole span is guaranteed to fit. Note this is copied verbatim from `netflowv5gen.go:27, 61-64, 102-109`, which has the identical defect — worth fixing both in one pass rather than propagating it a third time.

Confidence: 85.

**2. `netflowv9gen.go:33-35` vs `ipfixgen.go:32-34` — identical source/domain ID *and* identical template IDs mean the v9 and IPFIX generators corrupt each other's templates inside the netflow ingester's session map.**

netflowv9 uses `nfv9SourceID = 1`, `nfv9V4TemplateID = 256`, `nfv9V6TemplateID = 257`. ipfixgen uses `ipfixDomainID = 1`, `ipfixV4TemplateID = 256`, `ipfixV6TemplateID = 257`. But the two template 256s are structurally different: 18 field specifiers / 45 bytes for v9, 14 field specifiers / 60 bytes for IPFIX.

`getSessionKey(domain, ip)` at `ingesters/netflow/handlers.go:195-205` keys sessions on `(domainID, sourceIP)` only — the version is deliberately *not* part of the key, and `ipfix.Session.specifiers` is a flat `map[uint16][]TemplateFieldSpecifier` with no version partitioning (`registerUnaliasedTemplateRecord`, parser.go:464-477). Concrete scenario: someone testing the netflow ingester runs both generators with `-raw-udp-connection <ingester>:2055` from the same host (a very natural thing to do given this PR exists to exercise the v9 path). Both streams hash to the same `sessionKey`, and whichever packet arrives last overwrites template 256/257. The next packet from the *other* generator then has its data FlowSet decoded against the wrong template: either `readDataRecord` over-reads (v9 45-byte records parsed as 60-byte IPFIX records → short read, packet rejected at handlers.go:283-287 and silently dropped), or under-reads and yields structurally valid but semantically garbage fields. It flaps packet-by-packet, so it presents as intermittent ~50% packet loss with no error that names the cause.

Fix direction: give netflowv9 a distinct source ID (e.g. `nfv9SourceID = 2`) and distinct template IDs (e.g. 258/259) so the two generators cannot alias in a shared session. Distinct template IDs alone are the stronger fix, since thesource ID is attacker/user-controllable via config anyway.

Confidence: 80.

## Missing tests

`netflowv9gen_test.go` is a genuine round-trip test, not a smoke test — it reconstructs a fresh `ipfix.Session`, parses, looks up templates, interprets every field, and checks address-family lengths, protocol values, and `first <= last <= SysUptime`. That is substantially better than most generator tests in this package. But it is a *self-consistency* test: it validates the bytes only through the same library that produced them, so any encoding error shared by`Marshal` and `MessageHeader.unmarshal` is invisible to it. Specific gaps:

1. **Raw v9 header layout.** Nothing asserts the 20 header bytes directly: version at `[0:2]` == 9, count at `[2:4]`, SysUptime at `[4:8]`, ExportTime at `[8:12]`, sequence at `[12:16]`, source ID at `[16:20]`, all big-endian. The existing assertion `int(msg.Header.Length) == len(TemplateRecords)+len(DataRecords)` (test line ~57) is a tautology — the library writes `len(tmpl)+len(data)` into offset 2 in `marshalHeader` and reads offset 2 back into `Header.Length`for v9. A byte-offset assertion against the RFC 3954 §5.1 layout would be a real check.
2. **Template identity, not just cardinality.** The loop compares *both* parsed templates against `len(nfv9V4Template.FieldSpecifiers)`, which passes only because v4 and v6 happen to both have 18 fields. It should assert the template set is exactly `{256, 257}` and that template 256's specifiers deep-equal `nfv9V4Template.FieldSpecifiers` (FieldID and Length, in order) and 257's equal `nfv9V6Template`'s. Today, swapping FieldID 21/22 or setting `INPUT_SNMP` to Length 4would slip past most of this test.
3. **Monotonic non-wrapping SysUptime — the finding above.** Advance `ts` across a multi-week span (e.g. 256 samples an hour apart, or a direct call with `ts = boot + 50*24h`) and assert `msg.Header.SysUptime` is non-decreasing andstrictly greater than zero across consecutive calls, and that `FIRST_SWITCHED`/`LAST_SWITCHED` stay within `nfv9MaxFlowMS` of `SysUptime` rather than collapsing to 0. The current test advances `ts` by only 256 seconds, which can neverreach the wrap.
4. **FlowSet structure.** No test walks the FlowSets independently of the library: expect exactly one FlowSet with ID 0 (the v9 template set — note IPFIX uses 2, v9 uses 0, and `marshalTemplates` branches on version to get this right) followed by at most two data FlowSets with IDs 256 and 257, each with `Length == 4 + n*recordLen`, and the sum of FlowSet lengths plus 20 equal to `len(b)`.
5. **UDP datagram bound.** The package doc comment asserts each entry is "the data portion of a single NetFlow v9 UDP packet," but nothing asserts `len(b) <= 1472`. Worth pinning, since it currently holds only by luck of `nfv9MaxRecords = 15` × the 81-byte v6 record; raising `nfv9MaxRecords` to 18 would silently start producing fragmenting datagrams.
6. **Both address families and both next-hop branches are covered only probabilistically.** `rand`'s global source is auto-seeded in Go 1.20+, so nothing forces a v6 record, an ICMP record, a zero next hop, or a zero ToS to appear in any given run, and a failure cannot be reproduced. Either seed a local `*rand.Rand` deterministically or assert that all of {v4 record, v6 record, proto 1, proto 6, proto 17, zero next hop} were observed across the 256 iterations so thetest fails loudly if a branch stops being reachable. (`seedVars(128)` seeds the *data pools*, not `math/rand`.)
7. **Collision regression for finding 2.** A test asserting `nfv9V4TemplateID != ipfixV4TemplateID` and `nfv9SourceID != ipfixDomainID` (or, more meaningfully, that feeding one v9 message and one IPFIX message into a *single shared* `ipfix.Session` still decodes both correctly) would lock in the fix.
8. **Wiring.** No test asserts `getGenerator("netflowv9")` returns a non-nil `DataGen` and a non-nil finalizer, or that `"netflowv9"` appears in `getList()`. Cheap, and it's the kind of thing that rots when someone adds a type to`dataTypes` but forgets `finalizers`.

kris-watts-gravwell:
Thanks — finding 1 is real and I've fixed it in both v5 and v9. Pushing back on finding 2 and on one piece of the test list.

## Finding 2 — no template corruption is possible here

The mechanism is described accurately: `getSessionKey` ([handlers.go:195](https://github.com/gravwell/gravwell/blob/main/ingesters/netflow/handlers.go#L195)) omits the version, `Session.specifiers` is a flat `map[uint16][]TemplateFieldSpecifier`, and the two template 256s really are structurally different (18 fields / 45 bytes for v9 vs 14 fields / 60 bytes for IPFIX). But the conclusion doesn't follow from it.

**Every message from both generators carries its own templates ahead of its own data FlowSets**, and `readBuffer` walks sets in wire order, calling `registerTemplateRecord` as it goes. So each packet re-registers the templates it needs immediately before its own data records are decoded. Whatever the previous packet left in `specifiers` is overwritten before it can be read. The aliasing window never opens.

This isn't incidental — it's the property the generator was built around (it's why the template set leads every message) and the same property the ingester relies on when it attaches templates and re-marshals. The existing test already pins it by parsing each entry with a **fresh** `ipfix.Session` that has no prior template knowledge.

I tested the exact scenario: 100 messages interleaved v9/IPFIX through a **single shared** `ipfix.Session`, every record checked for field count and dictionary field names.

```
interleaved v9/ipfix through ONE shared session with no corruption
```

No corruption, no rejected packets, no "intermittent ~50% packet loss." Confidence 80 was too high.

I did make one change here, on different grounds: `nfv9SourceID` is now `2` rather than colliding with `ipfixDomainID`. Two generators model two different exporters, so sharing an observation domain was sloppy modelling regardless, and distinct IDs keep the sessions separate in any collector keyed on `(domain, source)`.

**Template IDs stay at 256/257.** That's what real v9 exporters emit, and moving them would make the generated data less realistic to buy protection against a problem that doesn't exist. I'd rather keep the corpus faithful.

## Missing test 7 — premise gone, but I took the better half

The `nfv9V4TemplateID != ipfixV4TemplateID` assertion would lock in a constraint I'm deliberately not adopting. But your own stronger suggestion — "feeding one v9 message and one IPFIX message into a *single shared* session still decodes both correctly" — is a genuinely good test, because it documents *why* the overlap is safe rather than just forbidding it. Added as `TestNetflowV9IpfixShareSession`.

## Finding 1 — one of the two suggested fixes isn't reachable

"Derive the boot offset from `cfg.Duration` so the whole span is guaranteed to fit" isn't available: `base.DataGen` is `func(time.Time) []byte`, with no path to the config. Went with the reboot approach instead, which is more robust anyway since it also holds for `-stream` runs of unbounded length.

## Everything else

The rest of the test list was fair and is implemented. Two calls worth flagging:

- You were right that `int(msg.Header.Length) == len(TemplateRecords)+len(DataRecords)` was a tautology — the library writes and reads the same offset. Replaced with byte-offset assertions against the RFC 3954 §5.1 layout plus an independent FlowSet walk.
- `seedVars(128)` does only seed the data pools, not `math/rand` — confirmed. Rather than seeding a local `*rand.Rand`, `TestNetflowV9Branches` asserts all of {v4, v6, proto 1/6/17, TCP flags, both next-hop branches, both ToS branches} were observed across 512 messages, so a branch going unreachable fails loudly.

Aside, unrelated to this PR: if anyone does try `-raw-udp-connection`, `RawConn.writeBytes` ([base/raw.go](https://github.com/gravwell/gravwell/blob/main/generators/base/raw.go)) issues a second `Write` for the trailing `"\n"`, which on a UDP conn is a separate 1-byte datagram. Harmless here — the ingester drops it as too short — but it's a wart worth its own issue.

tyler-elton-gravwell (approved):

Issue #2457, opened 2026-06-06
ingesters: tag not established until data actually flows


kris-watts-gravwell:
# Bug Report

Before opening a bug report, please complete the following checklist:
- [ ] Check the documentation for intended behavior, notes, warnings, and edge cases
- [ ] Check open AND closed issues to see if this is a duplicate
	- A closed duplicate issue may exist if this is a wontfix

## What are the steps to reproduce this issue?

1. setup cluster
2. install ingester (simple relay)
3. configure a listener that will put data to `foo-bar-baz`
4. Go to interface

## What happens?

1. Systems and Health page for ingesters shows that the listener is there and shows that the tag is being negotiated
2. go to search page and try to search on the tag (it does not exist)

## What were you expecting to happen?

Traditional ingesters should establish and negotiate the tag immediately

## Any other comments?

This is a regression

## What versions of software are you using?

5.9.0

PR #2779, opened 2026-09-16
establish tags immediately rather than when data flows

kris-watts-gravwell:
This PR addresses https://github.com/gravwell/gravwell/issues/2457

### Problem
`IngestMuxer.NegotiateTag` stopped negotiating. Since a569e39f it only queued the tag on
each connection's `toNegotiate` list; the actual wire negotiation happened in
`connSet.translateTag`, reached only from the write relay routine — i.e. when the first
entry carrying that tag was written.

Config tags were unaffected (they ride in on the auth handshake). Tags registered at
runtime were not: preprocessor routers (`jsonrouter`, `csvrouter`, `forwarder`,
`corelight`, `gravdup`), HEC dynamic tags, and HTTP ingester hot reloads. Those tags
showed up in Systems & Health (reported from ingester state) but did not exist on the
indexer and could not be searched until data flowed.

## This PR proposes...

- **`ingest/muxer.go`**
  - `NegotiateTag` split into `stageTag` (everything under `im.mtx`, returns the
    connections needing work) plus a negotiation loop that runs **with the muxer lock
    released**. Restores pre-regression behavior without reintroducing the reason it was
    made async — the old code held `im.mtx` across a synchronous indexer round trip,
    starving the write relay routines and state push timers.
  - On failure: log and close the connection. The tag is already in `im.tags`, so the
    reconnect negotiates it during the auth handshake.
  - New `connSet.negotiateOutstandingTags`, shared by the relay path and `NegotiateTag`.
    `tagTrans.active` is a dense slice indexed by intermediate tag, so tags must register
    in order — a `negMtx` serializes the two callers, and pending tags are peeked and only
    dropped once they register.
  - `tagTrans` locking was incomplete: `translate`, `hasTag` and `reverse` read `active`
    lock-free, and `registerTag`'s sync check ran outside the lock. Now an `RWMutex`
    covering both fields.
- **`ingest/utils.go`** — `tagMaskTracker` was a plain `[8192]byte` written by
  `NegotiateTag` under `im.mtx` but read by every `WriteEntry` with no lock. Now
  `[2048]uint32` with atomic or/and/load. The race detector flags the old version as soon
  as a tag is negotiated on a running muxer.

### Tests
- **`ingest/muxer_test.go`** — three tests against the in-process test indexer: immediate
  establishment, batch/repeat negotiation keeping local and remote tag sets aligned, and
  negotiation concurrent with an active write load. `ingest/testindexer_test.go` gains
  `hasTag`/`tagNames`.
- **`e2e/ingesters/HttpIngester/reload_test.go`** (+ `testdata/reload.conf`) —
  `TestHotReloadNegotiatesTag`: start on a randomized tag, SIGHUP in a config with a
  different tag, then assert the new tag exists in `client.GetTags()` (the indexer's view)
  and is searchable with zero entries — **without ever posting an entry**. The existing
  unit tests only checked `muxer.KnownTags()`, the ingester's local map, which is why the
  regression got through.

### Verification
- `go test ./ingest/ -race -count=1` — passes; all three new unit tests fail on the
  pre-fix `muxer.go`.
- `cd e2e && go test ./ingesters/HttpIngester/ -count=1` — passes; the new e2e test fails
  on the pre-fix `muxer.go` with `tag "reload-after-…" was not negotiated with the indexer
  during the hot reload, no data has flowed on it`.

## PR Tasks

## Reviewer Tasks


kris-watts-gravwell:
@claude review this

tyler-elton-gravwell (commented):
Clanker found a few things while reviewing:

### Important (80-89)

**1. `ingest/muxer.go:936-956` + `2130`/`1946-1949` — a tag negotiated while a connection is mid-(re)connect is silently dropped for that indexer, which is exactly the "no data will ever flow on this tag" case the PR is fixing.**
`stageTag` only queues work for connections already installed in `im.igst`. `getConnection` snapshots `im.tags`/`im.tagMap` under `im.mtx.RLock()` (line 2056-2130, covering `initConnection`'s tag handshake and `newTagTrans`), then releases the lock and spends an unbounded amount of time in `IdentifyIngester` → the `IngestOK` loop (which sleeps 10s per iteration while an indexer is still coming up) → `ConfigureStream`, and only installs `igst`/`tt` at line 1946. A `NegotiateTag` landing in that window is invisible to that destination: not in `tt.active`, not in `tt.toNegotiate`, and nothing reconciles after install.
Concrete scenario: HttpIngester with two indexers. `base.GetMuxer` returns as soon as `WaitForHot` sees *one* hot connection (`WaitForHotContext` returns on `cnt > 0`), then `includeStdListeners`/`hotReload` call `NegotiateTag` for each listener tag. Indexer B is still in its `IngestOK` sleep loop; it gets installed with a stale tag set. `NegotiateTag` returns nil, and because nothing ever writes on that tag, indexer B never learns it — the original bug, just narrowed rather than fixed.
Fix direction: reconcile at install time. In `connRoutine`, while holding `im.mtx` at 1946, diff `im.tagMap` against the freshly built translator and `registerTagForNegotiation` anything added since the handshake snapshot, then call `negotiateOutstandingTags` after dropping the lock (same pattern as `NegotiateTag`). A tag-generation counter captured before `initConnection` and compared at install would make the check cheap.

**2. `ingest/muxer.go:880-895` — `NegotiateTag` returns `nil` after a wire negotiation fails, so the new contract ("the tag exists on the backend when this returns") is unenforceable by callers, and the failure is inconsistent with the errors `stageTag` does propagate.**
`lerr` is logged and the connection is closed; `err` (from `stageTag`) is untouched. Two concrete consequences:
- Single-indexer ingester whose indexer rejects the tag (`ERROR_TAG_MAGIC` from `entryReader.go:490-501`, e.g. tag-count limits) or is older than `MINIMUM_TAG_RENEGOTIATE_VERSION` (`entryWriter.go:927`): `NegotiateTag` returns success to the HEC handler / router preprocessor, which then treats the tag as live. The only signal is a log line.
- Because `negotiateOutstandingTags` drains FIFO and aborts on the first error, `NegotiateTag("A")` can return nil when tag `B` (staged earlier by another goroutine) failed and `A` was never even attempted — and the warning it logs names `A`, not `B`.
Note also the inconsistency: a `registerTagForNegotiation` failure *does* surface as an error (via `stageTag`), while an actual negotiation failure does not.
Fix direction: decide and document the contract. At minimum, aggregate per-connection failures and return an error when zero connections negotiated successfully; if best-effort is intended, say so in the doc comment and make the "no connection got it" case distinguishable. Log the tag actually being negotiated (`v.name`) rather than the caller's `name`. Secondary: for a permanent condition like `serverVersion < MINIMUM_TAG_RENEGOTIATE_VERSION`, check `ig.ew.serverVersion` before queueing instead of bouncing a healthy connection once per new tag — reconnect cannot fix the version, and the handshake would have established the tag anyway.

**3. `ingest/muxer.go:880-894`, `2438-2456` — negotiation is serial across connections in the caller's goroutine, each round trip leads with a full `forceAck`, and there is no deadline; this reintroduces the stall the async change was made to avoid (just without `im.mtx`).**
`ew.NegotiateTag` (`entryWriter.go:933`) starts with `forceAckNoLock(ew.ctx)`, which flushes and waits for *every* outstanding ack (up to `MAX_UNCONFIRMED_COUNT` = 4096 entries) while holding `ew.mtx` and `igst.mtx` exclusively — so the relay routine for that connection cannot write for the duration, and `negMtx` is held across the whole drain, so the relay also blocks in `translateTag` behind it. `NegotiateTag` walks `pending` one connection at a time, so latency is the sum over all destinations.
Concrete scenario: HEC listener with `?tag=` dynamic tags against a 10-indexer cluster under load. Every first-seen tag costs 10 sequential (flush + ack-drain + RTT) round trips inside the HTTP handler, each one pausing that indexer's write pipeline. A slow indexer makes each leg cost ~`ackTimeout` (the configured connection timeout) before erroring, and `forceAckNoLock`'s only cancellation is `ew.ctx`, i.e. muxer shutdown. The ack-read timeouts keep this bounded rather than infinite, so I'm not calling it a hang, but it is a multi-second request stall plus a pipeline stall per new tag.
Fix direction: negotiate the `pending` connections concurrently (one goroutine per `connSet`, `WaitGroup`, collect errors) — each `connSet` has its own `tt`/`ig`, so `negMtx` already makes that safe — and give the whole thing a deadline (a `context`/`negotiateTagTimeout` plumbed into `ig.NegotiateTag`) so one wedged indexer can't hold `negMtx` and stall a relay routine for an unbounded number of ack timeouts.

## Missing tests

The three new unit tests are real (they use the in-process indexer, and `new-tag`/`concurrentX` can only appear in `ti.tags` via an actual wire negotiation, so they do pin the regression), but they only cover the happy path on a single destination. What's absent:

1. **Entries actually landing on the correct remote tag.** All three tests assert only that the indexer *minted* a name and that `Sync` succeeds. `testIndexer` counts entries but never records their tag values, so a translator ordering skew — exactly what `negMtx` and `registerTag`'s `int(local) != len(tt.active)` check exist to prevent — would leave every test green while entries silently land in the wrong well. Needed: have `testIndexer` record a `map[entry.EntryTag]uint64` of received tags, then assert that an entry written on local tag `T` for name `N` arrives with the indexer-side id the indexer minted for `N`.
2. **Concurrent `NegotiateTag` from multiple goroutines for distinct names.** `TestNegotiateTagWhileWriting` negotiates serially from one goroutine; the only concurrency is with the relay. Needed: N goroutines each negotiating a distinct name, then assert every name exists on the indexer, `KnownTags()` matches, and (with #1) every tag round-trips to the right remote id. This is what actually exercises the two-caller `negMtx`/dense-slice invariant, under `-race`.
3. **The negotiation-failure and reconnect-recovery path** — the riskiest new code, currently untested. Needed: a `testIndexer` knob to reject a named tag (return `ERROR_TAG_MAGIC`) or to drop the connection mid-negotiation, then assert `NegotiateTag`'s behavior, that the connection was closed and reconnected (`ti.Accepted()` increments), and that the tag is established via the auth handshake afterwards. This test is also what pins down the resolution of finding 2.
4. **Multi-destination coverage.** No test builds a muxer with two `testIndexer`s. Needed: a two-indexer test asserting both indexers receive the tag, plus a partial-failure variant (one indexer accepts, one rejects) asserting whatever the agreed contract is for the return value.
5. **A regression test for finding 1.** Gate one `testIndexer` so it stalls in `IngestOK`/handshake, call `NegotiateTag` during that window, then release and assert the tag eventually exists on that indexer *without any entry being written*. This should fail against the current patch.
6. **A pure unit test for `tagTrans`** (no network): `registerTag` out-of-order rejection, `peekToNegotiate`/`clearToNegotiate` FIFO behavior, `negotiationsPending` transitions, `hasTag` boundaries, `translate` fallback for an unknown tag. Cheap, fast, and it documents the invariant the rest of the change rests on.
7. **`tagMaskTracker` concurrency.** `TestTagBitMask` already walks all 65536 tags for the new offset/mask math (good coverage of the bit arithmetic), but there is no test with concurrent `add`/`has` goroutines hitting the same 32-bit word under `-race` — which is the whole justification for the change.
8. **e2e `TestHotReloadNegotiatesTag` can pass for the wrong reason.** Two gaps: (a) it polls for 30s, so it cannot distinguish "negotiated immediately" from "established by a connection bounce" — and the PR's own failure path closes the connection, so a broken negotiation still lands the tag via the handshake. It should also assert the muxer didn't reconnect; the ingester log is already pulled, so counting `reconnecting`/`re-connected` lines before and after the reload is nearly free. (b) Confirm `/api/tags` (`client.GetTags`) is indexer-derived and not aggregated from ingester state reports — the PR description itself says the broken tags *did* show up in Systems & Health from ingester state. If `/api/tags` includes that, the test passes on `next-patch` too and proves nothing; please run it against the base branch and confirm it fails.

## Minor (bundled, not blocking)

- `ingest/muxer.go:2218` (pre-existing, adjacent): `if int(v) > len(tt.active)` should be `>=`. With a non-contiguous `tagMap` (the comment at line 396 explicitly says "tag IDs can be all over the place" for a recovered tagcache), `v == len(tt.active)` slips through and `tt.active[v] = tg` panics inside the `connRoutine` goroutine, crashing the ingester. This PR leans harder on the dense/contiguous invariant, so it's worth tightening while you're here.
- `ingest/muxer.go:2474`, `2488`: `translate` now takes an `RWMutex` RLock on the per-entry path (two atomics per entry per connection). Each `tt` is effectively single-reader, so contention should be nil, but if it ever shows in a profile, an `atomic.Pointer` to an immutable `[]entry.EntryTag` swapped on register would give lock-free reads.
- `ingest/muxer.go:886-891`: the warning logs the caller's `name`, but the failure may belong to a different, earlier-queued tag (see finding 2) — log `v.name` from the failing iteration instead.

kris-watts-gravwell:
@claude review this and check that we address all the comments

ingest/muxer.go
```diff
@@ -2441,43 +2745,87 @@ func (tt *tagTrans) registerTag(local entry.EntryTag, remote entry.EntryTag) err
...
+func (tt *tagTrans) negotiationsPending() (ok bool) {
+	tt.mtx.RLock()
+	ok = len(tt.toNegotiate) > 0
+	tt.mtx.RUnlock()
+	return
+}
+
+// peekToNegotiate hands back the next tag awaiting negotiation without removing
+// it, the entry is only dropped via clearToNegotiate once it has actually been
+// registered.  Callers must hold negMtx.
```
tyler-elton-gravwell, line 2778: Does `negMtx` still exist? I can't see it anywhere...
```suggestion
// registered.
```

tyler-elton-gravwell (changes requested):
Besides the comment update, my clanker also found a few other issues that we might want to address:

**1. A waiter registered against a `tagTrans` that gets discarded by a reconnect is never reported, so `NegotiateTag` burns the full 10s and returns `ErrTimeout` for a tag that actually got established**
`ingest/muxer.go:1018-1026` (staging), `:1680-1692`, `:1889-1898`, `:1964-2012` (the bounce paths).

Failure scenario, single-indexer ingester:
1. `NegotiateTag("foo")` → `stageTag` finds connection A live, `w.add()` (pending=1), queues onto `tt1`, `wake()`.
2. Before the relay routine services `<-nc.tt.notify`, something else bounces the connection — the `<-tmr.C` case's `nc.ig.syncTimeout` fails (muxer.go:1901), or a `WriteEntry` fails (:1794), or a state push fails (:1917), or connRoutine's `getConnection` dies outright and `close(ncc)` makes the relay exit.
3. connRoutine rebuilds: `getConnection` → `initConnection(tgt, im.tags, ...)` — and `im.tags` now contains `foo`, so the indexer mints it in the *authentication handshake*. `reconcileTags` finds nothing missing. The tag is established and searchable.
4. `tt1`, with the queued `unNegotiatedTag{waiter: w}`, is garbage. Nothing in `syncAndCloseConnection`, `recycleConnection`, the `<-csc` swap, or connRoutine's teardown calls `tt.failQueued` or otherwise reports. `w.pending` stays 1 forever.
5. `NegotiateTag` blocks the full `tagNegotiationTimeout` (10s) and returns `ErrTimeout`.

Impact is not cosmetic: `ingesters/HttpIngester/main.go:330` does `if hcfg.tag, err = igst.NegotiateTag(v.Tag_Name); err != nil { return fmt.Errorf(...) }` — so a connection bounce racing startup/hot-reload tag setup now aborts ingester startup (after a 10s stall) for a tag that is perfectly fine. Pre-PR this call could not fail at all. Same for `otel_logs.go:398`, `amazon_firehose.go:174`, `hec_config.go:256`.

Fix direction: when a `tagTrans` is abandoned, settle its queue. Two reasonable shapes — (a) call `tt.failQueued(ErrNotRunning)` from the connection-teardown paths (cheap, but then the caller gets an error for an established tag), or (b) better, carry the pending waiters forward: have `reconcileTags` (or the install path) re-attach the old translator's queued waiters to the new one, and report success for any of them whose tag the handshake already covered (`tt.hasTag(local)`). (b) matches the documented contract; (a) at least stops the 10s stall.

**2. `NegotiateTag` returns `nil` immediately for any name already in `im.tagMap`, even when that tag has never been established on any indexer — so the documented contract ("a nil error means the tag is established on at least one indexer and is searchable") is false for the second and every subsequent caller**
`ingest/muxer.go:891` and `:964-968`.

`stageTag` returns `(tg, nil, nil)` on the `tagMap` hit, and `NegotiateTagContext` returns immediately when `w == nil`. Two concrete scenarios:
- Retry-after-failure: the HTTP ingester's reload negotiates `foo`, gets the `ErrTimeout` from finding 1 (or a genuine `ERROR_TAG_MAGIC`), fails the reload and logs it. The operator fixes whatever it was and SIGHUPs again → this time `NegotiateTag("foo")` returns `nil` instantly *without contacting anything*, the reload reports success, and the operator believes the tag is searchable. It may still be queued on a wedged relay routine.
- Concurrent callers: two simultaneous HEC requests with `?tag=foo` (`ingesters/HttpIngester/hec.go:134`). Caller A stages and waits for establishment; caller B returns `nil` immediately and proceeds as if the tag is searchable. `TestNegotiateTagConcurrentDistinctNames` uses distinct names, so it can't see this.

Fix direction: keep in-flight waiters in the muxer keyed by tag name (a `map[string]*negotiationWaiter` cleaned up on settle) so repeat and concurrent callers join the existing waiter instead of short-circuiting; alternatively, on the `tagMap` hit, check the live translators with `tt.hasTag(tg)` and build a fresh waiter if any is missing it.

**3. `TestNegotiateTagQueuedBehindFailure` cannot fail for the reason it exists — mutation-proven**
`ingest/muxer_test.go:696-733`.

The test's stated purpose is that a tag queued behind a failing one does not "sit out the whole negotiation timeout", but its guard is `time.After(30 * time.Second)` while `tagNegotiationTimeout` is 10s. I neutered `failQueued` to an immediate `return` and reran it three times: PASS every time, in ~10.0s, with `behind err: Timed out waiting for ingesters` — i.e. it burns exactly the timeout the test claims to prevent and still passes. Two other problems with it: the `ti.holdPostAuth()` call does not do what the comment claims (it parks *new* connections; the already-established connection's relay routine drains the queue normally), so whether the two tags ever co-exist in the queue is a race — in 5 runs I saw `behind` come back with the blocker's error 4 times and `nil` once; and the result for `behind` is only `t.Logf`'d, never asserted.

Fix direction: bound the wait well below `tagNegotiationTimeout` (2-3s), and assert `errs[1]` non-nil *and* mentioning `blocker` (i.e. that it came from `failQueued`, not from the timer). To make the pile-up deterministic rather than racy, the harness needs a way to hold an *established* connection's relay routine (the existing read `gate` does that — holding it stalls acks, which stalls the `forceAckNoLock` at the head of `NegotiateTag`).

**4. The `notify`-failure bounce is the only bounce path in `writeRelayRoutine` that skips `syncAndCloseConnection`**
`ingest/muxer.go:1680-1691` vs `:1735`, `:1783`, `:1797`, `:1859`, `:1880`, `:1893`, `:1919`, `:1932`.

Every other failure path calls `im.syncAndCloseConnection(nc)` (sync, close, then eject+recycle) before asking for a new conn set; the negotiation-failure path just signals `connFailure` and lets connRoutine's `igst.Close()` + `ejectOutstandingEntries()` clean up. No data loss (connRoutine does reverse-translate and recycle), but the final `syncTimeout` is skipped, so entries the indexer already received but had not yet acked are re-sent on the new connection: user-visible duplicates on every failed renegotiation. Same in the new `getNewConnSet` failure branch at `:1595-1608`. Fix direction: call `im.syncAndCloseConnection(nc)` before `getNewConnSet` in the notify branch, matching the other eight sites.

**5. `NegotiateTagContext`'s "a caller cannot wedge itself forever" claim does not hold, because `stageTag`'s `im.mtx.Lock()` is unbounded and ignores the context**
`ingest/muxer.go:886-897`, `:957`, and `:2173-2247`.

`getConnection` holds `im.mtx.RLock()` across `initConnection` — a full TCP dial plus auth plus tag handshake — and across `newTagTrans`. A `NegotiateTagContext` call with a 100ms deadline will still block in `stageTag` for the whole dial/handshake of any destination currently reconnecting, because the context is only consulted after staging. Pre-existing lock behavior, but the new doc comment makes a promise the code doesn't keep. Fix direction: either check `ctx.Err()` first and document that staging blocks on the muxer lock, or drop the "cannot wedge itself forever" wording.

Issue #2782, opened 2026-09-17
hosted runner plugin: Claude Compliance API


kyle-mallett-gravwell:
# Feature

Before opening a feature request, please complete the following checklist:

- [x] Check to see if this is a duplicate

## What is the feature to be added?

Hosted runner plugin to pull enterprise compliance activity, directory, conversation, project, artifact metadata, and session data from the Claude Compliance API.

## Why should we add this feature? (Business justification? What problem is the feature trying to solve?)

Increase the number of integrations and provide customers visibility into enterprise Claude and Claude Code activity.

## How does this feature address the above problem?

Adds read-only collection from the Claude Compliance API through the current Hosted Runner framework.

## Any other comments?

PR #2783, opened 2026-09-17
Plugin: Add Claude Compliance API hosted runner plugin

kyle-mallett-gravwell:
This PR addresses [(https://github.com/gravwell/gravwell/issues/2782)]

## This PR proposes...

- Adding a Claude Compliance API plugin to the current hosted framework

## PR Tasks

- [x] Bug fixes only: not applicable; this is a new feature.

## Reviewer Tasks


kris-watts-gravwell:
## Code review: correctness findings

Reviewed the full diff. I applied the patch to a worktree on `next-minor` — it builds, `go vet` is clean, and the included tests pass — then wrote throwaway probe tests to confirm each failure mode below before deleting them. Every "Verified" note is an observed result, not a reading of the code.

Findings 1, 2, 4 and 6 share a root shape worth naming up front: **`handleOne` writes entries as it goes but commits the checkpoint only after the last page.** Any error after partial writes therefore turns into an unbounded duplicate-ingestion loop rather than a bounded retry — the adapter retries every `JobErrorDelay` (30s) and each attempt re-writes everything it got through last time. `activities` has a monotonic `after_id` cursor that would support intermediate checkpointing.

---

### 1. Backlog over Max-Pages (or the 100k cap) replays entries forever — `plugin.go:213`

When a traversal exceeds `Max-Pages` or the hard 100,000-record cap, `handleOne` returns an error *after* having already written every fetched entry, so `rt.Put(c.key(), ...)` is never reached.

Verified with `Max_Pages=3` and `has_more=true`: three successive `Handle` calls wrote 3, 6, then 9 entries, with `saved` still nil each time.

For the `fullParents` datasets (`chats`, `projects`, `local-sessions`) the time filter is deliberately dropped, so an operator has no way to narrow the scan — the only escape is `Follow-Children="disabled"`.

### 2. More children than Max-Pending bricks the root dataset — `discovery.go:94`

When distinct discovered children exceed `Max-Pending` (default 10,000), the eviction scan finds no non-pending victim and returns an error *from inside `onRecord`*, which aborts the root traversal.

An org with >10k chats hits this on the very first poll: the root `chats` scan is unwindowed, so all of them are discovered and marked `Pending` immediately.

Verified with `Max_Pending=2` and 3 groups:
```
poll 0 err="pending child-work capacity reached" entries=3
poll 1 err=<nil>                                 entries=6   # 3 duplicates
poll 2 err="pending child-work capacity reached" entries=6
```
Group `g3`'s members are never fetched.

### 3. Whole worklist marshalled and fsynced once per record — `discovery.go:112`

`return persist()` runs unconditionally for every record, including datasets that produce no children at all. `activities` — what `example.conf` ships, with `Follow-Children="enabled"` — has no entry in `childWork`'s switch, so `children` is always nil, yet the full worklist is marshalled and `Put` anyway.

Measured: **500 activity records → 501 `rt.Put` calls.** `example.conf` also sets `State.Sync=true`, so each is a flushed state transaction.

Cost is O(records × worklist size): a 100k-record scan against a 10k-item worklist (~2 MB of JSON) marshals and writes on the order of 20 GB per poll. `persist()` should run once after the record loop, or only when children were actually added.

### 4. `_session` is bounded by Max-Entry-Bytes but the EV limit is ~63 KiB — `plugin.go:171`

The `session` envelope is size-checked against `Max_Entry_Bytes` (default 4 MiB), then attached as an enumerated value — `entry.MaxEvDataLength` is `63*1024 - evHeaderLen - 1`.

Verified with a 70,000-byte session blob: `err="Enumerated Data string is too large"`, `entries=0`, `saved=false`. The failure is deterministic, so every subsequent poll fails identically and that session's messages are never collected.

### 5. Discovered `chat-messages` are filtered to the lookback window — `plugin.go:96`

The parent `chats` scan intentionally drops its time filter so old chats are discovered, but the discovered child still applies `updated_at.gte`. Observed request pair on a first poll:

```
/v1/compliance/apps/chats?limit=100&order_by=updated_at
/v1/compliance/apps/chats/c1/messages?limit=100&order=asc
    &updated_at.gte=2026-09-07T00:00:00Z&updated_at.lte=2026-09-08T00:00:00Z
```

A chat last updated in 2020 is discovered, then its transcript request returns nothing. The child's own checkpoint then sets `since=now-overlap`, so the window only narrows — the history is permanently missed while the plugin reports success. `local-session-messages` and `remote-session-messages` have `Window=""` and are unaffected; `chat-messages` is the only child carrying a window in `catalog.go`.

### 6. One id-less parent record blocks the whole dataset — `discovery.go:220`

`childWork` errors when a parent record has no `id`/`uuid`; that propagates out of `onRecord` and aborts the entire traversal. Note `organizations` is already special-cased to `uuid`, so the assumption that every other family uses `id` is load-bearing.

Verified with three groups where the middle one lacks `id`:
```
poll 0 err="groups parent missing id" entries=2
poll 1 err="groups parent missing id" entries=4
```
`g2` (after the bad record) is never reached, and the good records re-ingest on every retry. A malformed single record should be logged and skipped.

### 7. Per-child checkpoints are never pruned — `plugin.go:206`

Each discovered child writes its own checkpoint under `c.key()` holding a manifest of every record it has seen. Nothing ever deletes those keys — not even the `Max-Pending` eviction at `discovery.go:97`, which removes the work item but leaves the state key behind.

10,000 chats × ~500 messages ≈ 5,000,000 manifest entries persisted in bolt, growing monotonically with lifetime tenant history, with deleted chats' checkpoints retained forever.

### 8. `credential()` uses a single `Read`, can truncate the key — `config.go:221`

`f.Read(b)` is not guaranteed to fill the buffer. On NFS/FUSE a short read silently yields a truncated API key; `Verify()` and `handleOne()` both succeed, and the only symptom is `Compliance HTTP 401; state retained` with no hint as to why. `os.ReadFile` (the `st.Size()` cap is already enforced above) removes the failure mode.

kris-watts-gravwell (changes requested):
see comments, couple high priority issues

kyle-mallett-gravwell:
@claude review

